PR #10, opened 2015-03-20
Add ng-conf 2015 youtube playlist

meDavid:


naomiblack:
I have a more recent PR that adds this playlist with more detail.

PR #12, opened 2015-03-21
Docs(resources): Add ng-conf video and forms blog post

naomiblack:
Also update chat to point to gitter


public/resources/css/module/_alert.scss
```diff
@@ -1,4 +1,5 @@
 .alert {
+  box-shadow: 0px 2px 5px rgba($coal, 0.3);
```
alexwolfe: :+1: nice catch ;)

PR #67, opened 2015-04-29
update dart angular2 version, don't display pub get output

kwalrath:
@alexwolfe, could you publish these small updates at your earliest convenience?

The existing text isn't wrong, but it shows old version numbers.


alexwolfe:
@kwalrath live :+1:

PR #73, opened 2015-04-30
Correct the showcase code

JonRimmer:
You probably don't want the first Angular 2 code people see to be wrong.


JonRimmer:
Never mind, I see @HenriqueLimas already submitted a PR for this.

PR #74, opened 2015-04-30
fix(quickstart): Update TS quickstart to use newly released TS 1.5.0 bet...

rkirov:
...a.

### DIFF
--- a/public/docs/js/latest/quickstart.jade
+++ b/public/docs/js/latest/quickstart.jade
@@ -42,8 +42,7 @@
     know to read this file so we don't need to configure them or add command-line options.
 
   pre.prettyprint
-    # We need to use an unreleased version of TypeScript
-    $ npm install -g mhegazy/typescript#v1.5-beta
+    $ npm install -g typescript@^1.5.0-beta
     $ tsc --watch
 
 // STEP 3 - Import Angular ##########################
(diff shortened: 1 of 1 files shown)


public/docs/js/latest/quickstart.jade
```diff
@@ -42,8 +42,8 @@
     know to read this file so we don't need to configure them or add command-line options.
 
   pre.prettyprint
-    # We need to use an unreleased version of TypeScript
-    $ npm install -g mhegazy/typescript#v1.5-beta
+    # make sure you are using 1.5.0-beta and above
```
mprobst: Add a comment on how to do that? E.g. `tsc --version` or so?

rkirov: I was thinkging using `$ npm install -g typescript@^1.5.0-beta` instead. Thoughts?

mprobst: Yeah, better make it impossible to make a mistake.

alexeagle: yes, test that it works if you already have an earlier one installed?

public/docs/js/latest/quickstart.jade
```diff
@@ -217,12 +217,16 @@
...
+    Add the System.js dependency in the <code>&lt;head&gt;</code> tag, so that
+    it looks like:
 
   pre.prettyprint.linenums
     code.
         &lt;head&gt;
+          &lt;title&gt;Angular 2 Quickstart&lt;/title&gt;
+          &lt;script src="https://github.jspm.io/jmcriffey/bower-traceur-runtime@0.0.87/traceur-runtime.js"&gt;&lt;/script&gt;
           &lt;script src="https://jspm.io/system@0.16.js"&gt;&lt;/script&gt;
+          &lt;script src="bundle/angular2.dev.js"&gt;&lt;/script&gt;
```
alexeagle: we don't have a release pushed to CDN yet?

alexeagle:
LGTM but thanks for looking David

davideast:
Looks good, but just one comment.

Is there any reason why we include the script block before the component?

```html
<script>System.import('app');</script>
<my-app></my-app>
```

rkirov:
@davideast nope, in fact if System was sync that would be an error, will move it at the end of the `<body>` as best practice.

public/docs/js/latest/quickstart.jade
rkirov: nope, have to land https://github.com/angular/angular/pull/1620 first

PR #104, opened 2015-05-10
change the variable `todos` to `names`

lisposter:
In the code example below it, we use the `names` variable, so, `todos` here will be confused.


alexwolfe:
Thanks :+1:

PR #106, opened 2015-05-12
ES5 Quickstart

davideast:
Added a the ES5 version of the quickstart.

@rkirov Can you check for errors?

@alexwolfe How should we integrate this into the navigation?

cc: @naomiblack


alexwolfe:
@davideast and I talked about this and have a plan for how to integrate. Going to close this pull for now, we'll have a combined pull that takes into account the differences for these two.

PR #112, opened 2015-05-14
Make TypeScript default code sample

aciccarello:
To be consistent with rest of guide, the TypeScript code example should be shown first.

Note: I don't know jade so I won't be offended if the PR is rejected.


alexwolfe:
Thanks, we're actually planning on updating so ES5 is shown first. If you want to update you can submit this pull again, also it looks like you'll need to sign the CLA.

PR #125, opened 2015-05-19
First round of guide content fixes

jeffbcross:
I'll be making some more fixes, based on my internal doc shared with @naomiblack and @alexwolfe


alexwolfe:
Thanks for fixing

Issue #128, opened 2015-05-20
Small 5 Min Quickstart issue?


WonSong:
In the Import Angular section in https://angular.io/docs/js/latest/quickstart.html... 

Shouldn't 

```
<reference path="typings/angular2/angular2.d.ts" >
```

be 

```
/// <reference path="typings/angular2/angular2.d.ts" />
```

Instead?

I am getting typescript compiler error using 1.5.0 beta without ```///``` and the enclosing slash ```/>```

im-tollu:
It should, according to [http://www.typescriptlang.org/Handbook#modules-splitting-across-files]

WonSong:
Closed by #129

Issue #119, opened 2015-05-18
Update Download Page Formatting


alexwolfe:
This page needs a few facelift items: 

1. Remove the X from 1.X the titles should read "Angular 1" etc. not "Angular 1.X" 
2. Replace the links with download buttons (or it could be nice to use hover cards https://angular.io/docs/js/latest/guide/) 
3. Update Nav to say "Install" not "Download"

alexwolfe:
@gorork Feel free to take a look at this one.

WonSong:
Please see pull request #129

PR #129, opened 2015-05-20
#128: 5 Min Quickstart small issue fix & #119: Download page formatting

WonSong:
Creating a pull request in case this is a real issue. 

Before: 

image

After:

image

### DIFF
--- a/public/resources/css/module/_card.scss
+++ b/public/resources/css/module/_card.scss
@@ -36,17 +36,24 @@
       padding: 0px;
       margin: 0px 0px 0px ($unit * 4);
 
-      a {
-        font-size: 14px;
-        display: block;
-        line-height: 24px;
-        text-decoration: none;
-        padding: 0px 4px;
+      &:not(.hover-card-container) {
+        a {
+          font-size: 14px;
+          display: block;
+          line-height: 24px;
+          text-decoration: none;
+          padding: 0px 4px;
 
-        &:hover {
-          text-decoration: underline;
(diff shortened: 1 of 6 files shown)


WonSong:
Signed in

WonSong:
Before:

image

image

After:

image

image

public/resources/css/module/_card.scss
```diff
@@ -36,18 +36,25 @@
...
+            text-decoration: underline;
+            background: $mist;
+          }
         }
       }
+
+      &.hover-card-container {
+        margin-left: 0;
+        margin-top: $unit * 2;
+      }
```
alexwolfe: I think you should just replace the entire containers on the install page with the cards instead of putting them inside the card module. You should then be able to remove these extras styles. Thanks.

alexwolfe:
@WonSong This is great work, thanks! I just had one comment otherwise looks good. Ping me when updated and I'll merge. Thanks!

WonSong:
@alexwolfe Np & Thanks! Please review. 

I did the below unwillingly because I wasn't able to find a helper class to remove padding left on the ```<ul>```. 

```
-@for $i from 1 through 10 {
+@for $i from 0 through 10 {
```

Would you have preferred replacing ```<ul>``` with a ```<div>``` tag or something instead?

image

alexwolfe:
@WonSong :+1: Nice work, thanks for the fixes.

PR #138, opened 2015-06-03
fix(index): display code example correctly

jackmu95:
This commit fixes the displaying of the „name“ template variable (see
screenshot below) in the code example on the front page.

https://goo.gl/qcLf5X

[Review on Reviewable](https://reviewable.io/reviews/angular/angular.io/138)


alexeagle:
Thanks! Merged as 45d136e

PR #139, opened 2015-06-03
Update the Dart guide's code and text to alpha.26 (from 23)

kwalrath:
This was a fairly substantial change because:

* changed for/if -> ng-for/if
* converted the code to put each file in its own library, instead of using `part of`. (I hadn't realized many people really don't like `part of`, as evinced by the comments on https://github.com/angular/angular/issues/1817.)
* added coverage of the angular2 transformer
* changed every pubspec to include the transformer and to specify a version range for browser (now that "any" is discouraged)
* ran dartfmt over all the files

I've tested everything, but I'd welcome more testing. Anyone up for a run-through of these docs?

Also... I should put the example code files somewhere online.

### DIFF
--- a/public/docs/dart/latest/guide/displaying-data.jade
+++ b/public/docs/dart/latest/guide/displaying-data.jade
@@ -17,22 +17,19 @@
     a <code>pubspec.yaml</code> file:
 
   code-tabs
-    code-pane(language="dart" name="main.dart" format="linenums").
-      // web/main.dart
-      library displaying_data;
-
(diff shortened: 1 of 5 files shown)


kwalrath:
I'm updating this to 26. (See also issue #141.)

kwalrath:
@davideast here are my changes that correspond to #142.

kwalrath:
OK, I'm going to commit and push the site. Scream if you disagree....

PR #148, opened 2015-06-12
Typo in code example of lib/show_properties.dart

cbart:
In the text show_properties.dart is (correctly) placed in the lib directory,
while in the comment at the top of the file, it's path is under web.


kwalrath:
Thanks for the fix!

kwalrath:
FYI, we probably won't deploy the site until next week, so don't worry if you don't see the fix online.

PR #163, opened 2015-07-10
Fixed ng-for typo && fix param in addTodo()

lisposter:
* `*ngfor` is not right, it should be `*ng-for`
* `addTodo()` should push the value of the DOM element


wardbell:
Although this doc and example will be replaced, your changes correct the current state of the code and are welcome as an intermediate step.Thanks to you for engaging and helping. We need that!

PR #161, opened 2015-07-04
User input js step by step guide small fixes

thierryve:
Closes #159

PR #167, opened 2015-07-12
Fixing typo for `NgIf` directive

styfle:
The directive was imported as `NgIf` but the view is using `NIf ` which I believe is a typo.


vicb:
Addressed in #161, please close this issue

PR #171, opened 2015-07-22
Provide more options for dartdocs

yjbanov:
specifically http://www.dartdocs.org

### DIFF
--- a/public/docs/dart/latest/api/index.jade
+++ b/public/docs/dart/latest/api/index.jade
@@ -2,9 +2,16 @@
   h2 Developer Preview
 
   p.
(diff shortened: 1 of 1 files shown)


naomiblack:
Sorry -- I know I told you to make a PR -- but I've been looking at what we have on dartdocs, and it's just really broken. It's basically TypeScript + broken links and dgeni artifacts, output with the dartdoc styling. I don't want to draw any attention to it until we're on top of the single source doc generation and actually have dart output not fake TS masquerading in dart clothes. So closing this one -- let's fix it properly and then announce.

PR #179, opened 2015-08-05
shred and watch build tooling ( dgeni, gulp, jade)

jtraband:
- New gulp files for automation
- New directories for documentation examples and fragments
- New doc-shredder dir ( wrapper over dgeni task to shred files from one dir into another based on #docregion tags, ( + internal test dir)
- New mixin utility function for jade/harp to pull in fragments
- New sample 'alternative' version of the 'setup.jade' page to show example of shredder
- Updated ReadMe.md


jtraband:
I signed it!

public/doc-shredder/test/LICENSE
```diff
@@ -0,0 +1,201 @@
+Apache License
```
btford: I don't think that this license is correct...

@naomiblack can you verify and prescribe the right way to do it?

public/_includes/_util-fns.jade
```diff
@@ -0,0 +1,35 @@
...
+
+- var getExtn = function(fileName) {
+-   var ix = fileName.lastIndexOf('.');
+-   return ix > 0 ? fileName.substr(ix+1) : "";
+- }
+
+// HACK: to avoid having to include a path in makeTabs calls
+- var currentPath = current.path;
+// need to back up to 'docs'
+- var pathToFrags = "../../../../../../../../../../../".substr(0, (currentPath.length-2)*3) + "_fragments/";
```
btford: I have trouble following this – you want to add `../` once for each of `currentPath`'s length minus 2? Would currentPath ever have a length > 13? Can you please add more explanation here, and maybe throw for cases when `currentPath` has a length that would be out of bounds?

Maybe an example of the expected values for the preceding couple of lines.

public/doc-shredder/test/test_source/app.js
```diff
@@ -0,0 +1,24 @@
...
+  // #docregion greet
+  /**
+   * @description  Display a greeting
+   * @param {string} name The name of the person to greet
+   */
+  greet: function(name) {
+    console.log('hello ' + name);
+  }
+  //   #enddocregion
+};
```
btford: missing EOF new lines

btford:
A couple things that could use better docs:

- [ ] Why `public/doc-shredder` has its own `package.json`, and how those deps are automatically `npm install`ed
- [ ] Why the generated fragments (and any other generated files) are checked in.

Can you add that to the `readme` in `public/doc-shredder` (or wherever you think is appropriate)?

Other than that, looks good to me! :+1:

jtraband:
Only the ‘test’ dir in doc-shredder has its own package.json because I wanted to isolate any test specific dependencies.  I’ll probably remove this ( the package.json) later because it doesn’t seem like I need it any longer.

The generated files are checked in so that we can avoid running the shredder over the entire _examples dir each time someone refreshes the repo ( the ‘shred-full’ gulp task).  The gulp ‘serve-and-watch’ shredder is only a ‘partial’ shredder.  It only shred’s files in directories changed during the current session.  Hope this makes sense.

Can I do this in the next version or do you want these comments put in this version?

public/doc-shredder/test/test_source/app.js
jtraband: This is a test file, and we want to insure that the shredder works with 'whatever' someone checks in for example code.

btford:
If you could amend this commit with what you just said in your comment,
that'd be perfect.

jtraband:
No problem.

jtraband:
Ok, removed Licence file and extra superflous package.json in doc-shredder test dir. Added additional readme files and comments.

PR #200, opened 2015-09-01
Dev tooling2

jtraband:
Added support for fragment styling via regexp for makeTabs and makeExample
Added gulp help support
Added gulp task to determine which examples and associated jade files are affected as a result of a given git commit.
... and associated refactoring and cleanup.

PR #203, opened 2015-09-09
IdeaBlade: dev-tooling-2

wardbell:
This is a re-submission of #200  (Dev Tooling).

I removed the spurious addition of `~/public/doc/ts/latest/api` and squashed into 2 commits

1. Mostly infrastructure

2. Revision of getting started example code and companion Dev Guide + unit tests for the QuickStart that exist solely as an example and foundation for evolving the infrastructure to run unit-tests.  Jay will be picking up on that.

The example code is still based on alpha-35 and system.js v.0.16.   I'm asking that you accept this anyway because (a) it works, (b) it's more accurate/current than what is published now and (c) we can clear the backlog of stuff we have in our IdeaBlade fork that is way in front of angular.io.

We will update for alpha-37 and system.js v.0.18+  very soon (but no sooner than alpha-37 release)

I counted 150 filename differences from master in this PR using `git diff --name-status master..dev-tooling-2`.  That's a lot of files. More than a PR should have ... which partially reflects how long it has been between PRs and also the number of _fragment files which are generated. 

>I am uncomfortable w/ keeping _fragment files under source code (unless they are snapshots of specific releases ... which we are very far from).  However, this is a decision to be determined; for the moment ... we're checking them in.

My review of the affected files shows that ALL BUT FOUR are in the strict domain managed by IdeaBlade under this project.

The four that *may* fall outside of our strict domain are:
* .gitignore
* README.md
* gulpfile.js
* package.json

I think you'll find those changes agreeable.

A summary of the others:

2  `public/_includes/..`   - jade functions for the infrastructure
4  `public/doc-shredder/..`
51 `public/docs/_examples/..`
65 `public/docs/_fragments/..`
2 `public/docs/_includes/styleguide/` - for the expansion of the style guide to teach use of infrastructure
4 `public/docs/js/latest/` - changes related to introducing the infrastructure to JS DevGuide docs
15 `public/docs/ts/latest/` - copies of the original `~js/latest` with substantive changes to `gettingStarted.jade`
3 `public/docs/xref-...`
- - - -
146

Add the 4 I mentioned at the top and you have the 150 files from the `diff`

I can't explain why github reports 133 file changes. Maybe someone can.

Happy to discuss any of this of course.


wardbell:
Can someone explain why googlebot is barking about our CLAs. Both Jay and I have signed them and we can't tell what email address might have been used in a commit that googlebot doesn't know.

btford:
@wardbell Googlebot isn't smart enough to deal with PRs that contain commits from more than one author. You can ignore it here, provided you and Jay have both signed the CLA already.

btford:
I have some questions that I left in the code. In general:

- [ ] We should try to use 2-space tabs everywhere to stay consistent
- [ ] There are a lot of places with separate references to `alpha34`. Is it possible/reasonable to consolidate them?

wardbell:
"If this has a code block, should we also be indenting it this deeply? Should the fragments be normalized"

> in regard to `e2e-spec-shared.js.md` from `public/docs/_examples/gettingstarted/e2e-spec.js`

That is an interesting observation.  I like your idea because it is quite likely that a guide author will extract a portion of deeply indented code and I think we would like to see that shifted as far left as possible.

I think we go all normalized w/r/t left alignment or not. I thought about offering an option and that's just gilding the lily.

I'll forward this to Jay who can certainly do it. We can see how it looks.

No action required for this PR #203

wardbell:
@btford Thanks on the CLA pass. I suspect this will happen a lot.  If you have the power to push it through, we'll keep making note of the fact that we signed.

wardbell:
Re: 2 spaces ... Yes.  I've only recently converted to 2 spaces for everything so there are artifacts that we must drive out of the code base.  We need to get jsHint and jscs in the tool mix so we can catch them. Also the TS equivalents.  Not sure how to schedule those efforts.  Right now we're fighting to get content.  But agree completely with the (eventual requirement) and I'm fixing those cases you identified in a forthcoming patch to this PR

wardbell:
Re: references to alpha34:  We will sweep the code to get it all on "latest" once alpha-37 comes out.  

We think we may be able to automate that because all samples will reference node_modules and we intend to automate management of the corresponding package.json files.

This is on our radar. Shouldn't be a blocker for this PR though, right?

wardbell:
Updated w/ your requested changes. Awaiting your approval sir

btford:
Great! Go ahead and merge.

wardbell:
That's a note to whom? I don't have the power, right? This is my first google PR so I'm new to the process.

naomiblack:
It's a note to me :) I was just reviewing but note that this change pushes the new getting started split to TS and JS live. Are we ready to ship it?

naomiblack:
this is now in master.

PR #221, opened 2015-09-22
Directive doctypes - initial functionality

petebacondarwin:
The new processor will convert classes that are annotated (decorated) as Directives, to a new doc type of `directive` and also extract the juicy directive metadata into the doc so that it can be displayed nicely.

### DIFF
--- a/public/api-builder/angular.io-package/templates/directive.template.html
+++ b/public/api-builder/angular.io-package/templates/directive.template.html
@@ -1,0 +1,33 @@
+{% include "lib/githubLinks.html" -%}
+{% include "lib/paramList.html" -%}
+{% extends 'class.template.html' -%}
+
+{% block additional %}
+.l-main-section
+  h2 Selector
+  .l-sub-section
+    h3.selector
+      code {$ doc.directiveOptions.selector $}
+
+.l-main-section
+  h2 Inputs
+  .l-sub-section
+    h3.input {% for binding, property in doc.inputs %}
+      code {$ property.bindingName | dashCase $}
+      | &nbsp;bound to&nbsp;
+      code {$ property.memberDoc.classDoc.name $}.{$ property.propertyName $}
+    :markdown
+{$ property.memberDoc.description | indentForMarkdown(2) | trimBlankLines $}
+  {% endfor %}
+
+.l-main-section
+  h2 Outputs
+  .l-sub-section
+    h3.input {% for binding, property in doc.outputs %}
+      code {$ property.bindingName | dashCase $}
+      | &nbsp;bound to&nbsp;
+      code {$ property.memberDoc.classDoc.name $}.{$ property.propertyName $}
+    :markdown
+{$ event.memberDoc.description | indentForMarkdown(2) | trimBlankLines $}
+  {% endfor %}
+{% endblock %}
--- a/public/api-builder/angular.io-package/templates/directive.template.html
+++ b/public/api-builder/angular.io-package/templates/directive.template.html
@@ -4,11 +4,14 @@
 
 {% block additional %}
 .l-main-section
-  h2 Selector
+  h2 Selectors
   .l-sub-section
+    {% for selector in doc.directiveOptions.selector.split(',') %}
     h3.selector
-      code {$ doc.directiveOptions.selector $}
+      code {$ selector $}
+    {% endfor %}
 
+{% if doc.inputs.length %}
 .l-main-section
   h2 Inputs
   .l-sub-section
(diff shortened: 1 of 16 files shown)


petebacondarwin:
This is related to https://github.com/angular/angular/issues/4313

pkozlowski-opensource:
Thnx @petebacondarwin - this looks like a good start! From what I can see it would allow us to display a selector and list properties from `@Directive` / `@Component` annotation (after adding an appropriate template). But thinking about it some more it looks like we would have hard time providing custom description for properties / events (and I guess we will need some to indicate what a given property does, what type of arguments it accepts etc.).

So maybe we should focus on parsing out `@Property` / `@Event` decorators (those are added to class members and I guess we could have documentation attached to those members).

Let me know if you need any more info here!

petebacondarwin:
We should be able to match up the docs on members of a directive class with the properties and events in the decorators.

petebacondarwin:
I don't see any directives in core that use `@Event` or `@Property`, yet. I guess we have to support both?

pkozlowski-opensource:
Yeh, core directives are still using the "old" syntax so yes, I agree that we need to support both. 

> We should be able to match up the docs on members of a directive class with the properties and events in the decorators.

This would be great. This matching would have to take both properties and setters into account (not sure if it makes impl any harder)

petebacondarwin:
@pkozlowski-opensource here is a bit more. It gets the `properties` and `events` for each directive and there is a basic template.
Check out that it now pulls the description from the relevant member and can use either the class decorator (`@Directive`) or the appropriate member decorator (`@Property` and `@Event`).

public/api-builder/angular.io-package/templates/directive.template.html
```diff
@@ -0,0 +1,37 @@
...
+{$ doc.description | indentForMarkdown(2) | trimBlankLines $}
+
+.l-main-section
+:markdown
+  ## Directive Information
+
+  ### Selector
+  `{$ doc.directiveOptions.selector $}`
+
+  ### Property Bindings
```
pkozlowski-opensource: The "binding" word doesn't get good press recently as it is really overloaded (currently we can have property bindings, host bindings, bindings in DI and each of them means a different things despite using the same name).

I would say let's change it to just `Properties`

public/api-builder/angular.io-package/templates/directive.template.html
```diff
@@ -0,0 +1,37 @@
...
+  {% for binding, property in doc.propertyBindings %}
+
+  #### `{$ property.bindingName $}` : bound to `{$ property.memberDoc.classDoc.name $}.{$ property.propertyName $}`
+
+  {$ property.memberDoc.description | indentForMarkdown(2) | trimBlankLines $}
+  {% endfor %}
+
+  ### Event Bindings
```
pkozlowski-opensource: Same here - let's drop "Bindings" and let's say "Events" or "Emitted events"

pkozlowski-opensource:
@petebacondarwin left 2 cosmetic comments. Once again I'm not familiar enough with dgeni internals to properly review the code. But I would say - let's take it in sooner than later so we can see how directive info is being rendered so we can improve iteratively on it if needed.

petebacondarwin:
Thanks for the feedback @pkozlowski-opensource.
I expect that someone with more skills than I will do a job on the template at a later stage.

Can you check that programmatic side of things, e.g. the extraction of data from the source, is inline with what to expect? In particular what data exactly needs to be extracted. Currently the `matchupDirectiveDecoratorsProcessor` is only doing events and properties but I appreciate that there are other things that are important.

pkozlowski-opensource:
@petebacondarwin ok, so let's taking a list of `@Component` properties, I think that selector, properties and events are the essential ones:
* selector - important for usage, we've got it covered
* properties - important for usage, we've got it covered
* events - important for usage, we've got it covered
* host - impl detail, might be helpful for users
* bindings - impl detail
* exportAs - bah, I didn't know that we've got it, not sure here :-)
* moduleId - tpl impl detail
* queries - impl detail
* compileChildren - impl detail

So, looking at this list IMO only `selector`, `properties` and `events` are essential for directive users. `host` might be useful and `exportAs` is to be checked. 

My advice: let's start with what we've got and iterate. 

@petebacondarwin could you rebase so I can test it with directives we've got documentation for?

petebacondarwin:
Rebased.

pkozlowski-opensource:
@petebacondarwin thnx! I'm just taking it for the test drive. Will report if I find anything.

In the meantime I've checked what is the deal with the `exportAs` property in the `@Component` / `@Directive` and it actually matters for template authors. Assuming that I've got a directive like so:

```
@Directive({selector: '[foo]', exportAs: 'foo'})
class Foo {
}
```

I can use it as a `<div foo #bar="foo">` - in other words I can declare a template local variable that points to a specific directive.

All in all I would say that yes, we need to parse out this property and have it in the output.

naomiblack:
any update on this one?

petebacondarwin:
I think that, once #339 lands, we can land this one as well @naomiblack.
While there is more that can be done with directives, it is good enough to get us started.

PR #222, opened 2015-09-23
Add Support for @example, @exampleTabs, @linkDevGuide tags

jtraband:
Support for new angular/angular inline tags: @example, @exampleTabs, @linkDevGuide
Updated gulp tasks to support shredding and watching of the angular/angular repo
Revised signatures for mixins ( to align with new inline tag defs) + fixup of an existing usage of these mixins.
Misc cleanup


btford:
In general, these commits seem like arbitrary snapshots in the development of this feature rather than specific features/fixes. This makes it harder for me to review.

I don't think it makes sense to rework these commits this time, but next time can you please spend a bit of time cleaning up commits to better communicate their intent? Thanks!

btford:
LGTM, merge away. :+1:

naomiblack:
merged!

PR #223, opened 2015-09-24
update dart docs to alpha.38

kwalrath:
@rkirov or @yjbanov could you please take a look? We wouldn't merge/publish this until alpha.38 is out, of course.

### DIFF
--- a/public/docs/dart/latest/guide/displaying-data.jade
+++ b/public/docs/dart/latest/guide/displaying-data.jade
@@ -18,7 +18,6 @@
 
   code-tabs
     code-pane(language="dart" name="web/main.dart" format="linenums").
-      import 'package:angular2/angular2.dart';
       import 'package:angular2/bootstrap.dart';
       import 'package:displaying_data/show_properties.dart';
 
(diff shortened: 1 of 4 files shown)


public/docs/dart/latest/guide/displaying-data.jade
```diff
@@ -431,16 +430,14 @@
...
-      <code>bootstrap.dart</code>.
+      <code>angular2.dart</code> and <code>bootstrap.dart</code>.
       Both <code>show_properties.dart</code> and <code>friends_service.dart</code>
-      import <code>angular2.dart</code> instead,
-      because they use Angular APIs but aren't entry points.
-      (They don't call <code>bootstrap()</code>.)
-      See <a href="/docs/dart/latest/quickstart.html#performance">Performance,
-      the transformer, and Angular 2 libraries</a> for more information.
+      import only <code>angular2.dart</code>
+      because they aren't entry points.
```
yjbanov: I think we no longer need to talk about non-entry points. You no longer import `bootstrap.dart` simply because you wouldn't be using the `bootstrap` function. Dart analyzer will flag any code that imports `bootstrap.dart` and does not use it.

I think we can rephrase this as (roughly): "To bootstrap your Angular 2 application in your entry point import `bootstrap.dart` and call `bootstrap`. To define a component or a directive, import `angular2.dart` and use `@Comonent` or `@Directive` respectively".

yjbanov: If you _happen_ to both bootstrap the app and define components in the same file, then, naturally, you'd import both.

yjbanov:
LGTM, but `alpha.38` isn't released yet

public/docs/dart/latest/guide/displaying-data.jade
kwalrath: Ah, so I should remove the angular2.dart import from all the web/main.dart files that don't use @Component (or any other annotation). That affects every example except for the very first ones (getting started & quickstart). I blame WebStorm. It didn't mark the import as unused. :)

kwalrath:
I've updated the imports and import discussions.

yjbanov:
Great! LGTM.

naomiblack:
reopening so we can update to alpha-39 when the issues in 38 are resolved. (38 won't be pushed to pub)

PR #235, opened 2015-10-05
cleans up various alignment and layout issues

robertmesserle:


naomiblack:
bullets (ul and ol) in the devguide are now too far to the left. See screenshot.
screen shot 2015-10-08 at 8 00 38 pm

PR #243, opened 2015-10-08
styling of download button

vikas-parashar:
- adds 'md-raised' 'md-primary' classes to download button
- removes 'button-secondary' class from download button

Download button in the bottom of the page has very low contrast in and the text is barely visible in both normal and hover state.

angular.io download button

Tested in [Color Contrast Analyzer](https://chrome.google.com/webstore/detail/color-contrast-analyzer/dagdlcijhfbmgkjokkjicnnfimlebcll?hl=en)
angular

Tried other button colors and went with primary button. 
after the changes:
changed

### DIFF
--- a/public/_includes/_cta-bar.jade
+++ b/public/_includes/_cta-bar.jade
@@ -1,3 +1,3 @@
 .cta-bar
   a(href="/docs/js/latest/quickstart.html" class="button button-large button-shield md-raised md-primary" md-button) Learn in 5 Min
-  a(href="download/" class="button button-large button-secondary" md-button) Download
+  a(href="download/" class="button button-large md-raised md-primary" md-button) Download


alexwolfe:
Thanks for your contribution. This is not a primary button (download) and should not be styled or emphasized the same way as a primary action (learn). It's possible to use a darker shade of the "blue grey" material design to increase the contrast ratio of the secondary button. This button should not be emphasized the same way though so please start with lighter (less heavy) shades of blue great. Thanks.

vikas-parashar:
So should I patch it as you suggested?

naomiblack:
The color contrast ratios in Angular Material are consistent with the spec at the time we built it. We do have an update coming -- at the ngmaterial level -- that will fix this for all sites that use Angular Material. 

Will not fix here.

PR #251, opened 2015-10-12
Quickstart guide: Add tsc 'experimentalDecorators' flag

franzheidl:
Add `--experimentalDecorators` flag to quickstart `tsc --watch …` command to so no error will be output.


naomiblack:
this guide is obsolete and will be replaced very shortly

PR #252, opened 2015-10-13
Devguide getting started 4

wardbell:
Update GettingStarted for alpha-40 with @View merged into @Component

Started to do the same for ToH-Pt1.  Realized there were many other gaps in conversion, presentation, and flow. Filled them.

### DIFF
--- a/public/docs/_examples/gettingstarted/ts/package.json
+++ b/public/docs/_examples/gettingstarted/ts/package.json
@@ -14,7 +14,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "angular2": "2.0.0-alpha.39",
+    "angular2": "2.0.0-alpha.40",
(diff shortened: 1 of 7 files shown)


public/docs/_examples/gettingstarted/ts/package.json
```diff
@@ -14,7 +14,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "angular2": "2.0.0-alpha.39",
+    "angular2": "2.0.0-alpha.40",
```
naomiblack: this will be 41 within 24 hours

wardbell: True but the other GS changes matter more.  And soon it will be 42 :-).  Please accept this as my next PR builds on it.

PR #253, opened 2015-10-13
Update getting started for alpha.41 - no more tsd

alexeagle:
cc @wardbell @johnpapa

This isn't totally finished. For example, it doesn't run the version of  `tsc` installed locally (though https://github.com/Microsoft/TypeScript/issues/5157 will fix that)

Ideally I'd like to point to updated getting started when alpha.41 is cut, so can we work through this?

### DIFF
--- a/public/docs/_examples/gettingstarted/ts/src/tsconfig.json
+++ b/public/docs/_examples/gettingstarted/ts/src/tsconfig.json
@@ -6,6 +6,6 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "removeComments": false,
-    "noImplicitAny": true
+    "noImplicitAny": false
   }
 }
--- a/public/docs/ts/latest/guide/gettingStarted.jade
+++ b/public/docs/ts/latest/guide/gettingStarted.jade
@@ -61,19 +61,18 @@ include ../../../../_includes/_util-fns
     `@Component` tells Angular that this class *is an Angular component*.
     The configuration object passed to the `@Component` method
     specifies a CSS selector for an HTML element named `my-app`.
-    When Angular sees `my-app`, it will know to
-    create and display an instance of our component.
-
-    `@View` is another decoration that describes how our
+    It also describes how our
     component renders on the screen. This one is dead simple,
     a single line of HTML announcing "My First Angular App".
+    When Angular sees `my-app`, it will know to
+    create and display an instance of our component.
 
     The `bootstrap` line tells Angular to start the application with this
     component at the application root.
     We'd be correct to guess that someday our application will
     consist of more components arising in tree-like fashion from this root.
 
-    In the top line we imported the `Component`, `View`, and `bootstrap` methods
+    In the top line we imported the `Component` and `bootstrap` methods
     from the Angular 2 library. That's the way we do things now.
     We no longer expect to find our code or any library code in a global namespace.
     We `import` exactly what we need, as we need it, from named file and library resources.
@@ -153,13 +152,6 @@ include ../../../../_includes/_util-fns
     compile locally and push the generated JavaScript to the server. We'll need some tools for that.
 
 
-    * We are writing TypeScript because we want strong-typing and some information
-    about the APIs we're using. If we wrote `AppComponent` in a TypeScript-aware editor,
-    we saw lots of red squiggly lines complaining about our code and
-    we received no guidance about what `Component`, `View`, and `bootstrap` can do.
-    We'll want to load TypeScript definition files to improve our coding experience.
-
-
     * Downloading JavaScript libraries from the web is OK for demos but
     it slows our development. Every time our app reloads, it must refetch these libraries.
     Don't count on browser caching.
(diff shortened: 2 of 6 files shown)


public/docs/_examples/gettingstarted/ts/src/tsconfig.json
```diff
@@ -1,11 +0,0 @@
-{
```
johnpapa: Why remove the `tsconfig.json` ?

We still need to transpile the TypeScript and these are the settings we are using for this and ToH.

cc / @alexeagle

alexeagle: The user needs a `tsconfig.json` - question is how do we instruct them how to create it.
This is a judgement call.

I like tsc --init because:
- most of the content of this file is default settings, you really only need three of them specified
- it is also in the copy-paste-free "from scratch" install path
- makes it more obvious that this file is the official configuration for tsc, not just some config file we invented

But it has drawbacks also:
- need to run it in the src subdir
- easy to run the wrong version of typescript which IIUC your 'npm run tsc' fixes by getting the local version

johnpapa: Right, for the reasons you point out I think its worth in getting started, just to give it to them. GS is about getting moving fat and getting early positive reinforcement. I think its better for now to give it to them.

Otherwise its as you described:
- go to the folder
- init
- change settings
- go back a folder

that is friction i would rather avoid in GS.

alexeagle: Okay, but then I would strongly suggest we remove --noImplicitAny from
tsconfig.json. This goes against what Martin and I are planning for our
talk (you can see the dry run tomorrow...)

johnpapa: you can remove that if its a problem. i think its a good thing, and prevents larger issues later. But we can discuss that more later. For now, I suggest you add the file back and remove the implict any

public/docs/ts/latest/quickstart.jade
```diff
@@ -60,13 +55,9 @@
 .l-main-section
   h2#section-transpile 3. Import Angular
 
-  p Inside of <code>app.ts</code>, import the type definitions from Angular:
+  p Inside of <code>app.ts</code>, your editor should be able to complete the available imports:
```
johnpapa: dont we also need to say the module system? when indicating systemjs it has failed often for me. when using commonjs it works.

johnpapa:
I recommend we do not ask the user to type the `tsc init` with all the parameters as there are 4 at least. Instead for GS it should just be "here is your file".

The file should include

```
{
  "compilerOptions": {
    "target": "ES5",
    "module": "commonjs",
    "sourceMap": true,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "removeComments": false,
    "noImplicitAny": false
  }
}
```

some might be removed

noImplicityAny should be true, IMO. But I see the value in not worrying folks with that implication this early.

public/docs/ts/latest/quickstart.jade
alexeagle: We don't need to specify since we hand the user a tsconfig.json file.

I've not tried it with tsc --module system. I agree that commonjs is the right setting

alexeagle:
I might not have time this week to finish this (still preparing a talk/demo) so if you guys would like to continue owning it, please feel free to take over. I just wanted to contribute some pointers.

johnpapa:
I'll defer to @wardbell as he owns the GS

wardbell:
I've updated GS to 42 including removal of tsd and typing files (I may need them back for testing but that's a story for another day).

Now tsc is complaining about the typing files it finds in the Angular bundles. I sent a slack on it.

public/docs/ts/latest/guide/gettingStarted.jade
```diff
@@ -253,8 +242,8 @@ include ../../../../_includes/_util-fns
     Enter these commands:
     ```
     npm init -y
-    npm i angular2@2.0.0-alpha.39 systemjs@0.19.2 --save --save-exact
-    npm i typescript tsd live-server --save-dev
+    npm i angular2@2.0.0-alpha.41 systemjs@0.19.2 --save --save-exact
```
johnpapa: this should be 42

public/docs/ts/latest/quickstart.jade
```diff
@@ -18,16 +18,11 @@
...
-    To get the benefits of TypeScript, we want to have the type definitions available for the compiler and the editor.
-    TypeScript type definitions are typically published in a repo called <a href="http://definitelytyped.org/">DefinitelyTyped</a>.
-    To fetch one of the type definitions to the local directory, we use the <a href="https://www.npmjs.com/package/tsd">tsd package manager</a>.
-
+    from this directory. Install typescript and angular from NPM.
+    
   code-example.
-    $ npm install -g tsd@^0.6.0
-    $ tsd install angular2 es6-promise rx rx-lite
+    $ npm install -g typescript
```
johnpapa: should this be -g ? i thought there were no globals yet at this point

public/docs/ts/latest/quickstart.jade
```diff
@@ -46,8 +41,8 @@
     are loaded, or configure your editor or IDE to do it.
 
   code-example.
-    $ npm install -g typescript@^1.5.0-beta
-    $ tsc --watch -m commonjs -t es5 --emitDecoratorMetadata app.ts
+    $ tsc --init --experimentalDecorators --target es5 --emitDecoratorMetadata
```
johnpapa: why do this? if you already have the tsconfig.json there is no need

wardbell: Ignore QuickStart ... it is dead ... soon to be replaced by Getting Started.

We'll remove QS very soon and replace with GS.

Alex may not know about this plan yet.

public/docs/ts/latest/guide/gettingStarted.jade
wardbell: Will be in next PR

alexeagle:
Thanks @wardbell I will drop this and look forward to your PR.

PR #256, opened 2015-10-14
Added Marc Laval

mlaval:
Adding Marc Laval to bios


naomiblack:
in master. (not yet live on angular.io)

PR #260, opened 2015-10-15
add links to Angular Connect live streams

petebacondarwin:


naomiblack:
in master. (not yet live on angular.io)

PR #262, opened 2015-10-15
(docs) Move ToH-pt2 to Jade and break toh-pt1 out of the tutorial index

wardbell:


naomiblack:
looking good. in master. (not yet live on angular.io)

PR #263, opened 2015-10-15
Fix the final markup to represent the tutorial

frankamp:
Also fixes the link to the next tutorial


wardbell:
Thanks for finding this!  Much appreciated. I'm manually merging your fix

PR #270, opened 2015-10-16
Update Dart QS to resemble TS one, use _examples

kwalrath:
I (or someone else) need to go through this one last time to make sure it's in shape to publish... but I think it should be OK.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,11 @@ node_modules
 bower_components
 jspm_packages
 typings
+packages
+build
+pubspec.lock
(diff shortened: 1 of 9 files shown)


naomiblack:
I can either merge it before my flight (at about 12:30pm Friday) or after I arrive in London. Let me know?

kwalrath:
2 reviewers have confirmed that this is good, so go ahead and merge it as soon as you practically can.

naomiblack:
merged and live!

PR #271, opened 2015-10-16
Fix styles property example, it requires an array

mbarbero:
It avoids compilation errors


wardbell:
Thanks for finding this!  Much appreciated. I'm manually merging your fix

mbarbero:
PR has been merged.

PR #274, opened 2015-10-16
Update toh-pt1.jade

jan-dh:
removed closing </input> tag


jan-dh:
I signed it!

wardbell:
Technically, the closing tag is an acceptable variant and many HTML authoring tools generate it (as happened here).

I'm removing it from this version of the doc but no promises that it will stay gone in future.

Thanks for your review though. Much appreciated.

PR #275, opened 2015-10-16
Update toh-pt2.jade

jan-dh:
moved the styling into an array (as it is the correct syntax)


wardbell:
Thanks for finding this!  Much appreciated. Someone else caught it too.

 I'm manually merging your fix

PR #277, opened 2015-10-16
Dev tooling 10

jtraband:
Fix for bug in api code gen where a {{ }} block occurs inside of backtics.

Fix is to translate backtic regions into code-example blocks.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -74,9 +74,16 @@ gulp.task('serve-and-sync', ['build-docs'], function (cb) {
 
 });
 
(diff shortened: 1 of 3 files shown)


petebacondarwin:
My only concern with this change is that, unless I am mistaken, it is adding behaviour to a processor that was designed for something else. We should create a new processor.

jtraband:
How about changing the name of the processor instead, because I need to do the matched backtic processing before I do the code-example substitution.

naomiblack:
merged.

PR #282, opened 2015-10-17
refactor(checkUnbalancedTicks): split out convertBackticksToCodeBlocks to maintain cohesion

petebacondarwin:
This improves upon changes made in 28a805733f6bd8b3a3394ea29c828c17b391c855
as part of #277


petebacondarwin:
@jtraband and @naomiblack - this is a better way of structuring the code change that was merged yesterday. It keeps the processors cohesive and focussed only on one task.
I added some tests for both processors for good measure, which could be expanded upon if anyone had the time.

petebacondarwin:
Landed as 3baf6c0561356e6388780d3cb220e85e1bdca3ac

PR #280, opened 2015-10-16
(docs) Wrote and added glossary

wardbell:

PR #284, opened 2015-10-17
(docs) added devguide template syntax

wardbell:


wardbell:
This includes "acceptable" stub pages for docs that are referenced by Template Syntax and the Glossary. These stub pages will be made real by AC.

This seems also to include the commit for the Glossary which I submitted as PR #280. I guess that wasn't accepted. It will be by incorporation of this PR.  I'd like that to land because there are a backlog of fixes to it, mostly from Pascal

PR #290, opened 2015-10-18
Fix jasmine-core paths in  jasmine-testing-101

Foxandxss:
If you want to use `jasmine-core` instead of `jasmine`, the paths are different.


Foxandxss:
There is another file on the guide with the same issue. Updated it as well.

jimthedev:
:+1: This is a really simple fix that needs merged so that the testing story can be actually useful.

wardbell:
Agreed. What was I thinking/doing?  Please accept

naomiblack:
thank you!

PR #296, opened 2015-10-19
hides api sections when there are no items to display

robertmesserle:
@naomiblack


robertmesserle:
Added spacing fix for buttons to war

PR #297, opened 2015-10-19
Update to Dart QS to alpha.44, move _examples dir

kwalrath:


kwalrath:
@naomiblack this brings the Dart quick start up to alpha 44 and updates the sample locations.

PR #300, opened 2015-10-19
docs(di): fixes and updates

PascalPrecht:
/cc @wardbell @naomiblack


public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -302,10 +302,10 @@ include ../../../../_includes/_util-fns
   Mission accomplished!  We don't even need the Angular Dependency Injection system to test the `HeroesComponent`.
   We simply create a bew `HeroesComponent` with a mock service and poke at it:
```
wardbell: Please change "bew" to "new" while you are here.

PR #320, opened 2015-10-29
fix spelling errors in the case sensitive environment like linux

simonkuang:
building passed in centos 7 hosted on a virtualbox vm


naomiblack:
thank you!

PR #335, opened 2015-11-02
docs(dependency-injection.jade): fix a typo

shairez:


wardbell:
Good catch. Should accept PR

PR #337, opened 2015-11-03
(docs) Angular 2 isn't ES6, it's TypeScript

timruffles:
Angular 2 isn't just ES6 + types and annotations, so this will confuse people.

As an example, here's a file using two huge features that aren't part of ES6 classes - [object properties and privacy](https://github.com/jeffmo/es-class-static-properties-and-fields).

### DIFF
--- a/public/features.jade
+++ b/public/features.jade
@@ -13,17 +13,17 @@
       material design UI components with responsive, cross-device support.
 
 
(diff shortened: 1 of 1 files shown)


wardbell:
I agree and I believe this is non-controversial.

Issue #342, opened 2015-11-03
Missing angular2 repo while running serve-and-sync


filipesilva:
Trying to run the `serve-and-sync` task after the steps in `Development Setup` outputs the following error:

```
[14:49:58] Error: build-api-docs task requires the angular2 repo to be at E:/sandbox/angular
    at Gulp.<anonymous> (E:\sandbox\angular.io\gulpfile.js:102:11)
    at module.exports (E:\sandbox\angular.io\node_modules\gulp\node_modules\orchestrator\lib\runTask.js:34:7)
    at Gulp.Orchestrator._runTask (E:\sandbox\angular.io\node_modules\gulp\node_modules\orchestrator\index.js:273:3)
    at Gulp.Orchestrator._runStep (E:\sandbox\angular.io\node_modules\gulp\node_modules\orchestrator\index.js:214:10)
    at E:\sandbox\angular.io\node_modules\gulp\node_modules\orchestrator\index.js:279:18
    at finish (E:\sandbox\angular.io\node_modules\gulp\node_modules\orchestrator\lib\runTask.js:21:8)
    at E:\sandbox\angular.io\node_modules\gulp\node_modules\orchestrator\lib\runTask.js:43:4
    at _fulfilled (E:\sandbox\angular.io\node_modules\dgeni\node_modules\q\q.js:798:54)
    at self.promiseDispatch.done (E:\sandbox\angular.io\node_modules\dgeni\node_modules\q\q.js:827:30)
    at Promise.promise.promiseDispatch (E:\sandbox\angular.io\node_modules\dgeni\node_modules\q\q.js:760:13)
```

Cloning https://github.com/angular/angular one level above solves this, but it is not documented in the readme.

wardbell:
My bad in my instructions. 

That gulp task rebuilds everything ... including the API docs which require access to the angular2 source code repo. The task assumes/requires that a clone of that repo reside under that name in a sibling folder to your angular.io clone.

You don't have to clone one level above (which would bring down a ton of irrelevant stuff). Just clone angular into a neighboring folder once and carry on.

Don't have a better way right now. Welcome suggestions.

filipesilva:
Ah that is what I meant, a sibling directory. I should have been more clear. In the PR, I word it as `Clone this repo and [angular](https://github.com/angular/angular) on the same parent directory`.

wardbell:
[...]

PR #343, opened 2015-11-03
docs(readme): clarify dev setup

### LINKED ISSUE #341: Missing serve-and-watch gulp task
Running `gulp serve-and-watch` as indicated in step `Development setup with watches` of readme gives the following output:

```
$ gulp serve-and-watch
[14:47:46] Using gulpfile E:\sandbox\angular.io\gulpfile.js
[14:47:46] Task 'serve-and-watch' is not in your gulpfile
[14:47:46] Please check the documentation for proper gulpfile formatting
```

The task doesn't seem to be in `gulpfile.js` anymore, as far as I can tell it was replaced by `build-and-serve` in https://github.com/angular/angular.io/commit/680c9740f2e73a91f3dafdabd585fc707b08fb7c. 

However, this one also runs browserSync, so the only difference between this task and `serve-and-sync` is that the latter also watches the ng2 source.

Maybe the `Development setup with watches` section should be removed and just retain `Development setup with watches and browser reload`?

filipesilva:
Close #341, #342

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,20 +8,15 @@ Angular.io is currently the preview site for Angular 2. This site also includes 
 
 ## Development Setup
 1. install [nvm](https://www.npmjs.com/package/nvm)
-2. cd into root directory `angular.io/`
-3. make sure you are using the latest node and npm by running `nvm use 4`.
-3. install local packages by running `npm install`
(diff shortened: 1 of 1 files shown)


wardbell:
The revision in this commit is still not as clear as it *could* and **should** be but it is correct and much better than what we have now. Therefore, I'm accepting it.

PR #350, opened 2015-11-05
Added Docs/DART forms example

tomplusplus:
WIP


tomplusplus:
I signed it!

kwalrath:
I've seen this complaint from googlebot before. I see that you *have* signed the CLA, so we needn't worry about it. (It'd be nice, but not crucial, to figure out how to avoid it in the future.)

public/docs/_examples/forms/dart/lib/hero_form_component.dart
```diff
@@ -0,0 +1,27 @@
...
+const List<String> _powers = const [
+  'Really Smart',
+  'Super Flexible',
+  'Super Hot',
+  'Weather Changer'
+];
+
+@Component(selector: 'hero-form')
+@View(
+    templateUrl: '../web/hero_form_component.html',
```
kwalrath: The ../web seems suspect. @kevmoo, how should it be done?

public/docs/_examples/forms/dart/lib/hero_form_component.dart
```diff
@@ -0,0 +1,27 @@
...
+    directives: const [CORE_DIRECTIVES, FORM_DIRECTIVES])
+class HeroFormComponent {
+  List<String> get powers => _powers;
+
+  bool submitted = false;
+
+  Hero model = new Hero(18, 'Dr IQ', _powers[0], alterEgo: 'Chuck Overstreet');
+
+  onSubmit() {
+    this.submitted = true;
```
kwalrath: delete `this.`

public/docs/_examples/forms/dart/web/hero_form_component.html
```diff
@@ -0,0 +1,42 @@
+<div class="container" [hidden]="submitted">
```
kwalrath: @kevmoo, shouldn't this file be under lib?

public/docs/_examples/forms/dart/web/main.dart
```diff
@@ -0,0 +1,6 @@
+import 'package:angular2/bootstrap.dart';
```
kwalrath: @yjbanov should we be using `show` in our imports? (That's not Dart style, but it seemed to be Angular style. Do we need to comply?)

public/docs/_examples/forms/dart/lib/hero.dart
```diff
@@ -0,0 +1,19 @@
...
+
+class Hero {
+  num number;
+  String name;
+  String power;
+  String alterEgo;
+
+  Hero(this.number, this.name, this.power, {this.alterEgo});
+
+  Map toJson() {
```
kwalrath: Could this be just `=> JSON.encode(this)`? See the [library tour's dart:convert coverage](https://www.dartlang.org/docs/dart-up-and-running/ch03.html#dartconvert---decoding-and-encoding-json-utf-8-and-more) for details.

kwalrath: Never mind, this isn't a Map.

kevmoo: Nope! JSON.encode is not that fancy.

It should use the arrow, though 

```dart
Map toJson() => {'number: number ...}
```

Note that `this` is not needed

public/docs/_examples/forms/dart/lib/hero_form_component.dart
kevmoo: Yeah, no. Lib should NEVER reference files outside lib. `hero_form_component.html` can go in the lib dir, though – at least it should

public/docs/_examples/forms/dart/web/hero_form_component.html
kevmoo: Yup

public/docs/_examples/forms/dart/web/main.dart
kevmoo: I think Angular has to because of typescript weirdness – not required at all unless you're trying to avoid a conflict

public/docs/_examples/forms/dart/lib/hero_form_component.dart
tomplusplus: Good to know. I'll update it and remember that for the future.

public/docs/_examples/forms/dart/lib/hero.dart
tomplusplus: I'll add the arrow. It's not needed for the final code output, but in the TS example they add it for debugging, then remove towards the end. I wasn't how one to one the Dart docs are going to be, so I put it there just in case.

tomplusplus:
I'm looking into the CLA issue. Im not sure why it's flagging as a different email that the commit author. They appear to be the same.

tomplusplus:
@kwalrath , @kevmoo  - Made the recommended updates. It looks like the bootstrap styles are no longer  there since we removed the dependency. I added a reference to the cdn. The TS example is using `npm` to bring it in. What is best practice for Dart ?

public/docs/_examples/forms/dart/lib/hero.dart
```diff
@@ -0,0 +1,13 @@
+library hero_form.hero;
+
+class Hero {
+  num number;
```
kevmoo: Make this an `int` me thinks – doubt this is meant to be a floating point number

public/docs/_examples/forms/dart/web/main.dart
yjbanov: Yeah, we prefer using them in TypeScript, but it's only a matter of style. No pressure to use them in Dart. Especially not on `bootstrap.dart`, which exports only one function `bootstrap`.

kwalrath:
@kevmoo do you have an answer for Tom's bootstrap CDN question?

kevmoo:
Hrm...I'm on the fence. For now I'd say link to a public CDN.

Let's keep the number of files down – the angular folks can chime in later if they have a preference either way.

CC @kwalrath

wardbell:
We are using npm instead of cdn in TS for at least two reasons

1. Ensures can run even the network is poor or unavailable.

1. We will write tests (unit and e2e) that may run in environment without cdn access.

I personally find it confusing to have mix of local and network file requirements.

Your call; that's the "Why" in TS land

kevmoo:
Thanks, @wardbell 

I guess we can fall back to the "disk is cheap" argument, too.

Tom: let's copy in the bootstrap bits required.

tomplusplus:
@kevmoo  Sounds good. Should I go back to adding bootstrap a dependency in `pubspec.yaml` or just download and manually add the files.

kevmoo:
@tomplusplus Dart doesn't have a 1-to-1 mapping with npm packages. Just copy bootstrap.min.css into the web directory and link to it directly.

Also: please rebase on origin/master down to one commit.

kevmoo:
Otherwise – LGTM :smile:

tomplusplus:
@kevmoo OK bootstrap is copied directly in and I rebased and squashed all commits down to one.

Thanks

kwalrath:
It looks like this is ready to go. Correct? If so, @tomplusplus please remove the [WIP] and I'll merge it (*not* using the big green button). The commit message should match the [angular commit message guidelines](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit), but  I can take care of that.

PR #352, opened 2015-11-06
docs(quickstart-ts): add note about ES6 shim

filipesilva:
Adding appendix to TS quickstart regarding ES6 support and how to use a shim to get the example working on older browsers.


wardbell:
Thanks for adding this.

PR #357, opened 2015-11-07
api-builder: add target package

petebacondarwin:
The `target` package allows us to filter blocks of content in our API docs by target language.

This PR implements this package also updates the gulp API build process.
* the `build-api-docs` task now runs two other tasks `build-js-api-docs` and `build-ts-api-docs`
* these two sub-tasks, which can be run independently, will generate API docs from the TS source code and output into the `docs/js/latest/api` and `docs/ts/latest/api` folders respectively.

The target package contains a new start-stop inline type definition of the form `{@target ts}...{@endtarget}`. The target can take one or more "targets" - in our case it is going to be either `ts` or `js`. If the target of a block doesn't match the current target environment for the dgeni run, the block is excluded.

For example you could change the comment in https://github.com/angular/angular/blob/2.0.0-alpha.45/modules/angular2/src/core/metadata.ts#L266-L330 to be:

```js
/**
 * {@link ViewMetadata} factory for creating annotations, decorators or DSL.
 *
 * {@target ts}
 * ### Example as TypeScript Decorator
 *
 * ```
 * import {Component, View} from "angular2/angular2";
 *
 * @Component({...})
 * @View({...})
 * class MyComponent {
 *   constructor() {
 *     ...
 *   }
 * }
 * ```
 * {@endtarget}
 * {@target js}
 * ### Example as ES5 DSL
 *
 * ```
 * var MyComponent = ng
 *   .Component({...})
 *   .View({...})
 *   .Class({
 *     constructor: function() {
 *       ...
 *     }
 *   })
 * ```
 * {@endtarget}
 * {@target js}
 * ### Example as ES5 annotation
 *
 * ```
 * var MyComponent = function() {
 *   ...
 * };
 *
 * MyComponent.annotations = [
 *   new ng.Component({...}),
 *   new ng.View({...})
 * ]
 * ```
 * {@endtarget}
 */
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "canonical-path": "0.0.2",
     "del": "^1.2.0",
     "dgeni": "^0.4.0",
-    "dgeni-packages": "^0.11.0",
+    "dgeni-packages": "^0.11.1",
     "diff": "^2.1.3",
     "fs-extra": "^0.24.0",
     "glob": "^5.0.14",
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -11,6 +11,7 @@ var Q = require("q");
 var delPromise =  Q.denodeify(del);
 var Minimatch = require("minimatch").Minimatch;
 var Dgeni = require('dgeni');
+var Package = require('dgeni').Package;
 var fsExtra = require('fs-extra');
 var fs = fsExtra;
 var exec = require('child_process').exec;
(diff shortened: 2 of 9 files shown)


petebacondarwin:
@naomiblack, @btford and @IgorMinar can you take a look at this and give some feedback?

package.json
```diff
@@ -30,7 +30,7 @@
     "canonical-path": "0.0.2",
     "del": "^1.2.0",
     "dgeni": "^0.4.0",
-    "dgeni-packages": "^0.11.0",
+    "dgeni-packages": "^0.11.1",
```
petebacondarwin: there was a bug in the new inline tag def processor that we need for the target package and is fixed in 0.11.1

IgorMinar:
can we have a predefined list of targets? if someone accidentally uses an unknown target, we should throw and print the list of allowed targets.

otherwise lgtm

petebacondarwin:
Done and landed as b68ef96

PR #361, opened 2015-11-09
@View decorator removed to be consistent with other examples

tapas4java:
@View decorator removed to be consistent with other examples available in angular.io site


tapas4java:
I signed it!

wardbell:
We'll take it. Thanks.

wardbell:
done

PR #371, opened 2015-11-13
chore(gulpfile): refine the tasks for watching and synching

petebacondarwin:


wardbell:
Waiting to hear from you regarding the fact that `node_modules` in the `_examples` folder are not ignored. I think we wanted that to be part of this batch of changes.

petebacondarwin:
Will take a look over the weekend

PR #375, opened 2015-11-14
fix(tutorial): Updated link to displaying data

brandonroberts:
Fixed link to displaying data in template.

Closes https://github.com/angular/angular/issues/5279


wardbell:
Brandon - I incorporated your fix within a bunch of link fixes in both ToH 1 and ToH 2.

Thanks for digging this up and fixing it. We appreciate it!

PR #380, opened 2015-11-17
[WIP] docs(samples): add Dart version of user-input example

tomplusplus:
@kevmoo , @kwalrath . Please Review. This is pretty much a direct copy from what exist in the jade files except with the most current version of angular.

Thanks.

### DIFF
--- a/public/docs/_examples/user-input/dart/lib/todo_list.dart
+++ b/public/docs/_examples/user-input/dart/lib/todo_list.dart
@@ -1,0 +1,41 @@
+library user_input.todo_list;
+
+import 'dart:html';
+
+import 'package:angular2/angular2.dart';
+
+@Component(selector: 'todo-list')
+@View(
+    template: '''
+<input #todotext (keyup)="doneTyping(\$event)">
+<button (click)="addTodo(todotext.value)">Add Todo</button>
+<ul>
+   <li *ng-for="#todo of todos">
+     {{ todo }}
+   </li>
+</ul>
+''',
+    directives: const [NgFor])
+class TodoList {
+  List<String> todos = [
+    'Eat breakfast',
+    'Walk dog',
+    'Breathe',
+    'Learn Angular'
+  ];
+
+  addTodo(String todo) {
+    window.console.log(todos.indexOf(todo));
+    if (todos.indexOf(todo) < 0) {
+      todos.add(todo);
+    }
+  }
+
+  doneTyping(KeyboardEvent event) {
+    if (event.keyCode == KeyCode.ENTER) {
+      InputElement e = event.target;
+      addTodo(e.value);
+      e.value = null;
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


public/docs/_examples/user-input/dart/lib/todo_list.dart
```diff
@@ -0,0 +1,41 @@
...
+class TodoList {
+  List<String> todos = [
+    'Eat breakfast',
+    'Walk dog',
+    'Breathe',
+    'Learn Angular'
+  ];
+
+  addTodo(String todo) {
+    window.console.log(todos.indexOf(todo));
```
kwalrath: Remove this line. (If we kept it, it would be "print" instead of window.console.log().)

kwalrath:
lgtm, after you fix/remove window.console.log. (If you've used that in any other samples, we should change it to print().)

kwalrath:
One more thing I forgot: Please add a comment with the text `#docregion` at the top of every file, using whatever the comment syntax is for that type of file. For example, in a Dart file:

```
// #docregion
...
```

More details at https://angular.io/docs/styleguide.html. (I'll take care of the forms examples, since I'm working on that guide right now.)

Could you also remove any filenames you see at the tops of files? Those aren't necessary any more with the new styles.

Thanks!!

public/docs/_examples/user-input/dart/lib/todo_list.dart
```diff
@@ -0,0 +1,41 @@
+library user_input.todo_list;
+
+import 'dart:html';
+
+import 'package:angular2/angular2.dart';
+
+@Component(selector: 'todo-list')
+@View(
```
kwalrath: I think you can drop @View and put the template in @Component.

public/docs/_examples/user-input/dart/lib/todo_list.dart
tomplusplus: Oh yeah - that snuck in there I meant to remove it.

public/docs/_examples/user-input/dart/lib/todo_list.dart
tomplusplus: Ok good to know. I've seen both ways through out other examples so i wasn't sure if we wanted to unify on one way over the other for all examples, or if we wanted to show different ways in different parts.

kwalrath: Yep, it started out with @View being required, but now @View shouldn't be required unless you have multiple views (which might not even be possible yet).

tomplusplus: I'm using @View in the other samples (pipes and forms) that has already been merged. Should i update those and do a new PR ?

kwalrath: I'll fix forms, but do please fix pipes. Thanks!

PR #195, opened 2015-08-21
Rename viewInjector to viewBindings in the developer guide

stevenroose:
This refactor has been made in the code, but was forgotten in the dev guide.

Fixes angular/angular#3781

PR #381, opened 2015-11-17
docs(samples): add Dart version of displaying-data example

tomplusplus:
@kevmoo , @kwalrath . Please review. This is the Displaying Data sample from the jade files with the current version of angular2 alpha46, and updates `viewBindings` to view `viewProviders` as per [this comment](https://github.com/angular/angular.io/pull/195#issuecomment-156787119).

Thanks.


public/docs/_examples/displaying-data/dart/lib/show_properties.dart
```diff
@@ -0,0 +1,41 @@
+// lib/show_properties.dart
+library displaying_data.show_properties;
+
+import 'dart:async';
+
+import 'package:angular2/angular2.dart';
+import 'package:displaying_data/friends_service.dart';
+
+@Component(selector: 'display', viewProviders: const [FriendsService])
+@View(
```
kwalrath: I think you can drop the @View and just specify the template in the @Component.

PR #386, opened 2015-11-18
mkdir error for OSX/Linux

devNoiseConsulting:
You need to include a "-p" argument to the mkdir command. Since the 'src' doesn't exist yet, the mkdir will return an error instead of creating the 2 directories.


devNoiseConsulting:
I signed it!

wardbell:
Closed by 412d3b5e59f7ccafcf901a17faac85838ab5748e

PR #391, opened 2015-11-21
docs: Quickstart bash command needs -p flag

estsauver:
Hi Angular, 

It's my first PR to angular, and I hope I'm following the contribution guidelines.

I noticed the `-p` flag was missing, I think this is the right place to change it?

CLA is signed.


johnapost-eb:
It looks like #386 already addresses this.

johnapost:
It looks like #386 already addresses this.

estsauver:
You are right!

PR #399, opened 2015-11-22
docs: Change label of case to 'kebab' for attrs

estsauver:
The docs refer to `this-style-of-casing` as snake case, but it's
actually "kebab case." Snake case looks `like_this_instead`.

https://en.wikipedia.org/wiki/Letter_case#Special_case_styles


wardbell:
Cute. Why not?

PR #405, opened 2015-11-25
docs: Remove accidental elvis operator

estsauver:
In the documentation, this section is discussing alternatives to
the elvis operator. This particular section is supposed to be
showing an alternative work around to the elvis operator, so
this error is especially confusing.

(https://angular.io/docs/ts/latest/guide/template-syntax.html#expression-ops)


wardbell:
Good catch.

PR #406, opened 2015-11-25
docs(hierarchical-injector): improve chapter

cburgdorf:
Hey @wardbell,

here is the awaited PR. Not only does that bring the proper runnable example code it also replaces all the old fashioned code with more modern one using `providers: [...]`, `@Input()`, `@Output()` and also dropping `@View()`.

Maybe we can get this PR in before we look at the other rough spots of the chapter?

/cc @PascalPrecht


wardbell:
Good idea. Done.

PR #407, opened 2015-11-25
Updated code example to build array of Hero

bsimpson:
Previously the instructions indicated that this builds an array of Hero instances:

```
var HEROES: Hero[] = [
      { "id": 11, "name": "Mr. Nice" },
      { "id": 12, "name": "Narco" },
      { "id": 13, "name": "Bombasto" },
      { "id": 14, "name": "Celeritas" },
      { "id": 15, "name": "Magneta" },
      { "id": 16, "name": "RubberMan" },
      { "id": 17, "name": "Dynama" },
      { "id": 18, "name": "Dr IQ" },
      { "id": 19, "name": "Magma" },
      { "id": 20, "name": "Tornado" }
    ]
```

```
console.log(HEROES[0]) # => Object {...}
```

This actually builds an array of Object instances as we are not instantiating `Hero`. This is confusing as a user would think they can add a function to the `Hero` class and access it via an instance.

In order to instantiate an array of Hero instances, we need to map over the array and instantiate `Hero` for each entity:

```
var HEROES: Hero[] = [
      { "id": 11, "name": "Mr. Nice" },
      { "id": 12, "name": "Narco" },
      { "id": 13, "name": "Bombasto" },
      { "id": 14, "name": "Celeritas" },
      { "id": 15, "name": "Magneta" },
      { "id": 16, "name": "RubberMan" },
      { "id": 17, "name": "Dynama" },
      { "id": 18, "name": "Dr IQ" },
      { "id": 19, "name": "Magma" },
      { "id": 20, "name": "Tornado" }
    ].map(function(heroAttribute) {
      var hero: Hero;
      hero = new Hero();
      hero.id = heroAttribute.id;
      hero.name = heroAttribute.name;
      return hero;
    });
```

Now we have an array of Hero entities:

```
console.log(HEROES[0]) # => Hero {...}
```

### DIFF
--- a/public/docs/ts/latest/tutorial/toh-pt2.jade
+++ b/public/docs/ts/latest/tutorial/toh-pt2.jade
@@ -60,10 +60,17 @@ include ../../../../_includes/_util-fns
       { "id": 18, "name": "Dr IQ" },
       { "id": 19, "name": "Magma" },
       { "id": 20, "name": "Tornado" }
-    ];
+    ].map(function(heroAttribute) {
+      var hero: Hero;
+      hero = new Hero();
+      hero.id = heroAttribute.id;
+      hero.name = heroAttribute.name;
+      return hero;
+    });
     ```
     The `HEROES` array is of type `Hero`.
     We are taking advantage of the `Hero` class we coded previously to create an array of our heroes.
+    By mapping over our array of objects and instantiating a new class `Hero` and setting the attributes, we have have an array of heroes (`HEROES: Hero[]`)
     We aspire to get this list of heroes from a web service, but let’s take small steps
     on this road and start by displaying these mock heroes in the browser.
 


bsimpson:
I signed it!

wardbell:
You are right. However, This isn't an important point for the tour ... more of a distraction really ... we want to keep the reader's eyes on the prize. My instinct is that your solution adds complexity but not much Angular 2 insight. Perhaps we'd better off saying ...

>We've created an array of Heroes that are structural equivalent to an instance of the `Hero` class which is good enough for our tour.

Or perhaps we should give the `Hero` class a ctor and rewrite this to instantiate `Hero` instances?

I want to think about it. I'm leaving this open while I do. 

That said, thanks for your close reading of the text and thoughtful comment. We need that kind of help.

Keep your eyes on this and see how we decide.

And keep those suggestions coming.

bsimpson:
Understandable @wardbell . It doesn't add to the Angular 2 tutorial. That being said, if we aren't using the `class Hero`, I don't think we need it at all? I'm reasonably sure I can delete this class definition and the code works the same. At no point do we initialize a `Hero` instance?

filipesilva:
Having the type there (even if it's not a real instance of `Hero`) helps typescript infer a few things and provide better tooling. 

IIRC as far as typescript tooling is concerned, if you say it's a `Hero` and the object has all the required fields of a `Hero`, typescript will do all typechecks as if it was a hero. And if it doesn't have all the required fields, it'll warn you as well.

This is what I get in Atom if I try to say put a fake hero object that doesn't have an `id` into what I say is a `Hero` array.

image

wardbell:
Also ... you've only seen the first two chapters of the tutorial.  The `Hero` class becomes more important as the app evolves. 

That you haven't seen the rest of the tutorial is **entirely my fault**. @johnpapa  wrote it; it's up to me to get it published ... and I have been tardy in doing so.  Intend to fix that soon.

bsimpson:
Ah, good to know. Thanks again for looking this PR over. I'm happy to close
if we decide the instantiation of Hero objects is not the critical path.

PR #409, opened 2015-11-25
docs(guide): add Dart forms guide

kwalrath:
I don't intend to merge this until next week, but do let me know if you find any problems. In particular, I'm wondering if my approach to the examples (including a copy of every single version) is OK. If not, it'll require some code contortions that'll result in a Frankenstein example that doesn't work correctly.

To see the text diffs from the TS version, go to https://github.com/kwalrath/angular.io/commit/084e4a40a8fe581cccf613c06df5761fae5db63a#diff-80199a70f2fd234f9b3fb0dade4031b0.


kwalrath:
@wardbell Is it OK to have many copies of the sample? I'll ask some other questions in comments on the author guide.

wardbell:
I'd prefer that we do not do that. It's pretty confusing. It's also a bear to maintain.

Didn't I avoid that in the TS version?

I have finessed this elsewhere by creating a single sample with a shell page, router, and links to navigate among the variations.  Maybe we should talk about this first?

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+:marked
+  Here at the *Hero Employment Agency* we use this form to maintain personal information about the
+  heroes in our stable. Every hero needs a job. It's our company mission to match the right hero with the right crisis!
+
+  Two of the three fields on this form are required. Required fields have a green bar on the left to make them easy to spot.
+
+  If we delete the hero name, the form displays a validation error in an attention-grabbing style:
+
+figure.image-display
+  img(src="/resources/images/devguide/forms/hf-2.png" alt="Invalid, Name Required")
```
wardbell: These form images have always felt to big to me.  After some experimentation, I think 400px works.

Suggest you do as I am doing and replacing `.png" alt`  with `.png" width="400px" alt`   or some such thing.

Review to make sure you like the result and adjust as necessary

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+  We can't lay out the form until we know what the model looks like.
+
+  A model can be as simple as a "property bag" that holds facts about a thing of application importance.
+  That describes well our `Hero` class with its three required fields (`id`, `name`, `power`)
+  and one optional field (`alterEgo`).
+
+  In the `lib` directory, add a file called `hero.dart`
+  with the following code:
+
++makeExample('forms/dart/angular2_forms-2-COMPONENT/lib/hero.dart', 'all', 'lib/hero.dart')
```
wardbell: Good idea adding the file name to the example title bar

I notice you do this elsewhere when showing a fragment of a file. I find this confusing ... because it isn't the whole file.

One way to be clear would be to add the suffix " (excerpt)" at the end as in "web/index.html (excerpt)"

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+
+:marked
+  It's an anemic model with few requirements and no behavior. Perfect for our demo.
+
+  Here's an example of creating and using a Hero object:
+
++makeExample('forms/dart/angular2_forms-2-COMPONENT/lib/hero.dart', 'newhero')(format=".")
+
+:marked
+  The `alterEgo` is optional, so the constructor lets us omit it: note the
```
wardbell: I found it confusing to read this after the sample.  So I revised the TS version to pull this up.  Mine now reads:

```
:marked
  It's an anemic model with few requirements and no behavior. Perfect for our demo.

  The TypeScript compiler generates a public field for each `public` constructor parameter and
  assigns the parameter’s value to that field automatically when we create new heroes.
  
  The `alterEgo` is optional and the constructor lets us omit it; note the (?) in `alterEgo?`.

  We can create a new hero like this:
```
Then the example follows ... and this sentence is no longer needed.

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+  with the following code:
+
++makeExample('forms/dart/angular2_forms-2-COMPONENT/lib/hero.dart', 'all', 'lib/hero.dart')
+
+:marked
+  It's an anemic model with few requirements and no behavior. Perfect for our demo.
+
+  Here's an example of creating and using a Hero object:
+
++makeExample('forms/dart/angular2_forms-2-COMPONENT/lib/hero.dart', 'newhero')(format=".")
```
wardbell: By adding `(format=".")` you eliminated the line numbers ... and also the horizontal scrollbar. The example can't be read when the browser is made narrow (or on mobile).

In general, I remove line numbers only for one-liners that are short (and I try to keep my lines short ... not always with success).

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+
++makeExample('forms/dart/angular2_forms-2-COMPONENT/lib/hero_form_component.dart', null, 'lib/hero_form_component.dart')
+
+:marked
+  There’s nothing special about this component, nothing to distinguish it from any component we've written before,
+  nothing form-specific about it ... except, perhaps, the tell-tale `FORM_DIRECTIVES` import.
+
+  Understanding this component requires only the Angular 2 concepts covered in previous guides.
+
+  1. The code imports a standard set of symbols from the Angular library.
```
wardbell: We're going to have to sweep ALL documents and pull out every place that we add `FORM_DIRECTIVES` and `CORE_DIRECTIVES`.  A PITA for us but good for developers.

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+  Down the road, we can inject a data service to get and save real data
+  or perhaps expose these properties as inputs and outputs
+  <!--TODO: link to (./template-syntax.html#inputs-outputs) --> for binding to a
+  parent component. None of this concerns us now, and these future changes won't affect our form.
+
+  1. We threw in a `diagnostic` property to return a
+  string describing our model.
+  It'll help us see what we're doing during our development; we've left ourselves a cleanup note to discard it later.
+
+  Why isn't the template inline in the component file?
```
wardbell: Better than the original tortured sentence. I adopted something like it

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+
+.callout.is-important
+  header Angular forms do not require a style library
+  :marked
+    Angular makes no use of the `container`, `form-group`, `form-control`, and `btn` classes or
+    the styles of any external library. Angular apps can use any CSS library
+    ... or none at all.
+
+:marked
+  Let's add the stylesheet.
```
wardbell: Great improvement. Took it.

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+.l-main-section
+:marked
+  ## Two-way data binding with ***ng-model**
+  Running the app right now would be disappointing.
+
+figure.image-display
+  img(src="/resources/images/devguide/forms/hf-3.png" alt="Early form with no binding")
+:marked
+  The problem is that we aren't binding to the `Hero` yet.
+  We know how to do that from earlier guides.
```
wardbell: I prefer "chapters" to "guides" ... as we discussed.

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+
+.l-main-section
+:marked
+  ## Two-way data binding with ***ng-model**
+  Running the app right now would be disappointing.
+
+figure.image-display
+  img(src="/resources/images/devguide/forms/hf-3.png" alt="Early form with no binding")
+:marked
+  The problem is that we aren't binding to the `Hero` yet.
```
wardbell: You did well to replace my sentence. But you didn't state the problem.

I changed this sentence in the TS to: "We don't see hero data because we are not binding to the `Hero` yet."

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+  We know how to do that from earlier guides.
+  [Displaying Data](./displaying-data.html) demonstrated
```
wardbell: You improved this section but I still find it clumsy.

I updated the TS version to the following (call them chapters rather than guide as we discussed):
```
  We don't see hero data because we are not binding to the `Hero` yet.
  We know how to do that from earlier chapters.
  [Displaying Data](./displaying-data.html) taught us Property Binding.
  [User Input](./user-input.html) showed us how to listen for DOM events with an
  Event Binding and how to extract values from the screen.
```

public/docs/dart/latest/guide/forms.jade
wardbell: You improved this section but I still find it clumsy.

I updated the TS version to the following (calling them chapters rather than guide as we discussed):
```
  We don't see hero data because we are not binding to the `Hero` yet.
  We know how to do that from earlier chapters.
  [Displaying Data](./displaying-data.html) taught us Property Binding.
  [User Input](./user-input.html) showed us how to listen for DOM events with an
  Event Binding and how to extract values from the screen.
```

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+figure.image-display
+  img(src="/resources/images/devguide/forms/ng-model-in-action.png" alt="ng-model in action")
+:marked
+  The diagnostic is evidence that values really are flowing from the input box to the model and
+  back again. **That's two-way data binding!**
+
+  Let's add similar `[(ng-model)]` bindings to Alter Ego and Hero Power.
+  We'll ditch the input box binding message
+  and add a new binding at the top to the component's `diagnostic` property.
+  Then we can confirm two-way data binding *to the entire Hero model*.
```
wardbell: Better. But better still might be:
```
Then we can confirm that two-way data binding works *for the entire Hero model*.
```

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+
+  After revision, the core of the form should have three `[(ng-model)]` bindings that
+  look much like this:
+
++makeExample('forms/dart/angular2_forms-4-DIAGNOSTIC/lib/hero_form_component.html', 'ng-model-2', 'lib/hero_form_component.html')(format="linenums:4")
+
+:marked
+  If we ran the app right now and changed all the values, it might look like this:
+figure.image-display
```
wardbell: How did you get such a different image from my TS version? Can you prove that it displays the way you're showing it in DART? It doesn't look like that for me in plunker.

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+  img(src="images/ng-model-in-action-2.png" alt="ng-model in super action")
+:marked
+  The diagnostic near the top of the form
+  confirms that our changes to the values are reflected in the model.
+
```
wardbell: Better. I changed the TS version of this segment to:
```
:marked
  If we ran the app right now and changed every Hero model property, the form might display like this:
figure.image-display
  img(src="/resources/images/devguide/forms/ng-model-in-action-2.png" width="400px" alt="ng-model in super action")
:marked
  The diagnostic near the top of the form
  confirms that all of our changes are reflected in the model.
```

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+  ** We're done with the diagnostic binding. Delete it now.**
+
+
+.alert.is-helpful
```
wardbell: Delete this alert. It is no longer meaningful.

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+  :marked
+    Although `NgModel` is officially a form directive,
+    we can use `[(ng-model)]` and two-way binding outside of forms too.
+
+.l-sub-section
+  h3 Inside [(ng-model)]
+  :marked
+    *This section is an optional deep dive into [(ng-model)]. Not interested? Skip ahead!*
```
wardbell: I like!  Stole it.

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+  A form isn't just about data binding. We'd also like to know the state of the controls on our form.
+  The `NgControl` directive keeps track of control state for us.
+
+.callout.is-helpful
+  header NgControl requires Form
+  :marked
+    The `NgControl` is one of a family of `NgForm` directives that can only be applied to
+    a control within a `<form`> tag.
+:marked
+  Our application can ask an `NgControl` instance whether
```
wardbell: Better!

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
++makeExample('forms/dart/angular2_forms-6-NGCONTROL/lib/hero_form_component.html', 'ng-control-1', 'lib/hero_form_component.html')(format="linenums:6")
+:marked
+  Be sure to assign a unique name to each `ng-control` directive.
+
+.l-sub-section
+  :marked
+    Angular registers controls under their `ng-control` names
+    with the `NgForm` directive.
+    We didn't add the `NgForm` directive explicitly but it's here;
+    we'll talk it [later in this guide](#ng-form).
```
wardbell: Missing the word "about". Change to `we'll talk about it`

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+
+
+.l-main-section
+  h2 Add custom CSS for visual feedback
+
+  p.
+    <code>NgControl</code> doesn't just track state. It updates the control to
+    have three classes that describe the control's state.
+
+  table
```
wardbell: NICE!

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+:marked
+  Let's add a temporary local template variable
+  <!-- TODO: make that a link to (./template-syntax.html#local-vars) -->
+  named **spy**
+  to the Name `<input>` tag and use the spy to display those classes.
+
++makeExample('forms/dart/angular2_forms-7.1-CSS-SPY/lib/hero_form_component.html', 'ng-control-2')(format="linenums:6")
+
+:marked
+  Now let's run the app, focusing our attention on the Name input box
```
wardbell: Better. Better still?
```
  Now run the app, focusing the *Name* input box.
  Follow the next four steps *precisely*
```

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+figure.image-display
+  img(src="/resources/images/devguide/forms/ng-control-class-changes.png" alt="Invalid Form")
+
+:marked
+  The (`ng-valid` | `ng-invalid`) pair are the most interesting to us, because we want to send a
+  strong visual signal when the values are bad. We also want to mark required fields.
+
+  We can do both at the same time with a colored bar on the left of the input box:
+
+figure.image-display
```
wardbell: I am adding an animated gif in my TS PR that may help make this clearer.

When I do, you should be able to use this:
```
  The actions and effects are as follows:
figure.image-display
  img(src="/resources/images/devguide/forms/control-state-transitions-anim.gif"  alt="Control State Transition")
:marked
  We should have seen the following four sets of class names and their transitions:
figure.image-display
  img(src="/resources/images/devguide/forms/ng-control-class-changes.png" width="400px" alt="Control State Transitions")
```

public/docs/dart/latest/guide/forms.jade
```diff
@@ -0,0 +1,662 @@
...
+
+  Angular recognizes that syntax and sets the `name` variable
+  to the `Control` object identified by the `ng-control` directive that,
+  not coincidentally, we called "name".
+
+  We bind the `Control` object's `valid` property to the element's `hidden` property.
+  While the control is valid, the message is hidden;
+  if it becomes invalid, the message is revealed.
+<a id="ng-form"></a>
+.l-sub-section
```
wardbell: I suggest that you replaces this sub-section with the following
```
.l-sub-section
  :marked
    ### The NgForm directive
    Recall from the previous section that `ng-control` registered this input box with the
    `NgForm` directive as "name".

    We didn't add the **[`NgForm`](../api/core/NgForm-class.html) directive** explicitly.
    Angular added it surreptiously, wrapping it around the `<form>` element

    The `NgForm` directive  supplements the `form` element with additional features.
    It collects `Controls` (elements identified by an `ng-control` directive)
    and monitors their properties including their validity.
    It has its own `valid` property which is true only if every contained
    control is valid.

    In this example, we are pulling the "name" control out of its `controls` collection
    and assigning it to the template local variable so that we can
    access the control's properties  &mdash; such as the control's own `valid` property.
```

public/docs/dart/latest/guide/forms.jade
kwalrath: I only used line numbers when it was code I wanted them to actually type in.

(Come to think of it, that's part of the reason why I was keeping so many copies of the app around—so I'd know the line numbers for each sample. Of course, with `#docregion` you still have to do some subtracting to get the real line # that they'd see.)

public/docs/dart/latest/guide/forms.jade
kwalrath: When it's a fragment I specify line #s that don't start at 1 (assuming that I do it right, and that the fragment isn't from the beginning of the file).

kwalrath: To be clear, I'm not opposed to " (excerpt)". I just happened to convey the info another way, and I didn't think of changing the text at the top.

public/docs/dart/latest/guide/forms.jade
wardbell: I saw that.  I'm worried for you on two counts
1) Those numbers are killer hard to get right in the first place (what with #docregions) and almost impossible to maintain.

2) Colored dark red against black. they're barely legible anyway.

I fear a ton of effort for paltry reward. YMMV.

public/docs/dart/latest/guide/forms.jade
kwalrath: I like the result. Thanks!

public/docs/dart/latest/guide/forms.jade
kwalrath: Added " (excerpt)" to all titles of incomplete files. I'll need to take a closer look before I remove the line #s. (I think that having line #s start at 1 is misleading for excerpts that aren't at the start of the file.)

public/docs/dart/latest/guide/forms.jade
kwalrath: I like it. Done.

public/docs/dart/latest/guide/forms.jade
kwalrath: It doesn't work in Dart in alpha46. @vsavkin is apparently working on it. I'll leave a note to remove this.

public/docs/dart/latest/guide/forms.jade
kwalrath: I like it. Done.

(Do we really need to capitalize Property Binding and Event Binding? I saw some places where it wasn't capitalized... and I liked that.)

public/docs/dart/latest/guide/forms.jade
kwalrath: Better. Done.

public/docs/dart/latest/guide/forms.jade
kwalrath: The toString() method that I added to Hero produces this output. It's a custom string. Overriding toString() is a really common thing to do in Dart. The alternative (which we originally did) was to add and call a toJson() method, which was fine but produced output that's less readable and less Dart-y. It didn't seem like JSON was important here—just easy to produce—so I used the string instead.

public/docs/dart/latest/guide/forms.jade
kwalrath: I like this better (although I changed _display_ to _look_ because _display like this_ sounded weird to me; _display something like this_ might be more correct, but then it's getting wordy).

public/docs/dart/latest/guide/forms.jade
kwalrath: Done.

public/docs/dart/latest/guide/forms.jade
kwalrath: Done

public/docs/dart/latest/guide/forms.jade
kwalrath: I do like it better, although I worry that "focus" could be misconstrued as setting the place where keyboard input goes. How about:

  Now run the app, and look at the Name input box.
  Follow the next four steps *precisely*:

Also... do we need to italicize UI names? I was relying on capitalization alone to set them apart. (I once had feedback that many people have trouble reading italics, so I avoid them whenever possible. People seem to read boldface more easily, but I try to use it sparingly since it attracts so much attention.)

public/docs/dart/latest/guide/forms.jade
kwalrath: After trying this out, I think we don't need both the image and the animation, but I'm not sure which to drop. (Animation could be annoying; images take up more space.) One option is to do the following, so that people who can't see the animation/image for some reason (maybe they're blind or have processing issues) can get the info:

```
  Now run the app, and look at the Name input box.
  Follow the next four steps *precisely*:

  1. Look but don't touch.
  1. Click inside the name box, then click outside it.
  1. Add slashes to the end of the name.
  1. Erase the name.

  The classes are displayed as follows:

  1. `form-control ng-untouched ng-pristine ng-valid` (initial state)
  1. `form-control ng-pristine ng-valid ng-touched` (after clicking)
  1. `form-control ng-valid ng-touched ng-dirty` (after changing)
  1. `form-control ng-touched ng-dirty ng-invalid` (after erasing)

figure.image-display
  img(src="/resources/images/devguide/forms/control-state-transitions-anim.gif"  alt="Control State Transition")
```

public/docs/dart/latest/guide/forms.jade
kwalrath: I did this even before looking at this note! However... I have a few minor differences in bolding and capitalization. You'll see them when I post my changes...

kwalrath:
@wardbell: This might be ready to go.

wardbell:
Hi Kathy. Are you expecting me to do something here? I don't feel competent to merge DART content

kwalrath:
I'm happy to merge it if you're happy (enough) with the text style. The content is just a Darty version of the TS content, so I believe it's much better than nothing. I suppose I could ask @yjbanov to review it, but he's probably very busy, so I don't think that's a showstopper.

wardbell:
I thought what I read before was fine and now you've improved it. I say "go for it!"

kwalrath:
Great, thanks!​

PR #417, opened 2015-11-30
(docs) devguide intro

wardbell:
Replaces the [Developers Guide trampoline page](https://angular.io/docs/ts/latest/guide/) with real content that (a) describes the guide and how to read it AND (b) provides an architectural overview of Angular 2's essential building blocks.

Of course such an overview strongly reflects one particular perspective. Is that perspective sufficiently consistent with the Angular Team's to warrant publication?  Let's find out.


public/docs/ts/latest/guide/index.jade
```diff
@@ -1,10 +1,501 @@
...
++makeExample('intro/ts/app/hero-detail.component.html')(format=".")
+:marked
+  In two-way binding, a data property value flows to the input box from the component as with property binding.
+  The user's changes also flow back to the component, resetting the property to the latest value,
+  as with event binding.
+  
+  Angular processes *all* data bindings once per JavaScript event cycle, 
+  depth-first from the root of the application component tree. 
+figure
+  img(src="/resources/images/devguide/intro/component-databinding.png" alt="Data Binding" style="float:left; width:300px; margin-left:-40px;margin-right:10px" )
```
PascalPrecht: I really like those graphics. This one here shows the connection between a component an the outer template where the component is used. It can be a bit confusing, because a component can have its own template/view and it's not entirely clear what template we have in this figure.

Maybe we should make clear that we have outer/inner template? "component template" doesn't work because everything is a component after all...

public/docs/ts/latest/guide/index.jade
```diff
@@ -1,10 +1,501 @@
...
+:marked
+  A **Component** controls a patch of screen real estate that we could call a "view". 
+  The shell at the application root with navigation links, that list of heroes, the hero editor ... 
+  they're all views controlled by Components.
+  
+  We define a Component's "application logic" - what it does to support the view - inside a class.
+  The class interacts with the view through an API of properties and methods. 
+  
+  <a id="component-code"></a>
+  A `HeroListComponent`, for example, might have a `heroes` property that returns an array of heroes
```
PascalPrecht: Because I'm good a bike shedding: shouldn't this be `HeroesListComponent`?

public/docs/ts/latest/guide/index.jade
```diff
@@ -1,10 +1,501 @@
...
++makeExample('intro/ts/app/hero-list.component.ts', 'class')
+:marked
+  Angular creates, updates, and destroys components as the user moves through the application.
+  The developer can take action at each moment in this lifecycle through optional [Lifecycle Hooks](#).
+.l-sub-section
+  :marked
+    Our example shows the service arriving via the class constructor.
+    Who calls that constructor? Who provides the service?
+    For the moment, have faith that Angular is going to deliver an
+    appropriate `HeroService` when we need it.
```
PascalPrecht: Love this paragraph. Good job

public/docs/ts/latest/guide/index.jade
```diff
@@ -1,10 +1,501 @@
...
+:marked
+  We recognize `<h2>` and  `<div>`. 
+  But there's other markup that no one told us about in school.
+  What is`*ng-for`, <code>{&zwnj;{hero.name}}</code>, `(click)`, `[hero]`, and `<hero-detail>`?
+  These are examples of Angular's [template syntax](template-syntax.html). 
+  We will grow accustomed to that syntax and may even learn to love it.
+  We'll begin to explain it in a moment.
+  
+  Before we do, focus attention on the last line. 
+  The `<hero-detail>` tag is a custom element representing the `HeroDetailComponent`.
```
PascalPrecht: Not entirely sure if we should use `custom element` as a term here. Ppl might think that Angular uses the Custom Elements technology to create directives, which is not the case.

Maybe we can just stick with `custom component/directive` ? Maybe it's not even a problem at all...

public/docs/ts/latest/guide/index.jade
```diff
@@ -1,10 +1,501 @@
...
+  
+  * `selector` - a css selector that tells Angular to create and insert an instance of this component 
+  where it finds a `<hero-list>` tag in *parent* HTML. 
+  If the template of the application shell (a Component) contained
+<div style="margin-left:30px">
+code-example(language="html").
+    &lt;hero-list>&lt;/hero-list>
+</div>  
+:marked
+  >Angular would replace that tag with an instance of the `HeroListComponent` view.
```
PascalPrecht: I'd propose to change or remove this sentence, since Angular 2 **does not** replace a tag with a component's template. Due to Shadow DOM (Emulation), a component always needs a host element, which is the one we're specifying with `selector`. AFAIK, there isn't even a way to replace it.

This is also why the `replace` property in the DDO in A1 is marked as deprecated.

public/docs/ts/latest/guide/index.jade
```diff
@@ -1,10 +1,501 @@
...
+  to the component class definition. Angular discovers this metadata at runtime and thus knows how to do "the right thing".
+  
+  The template, metadata, and component together describe the view.
+
+  We apply other metadata decorators in a similar fashion to guide Angular behavior. 
+  The `@Injectable`, `@Input`, `@Output`, `@RouterConfig` are a few of the more popular decorators
+  we'll master as our Angular knowledge grows.
+<br clear="all">
+:marked
+  The architectural take-away is that we must add metadata to our JavaScript code
```
PascalPrecht: Maybe we should remove "JavaScript" here and just talk about adding metadata to code? Technically JavaScript doesn't support decorators, not sure if this would confuse ppl. Because on the other hand, we don't mention anything about ES5 code until this point.

I think it's totally legit to say sth. like "decorators allow us to add metadata to our code".

public/docs/ts/latest/guide/index.jade
```diff
@@ -1,10 +1,501 @@
...
+  so that Angular knows what to do.
+
+.l-main-section
+<a id="data-binding"></a>
+:marked
+  ## Data Binding
+  Without a framework, we would be responsible for pushing data values into the HTML controls and turning user responses
+  into actions and value updates. Writing such push/pull logic by hand is tedious, error-prone and a nightmare to
+  read as the experienced jQuery programmer can attest
```
PascalPrecht: Haha, you made a funny here. Very nice!

public/docs/ts/latest/guide/index.jade
PascalPrecht: Actually, in the context where this figure is used, it is sort of explained. You can ignore the comment above, sorry for this.

public/docs/ts/latest/guide/index.jade
```diff
@@ -1,10 +1,501 @@
...
+:marked
+  ## The Directive
+figure
+  img(src="/resources/images/devguide/intro/directive.png" alt="Parent child" style="float:left; width:150px; margin-left:-40px;margin-right:10px" )
+:marked
+  Our Angular templates are *dynamic*. When Angular renders them, it transforms the DOM 
+  according to the instructions given by a **Directive**.
+  
+  A directive is a class with directive metadata. In TypeScript we'd apply the `@Directive` decorator
+  to attache metadata to the class.
```
PascalPrecht: attache -> attach

public/docs/ts/latest/guide/index.jade
```diff
@@ -1,10 +1,501 @@
...
+  
+  **Structural** directives alter layout by adding, removing, and replacing elements in DOM.
+  
+  We see two built-in structural directives at play in our [example](#template) template:
++makeExample('intro/ts/app/hero-list.component.html', 'structural')(format=".")
+:marked
+  * [`*ng-for`](displaying-data.html#ng-for) tells Angular to stamp out one `<div>` per hero in the `heroes` list.
+  * [`*ng-if`](displaying-data.html#ng-if) includes the `HeroDetail` component only if a selected hero exists.
+  
+  **Attribute** directives alter the appearance or behavior of a particular element.
```
PascalPrecht: Here I also like to point out that attribute directives are applied to **existing** elements. It's a little word that has a massive impact on how the user understands the difference between the directive types.

PascalPrecht: Here I also like to point out that attribute directives are applied to **existing** elements. It's a little word that has a massive impact on how the user understands the difference between the directive types.

public/docs/ts/latest/guide/index.jade
```diff
@@ -1,10 +1,501 @@
...
++makeExample('intro/ts/app/hero-list.component.ts', 'providers')
+:marked
+  ... in which case we get a new instance of the
+  service with each new instance of that component.
+  
+  We've vastly over-simplified dependency injection for this overview.
+  We can learn the full story in the [Dependency Injection](dependency-injection.html) chapter.
+  
+  The points to remember are:
+  * dependency injection is wired into the framework and used everywhere.<br><br>
```
PascalPrecht: I'd also mention that it can be used as a standalone system and integrates with other frameworks. One nice thing about the new DI is that it *is* usable standalone, without being wired into the framework.

public/docs/ts/latest/guide/index.jade
```diff
@@ -1,10 +1,501 @@
...
+  * the `Injector` is the main mechanism.
+    * an injector maintains a *container* of services that it created.
+    * an injector can create a new service instance from a *provider*.
+  * a *provider* is a recipe for creating a service.
+ 
+  * we register *providers* with injectors.
```
PascalPrecht: How about "We configure injectors with providers" ?

public/docs/ts/latest/guide/index.jade
```diff
@@ -1,10 +1,501 @@
...
+  1. [Directive](#directive)
+  1. [Service](#service)
+  1. [Dependency Injection](#dependency-injection)
+  
+  That's a foundation for everything else in an Angular application
+  and it's more than enough to get going.
+  But it doesn't include everything we'll need or want to know.
+   
+  Below is a brief, alphabetical list of other important Angular features and services. 
+  Most of them are covered in this Developers Guide (or soon will be):
```
PascalPrecht: Developers Guide -> Developer Guide ?

PascalPrecht:
@wardbell GREAT WORK!

It's a very very nice overview, thanks for writing this down. I left some comments, feel free to comment/ignore/make use of them.

public/docs/ts/latest/guide/index.jade
filipesilva: Yeah I also spent a while re-reading this one.

filipesilva:
@wardbell I unfortunately pushed a commit with comments while @PascalPrecht was doing his PR comments, so some of his were marked as outdated.

PascalPrecht:
Actually, **committing** comments into the PR feels kind of backwards. We have github comments for this, and I don't think that this is very practical. Let's talk about this.

filipesilva:
To be perfectly honest, I didn't even notice that this PR was from `IdeaBlade` directly into `angular`. My bad.

public/docs/ts/latest/guide/index.jade
wardbell: It's writer's choice: singular or plural.  We're allowed to say "HeroList" to describe a list of heroes; the "list" part conveys the plurality.  "HeroesList" is equally correct.  I'm keeping it ;-)

public/docs/ts/latest/guide/index.jade
wardbell: I struggled with this.  Of course it isn't an element at all. But I'm targeting the novice and I felt that imprecision was acceptable in the service of learning.

It will be interesting to see what others say.

public/docs/ts/latest/guide/index.jade
wardbell: Good catch!  Now says "Angular inserts an instance of the `HeroListComponent` view between those tags."

public/docs/ts/latest/guide/index.jade
wardbell: Agree.  

I deliberately did NOT say "add decorators ...". The fact is that we "must add metadata to our ... code" no matter what kind of code (DART, JS, TS, ES2015) it is.

public/docs/ts/latest/guide/index.jade
wardbell: Good point. Done.

public/docs/ts/latest/guide/index.jade
wardbell: Nah!  Trying to cover the mainstream of DI for the novice A2 developer. Your observation would be a good one for the DI chapter itself.

public/docs/ts/latest/guide/index.jade
wardbell: Reminds me that the current title is "Developers Guides" which is totally confusing!

PascalPrecht:
@wardbell maybe you can update the commit message to be sth. like: "docs(intro): introduce intro chapter"?

If we want to follow the Angular commit message conventions, it has to be `type(component): subject`. Where `component` can be optional.

wardbell:
Too late. I've been following my own wacky convention (which makes it "not a convention").  Maybe I'll get it right next time.

public/docs/ts/latest/guide/index.jade
PascalPrecht: It is at least an `HTMLUnknownElement`, which ofc, starters don't have to care about either. I just don't want them to confuse this with actual `CustomElements`.

wardbell: I floated the term "Component Element" on slack. I think it is good for noobs and pros.  We'll see if anyone complains.

PR #418, opened 2015-11-30
chore(doc-gen): update modules to include in API docs

petebacondarwin:
The public exported modules have been reorganized

### DIFF
--- a/public/resources/js/controllers/app-controller.js
+++ b/public/resources/js/controllers/app-controller.js
@@ -60,11 +60,14 @@ angularIO.controller('AppCtrl', ['$mdDialog', '$timeout', '$http', '$sce', funct
   };
 
   vm.apiSections = [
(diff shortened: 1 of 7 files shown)


naomiblack:
I just tried this but when I run `gulp build-api-docs` I only see core, router, and test in the API doc export. All the directives, for example, are still missing. app-data.json doesn't show these missing libs.

Missing:
- common
- animate
- instrumentation
- compiler
- platform/browser
- platform/http
- platform/router

naomiblack:
Sorry -- pulled the wrong PR. Trying again :)

naomiblack:
I've merged this to master and pushed to the live site even though it's unfinished because it's still much better than the broken content we were showing. Will file issues for the remaining problems (@petebacondarwin please add any I missed, or just fix them!)

PR #434, opened 2015-12-02
[WIP] Dart - dev guide - displaying data example

tomplusplus:
@kevmoo @kwalrath  - Please review 
This is the conversion of the new TS displaying-data example

### DIFF
--- a/public/docs/_examples/displaying-data/dart/lib/app_component_1.dart
+++ b/public/docs/_examples/displaying-data/dart/lib/app_component_1.dart
@@ -5,11 +5,13 @@ import 'package:angular2/angular2.dart';
 
 @Component(
     selector: 'my-app',
+// #docregion template
     template: '''
 <h1>{{title}}</h1>
-<h2>My favorite hero is: {{myHero}}</h2>''')
+<h2>My favorite hero is: {{myHero}}</h2>'''
+// #enddocregion template
+    )
 class AppComponent {
   String title = 'Tour of Heroes';
   String myHero = 'Windstorm';
 }
-// #enddocregion
--- a/public/docs/_examples/displaying-data/dart/lib/app_component_1.dart
+++ b/public/docs/_examples/displaying-data/dart/lib/app_component_1.dart
@@ -1,0 +1,15 @@
+// #docregion
+library displaying_data.app_component_1;
+
+import 'package:angular2/angular2.dart';
+
+@Component(
+    selector: 'my-app',
+    template: '''
+<h1>{{title}}</h1>
+<h2>My favorite hero is: {{myHero}}</h2>''')
+class AppComponent {
+  String title = 'Tour of Heroes';
+  String myHero = 'Windstorm';
+}
+// #enddocregion
--- a/public/docs/_examples/displaying-data/dart/lib/app_component_2.dart
+++ b/public/docs/_examples/displaying-data/dart/lib/app_component_2.dart
@@ -1,0 +1,30 @@
+// #docregion
+library displaying_data.app_component_2;
+
+import 'package:angular2/angular2.dart';
+
+const List<String> _heroes = const [
+  'Windstorm',
+  'Bombasto',
+  'Magneta',
+  'Tornado'
+];
+
+@Component(
+    selector: 'my-app',
+    template: '''
+<h1>{{title}}</h1>
+<h2>My favorite hero is: {{myHero}}</h2>
+<p>Heroes:</p>
+<ul>
+  <li *ng-for="#hero of heroes">
+    {{ hero }}
+  </li>
+</ul>''',
+    directives: const [NgFor])
+class AppComponent {
+  String title = 'Tour of Heroes';
+  List<String> heroes = _heroes;
+  String myHero = _heroes[0];
+}
+// #enddocregion
--- a/public/docs/_examples/displaying-data/dart/lib/app_component_2.dart
+++ b/public/docs/_examples/displaying-data/dart/lib/app_component_2.dart
@@ -1,17 +1,14 @@
 // #docregion
 library displaying_data.app_component_2;
 
+// #docregion imports
 import 'package:angular2/angular2.dart';
 
-const List<String> _heroes = const [
-  'Windstorm',
-  'Bombasto',
-  'Magneta',
-  'Tornado'
-];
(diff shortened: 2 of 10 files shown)


public/docs/_examples/displaying-data/dart/lib/app_component_1.dart
```diff
@@ -0,0 +1,15 @@
...
+@Component(
+    selector: 'my-app',
+    template: '''
+<h1>{{title}}</h1>
+<h2>My favorite hero is: {{myHero}}</h2>''')
+class AppComponent {
+  String title = 'Tour of Heroes';
+  String myHero = 'Windstorm';
+}
+// #enddocregion
```
kwalrath: I never bothered with `#enddocregion` since the EOF is implicitly the end. Does the original have this comment? (Just curious; I don't think it matters.)

public/docs/_examples/displaying-data/dart/lib/app_component_2.dart
```diff
@@ -0,0 +1,30 @@
...
+    template: '''
+<h1>{{title}}</h1>
+<h2>My favorite hero is: {{myHero}}</h2>
+<p>Heroes:</p>
+<ul>
+  <li *ng-for="#hero of heroes">
+    {{ hero }}
+  </li>
+</ul>''',
+    directives: const [NgFor])
```
kwalrath: Soon, apparently, we won't need to declare these built-in directives. But not yet. (I think it works in TS as of 46 but not yet Dart, even as of 47.)

public/docs/_examples/displaying-data/dart/web/main.dart
```diff
@@ -0,0 +1,16 @@
+// #docregion
+import 'package:angular2/bootstrap.dart';
```
kwalrath: Nice. I just did some similar multiple-version work on the Forms example, but I just created copies of the HTML file and didn't make it easy to switch between them. I like this v1/v2/v3/v4 approach. I suppose the only downside is when, even when the HTML is the only thing that changes, you'd have to create extra Dart files to go along with them.

public/docs/_examples/displaying-data/dart/web/main.dart
```diff
@@ -0,0 +1,16 @@
...
+import 'package:displaying_data/app_component_2.dart' as v2;
+import 'package:displaying_data/app_component_3.dart' as v3;
+import 'package:displaying_data/app_component_final.dart' as v4;
+
+main() {
+//  bootstrap(v1.AppComponent);
+//  bootstrap(v2.AppComponent);
+//  bootstrap(v3.AppComponent);
+  bootstrap(v4.AppComponent);
+  bootstrap(AppCtorComponent);
```
kwalrath: Dart usually avoids abbreviations in names. @kevmoo do you think it matters here?

public/docs/_examples/displaying-data/dart/lib/app_component_1.dart
tomplusplus: Im not sure. I just noticed it with named regions. It's implicit already I wont worry about it either.

public/docs/_examples/displaying-data/dart/lib/app_component_2.dart
tomplusplus: That's good news. It seemed really odd to have to in the first place. I thought maybe it was for better performance somehow.

public/docs/_examples/displaying-data/dart/web/main.dart
tomplusplus: That's a good point. But I agree with you. For the use case of teaching dev's angular2/Dart, I think this approach does feel easier to follow when looking at the examples directly.

public/docs/_examples/displaying-data/dart/web/main.dart
tomplusplus: So `v1` should be `version1'  etc... ?

kwalrath: Sorry, I should've been more specific. v1, v2, etc. seemed fine to me, since v is a very common abbreviation.

It's Ctor that we don't usually see in Dart. Maybe we should change it to "Creator" or something other than "Constructor"? I don't want it confused with a Dart constructor.

kevmoo: Looking at the TS example – I don't think we should even have the ctor thing in there – not sure what's in needed for.

https://github.com/angular/angular.io/blob/master/public/docs/_examples/displaying-data/ts/src/app/bootstrap.ts

bootstrap should only be called once anyway.

We should also comment out the imports of all of the components except the one being bootstrapped

tomplusplus: removing reference to ctor in `main.dart` but leaving the lib file just in case we want it later. Let me know if I should so otherwise.

tomplusplus:
ugh, I messed up squashing two commits. So its a bit funky. If everything looks good I'll rebase to master and squash everything into one commit.

kwalrath:
Looks good to me. (Wow, alpha48 already!)

public/docs/_examples/displaying-data/dart/lib/app_component_3.dart
```diff
@@ -0,0 +1,51 @@
...
+  new Hero(13, 'Bombasto'),
+  new Hero(15, 'Magneta'),
+  new Hero(20, 'Tornado')
+];
+
+// #enddocregion heroes
+// #docregion class
+class AppComponent {
+  String title = 'Tour of Heroes';
+  List<Hero> heroes = _heroes;
```
kevmoo: Use a property here `List<Hero> get heros => _heros;`

kevmoo:
...and now 53!

Ditto – looks good

kwalrath:
Looks like this languished a bit. Could you please check this against the current TS version and update it to beta 0? Thanks!

tomplusplus:
OK I updated it to beta.0 and made some minor changes to match the TS example. @kevmoo and @kwalrath, is it ready to remove [WIP] ?

kwalrath:
LGTM

kwalrath:
I'm going to merge this in, so I can update the chapter.

PR #435, opened 2015-12-02
chore(docs-shredder): fixes misspelled word

henryw4k:
'overriden' should be 'overridden'

http://dictionary.reference.com/browse/overridden


PascalPrecht:
Thanks! Bonus points if you could update the commit message to be sth. like `chore(docs-shredder): fixes misspelled word`

PR #444, opened 2015-12-04
docs: typo motiviate -> motivate

heminm:


wardbell:
We're doing some rework on these tutorial chapters. Holding off on accepting this one in case it becomes irrelevant due to a rewrite of that section. If it stays relevant, we'll merge it. Thanks for pointing it out.

wardbell:
We fixed in the forthcoming massive update to ToH.  Thanks for catching it.

PR #446, opened 2015-12-05
docs(developer-guides): incorrect QuickStart link in How to Read this…

denismosolov:
… Guide section

quickstart -> quickstart.html prevent 404 page


PascalPrecht:
@wardbell is this how links in the docs project work?

Otherwise LGTM!

wardbell:
Yup. My bad. Thanks for the catch.

I keep making this mistake because every other CMS I've ever worked with added '.html' by default <grumble>

PR #448, opened 2015-12-06
[WIP] (docs) Dart - dev guide/intro example

tomplusplus:
@kwalrath @kevmoo Please review.


public/docs/_examples/intro/dart/lib/hero_detail_component.dart
```diff
@@ -0,0 +1,10 @@
+library developer_guide_intro.hero_detail_component;
+
+import 'package:angular2/angular2.dart';
+import 'package:developer_guide_intro/hero.dart';
+
+@Component(selector: 'hero-detail', templateUrl: 'hero_detail_component.html')
```
kwalrath: No directives field is necessary? (I'm comparing this to https://github.com/tomplusplus/angular.io/blob/DeveloperGuides/intro/public/docs/_examples/intro/ts/src/app/hero-detail.component.ts)

kwalrath: Although, to be honest, this looks correct and the ts version looks weird to me.

public/docs/_examples/intro/dart/lib/hero_detail_component.html
```diff
@@ -0,0 +1,8 @@
+<!-- #docregion -->
+<hr>
+<h4>{{hero.name}} Detail</h4>
+<div>Id: {{hero.id}}</div>
+<div>Name:
```
kwalrath: The ts version has #docregion ng-model here

public/docs/_examples/intro/dart/lib/hero_list_component_1.html
```diff
@@ -0,0 +1,7 @@
+<!-- #docregion -->
```
kwalrath: The `#docregion`s are different in the ts version. I'm going to stop marking this, but it'd be great if this version had the same `#docregion`s as the ts version.

public/docs/_examples/intro/dart/lib/hero_list_component.html
```diff
@@ -0,0 +1,8 @@
+<!-- #docregion -->
+<h2>Hero List</h2>
+
+<div *ng-for="#hero of heroes" (click)="selectHero(hero)">
+    {{hero.name}}
```
kwalrath: We usually do 2-space indents, which is what the ts version uses.

public/docs/_examples/intro/dart/lib/hero_list_component.dart
```diff
@@ -0,0 +1,25 @@
...
+
+import 'package:angular2/angular2.dart';
+import 'package:developer_guide_intro/hero.dart';
+import 'package:developer_guide_intro/hero_detail_component.dart';
+import 'package:developer_guide_intro/hero_service.dart';
+
+@Component(
+    selector: 'hero-list',
+    templateUrl: 'hero_list_component.html',
+    directives: const [CORE_DIRECTIVES,HeroDetailComponent],
```
kwalrath: Be sure to run the formatter on this (and all the other .dart files). I'm pretty sure there should be a space after the comma.

public/docs/_examples/intro/dart/lib/hero_list_component.dart
```diff
@@ -0,0 +1,25 @@
...
+    providers: const [HeroService])
+class HeroListComponent {
+  List<Hero> heroes;
+  Hero selectedHero;
+
+  HeroListComponent(HeroService heroService){
```
kwalrath: I think there should be space before the { (but the formatter would know for sure).

public/docs/_examples/intro/dart/lib/hero_service.dart
```diff
@@ -0,0 +1,20 @@
...
+import 'package:developer_guide_intro/hero.dart';
+import 'package:developer_guide_intro/backend_service.dart';
+import 'package:developer_guide_intro/logger_service.dart';
+
+@Injectable()
+class HeroService {
+  var _backendService;
+  var _logger;
+  HeroService(){
+    _logger = new Logger();
```
kwalrath: Shouldn't Logger and BackendService be provided using arguments to the constructor, as in the ts version? So:

```HeroService(Logger this._logger, BackendService this._backendService) {}```

File: public/docs/_examples/intro/dart/lib/hero_service.dart [hunk 8]
```diff
@@ -0,0 +1,20 @@
...
+    _backendService = new BackendService();
+  }
+  getHeroes(){
```
kwalrath: Usually we'd have a return type, although it's not required. So:

```
List<Hero> getHeroes() {
  ...
}
```

public/docs/_examples/intro/dart/lib/hero_service.dart
```diff
@@ -0,0 +1,20 @@
...
+    List<Hero> heroes = _backendService.getAll(Hero);
```
kwalrath: Usually we wouldn't have a type here. We'd just have:

```
return _backendService.getAll();
```

(Does this code even run?)

public/docs/_examples/intro/dart/lib/logger_service.dart
```diff
@@ -0,0 +1,20 @@
+library developer_guide_intro.logger_service;
+
+import 'dart:html';
+
+import 'package:angular2/angular2.dart';
+
+@Injectable()
+class Logger {
```
kwalrath: I wonder if we could/should use https://pub.dartlang.org/packages/logging here. @kevmoo?

public/docs/_examples/intro/dart/lib/logger_service.dart
```diff
@@ -0,0 +1,20 @@
...
+  void log(dynamic msg) {
+    window.console.log(msg);
```
kwalrath: It looks like you could just use print() instead. Also, we don't usually use dynamic in signatures. Object would do.

```
void log(Object msg) {
  print(msg);
}

Comment (kwalrath) [inline public/docs/_examples/intro/dart/lib/logger_service.dart]: Another possibility (although some people object to using `=>` with methods that return void):

log(msg) => print(msg);
OR
void log(Object msg) => print(msg);

File: public/docs/_examples/intro/dart/lib/logger_service.dart [hunk 12]
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  void error(dynamic msg) {
```
kwalrath: Ah, now I see why you used window.console.log() above. Oh well.

public/docs/_examples/intro/dart/pubspec.yaml
```diff
@@ -0,0 +1,9 @@
+name: developer_guide_intro
+description: Developer Guide Intro
+version: 0.0.1
+dependencies:
+  angular2: 2.0.0-alpha.47
```
kwalrath: 47 -> 48

public/docs/_examples/intro/dart/web/index.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Intro to Angular 2</title>
```
kwalrath: Use 2-space indents and as few diffs from the ts version as possible.

kwalrath:
@kevmoo could you please take a look?

public/docs/_examples/intro/dart/lib/hero_detail_component.dart
tomplusplus: In the TS version the directive they are calling is itself. That didn't seem right me either.

public/docs/_examples/intro/dart/lib/hero_detail_component.html
tomplusplus: I did this before I got word back from you on the named `docregion`s. I'll go through all the current PR's I have out and update them.

public/docs/_examples/intro/dart/lib/hero_list_component_1.html
tomplusplus: I did this before I got word back from you on the named docregions. I'll go through all the current PR's I have out and update them. :)

public/docs/_examples/intro/dart/lib/logger_service.dart
tomplusplus: Yeah that was just to be consistent between the methods. I thought using `print` for the first one then `window.console.*` would seem a bit confusing. To be honest though Im not sure what the point of the logger service is even for. I don't see a reference to it anywhere. I wonder if it's in the works for one of the upcoming examples

kwalrath: It looks like that source code isn't in the page. I suspect the Logger is just an example of a Service. We can leave it as-is, I think.

public/docs/_examples/intro/dart/lib/hero_service.dart
tomplusplus: It does run, but I agree with you. The TS example had the extra `let heroes`. Sets it then returns it. I didn't see a reason for it but converted it over 'as is' just in case I was missing something. I'll update it.

tomplusplus: Oh ignore that last comment. The TS version is using the logger service here to log the length of the heroes returned form backend service. I must have left that part out on accident.

tomplusplus:
Updated. Waiting on to hear if we should use [https://pub.dartlang.org/packages/logging](https://pub.dartlang.org/packages/logging) v.s. the custom logger service as the TS example does.

kwalrath:
I don't think we should use the logging package. It's all hidden code (makes the example run, but isn't quoted on the page), so it doesn't matter what we use.

kwalrath:
@kevmoo any feedback before I take this code?

public/docs/_examples/intro/dart/lib/backend_service.dart
```diff
@@ -0,0 +1,25 @@
...
+library developer_guide_intro.backend_service;
+
+import 'package:angular2/angular2.dart';
+import 'package:developer_guide_intro/logger_service.dart';
+import 'package:developer_guide_intro/hero.dart';
+
+@Injectable()
+class BackendService {
+  Logger _logger;
+  getAll(type) {
```
kevmoo: return type `List`

public/docs/_examples/intro/dart/lib/backend_service.dart
```diff
@@ -0,0 +1,25 @@
+// #docregion
+library developer_guide_intro.backend_service;
+
+import 'package:angular2/angular2.dart';
+import 'package:developer_guide_intro/logger_service.dart';
+import 'package:developer_guide_intro/hero.dart';
+
+@Injectable()
+class BackendService {
+  Logger _logger;
```
kevmoo: `final Logger _logger;`

public/docs/_examples/intro/dart/lib/backend_service.dart
```diff
@@ -0,0 +1,25 @@
...
+    // TODO get from the database and return as a promise
+    if (type == Hero) {
+      return [
+        new Hero('Windstorm', power: 'Weather mastery'),
+        new Hero('Mr. Nice', power: 'Killing them with kindness'),
+        new Hero('Magneta', power: 'Manipulates metalic objects')
+      ];
+    }
+    _logger.error('Cannot get object of this type');
+    throw new Error();
```
kevmoo: `throw new ArgumentError(" ... with the log content ...");`

public/docs/_examples/intro/dart/lib/hero_service.dart
```diff
@@ -0,0 +1,20 @@
...
+
+import 'package:angular2/angular2.dart';
+import 'package:developer_guide_intro/hero.dart';
+import 'package:developer_guide_intro/backend_service.dart';
+import 'package:developer_guide_intro/logger_service.dart';
+
+// #docregion class
+@Injectable()
+class HeroService {
+  BackendService _backendService;
```
kevmoo: both of these can be final, right?

public/docs/_examples/intro/dart/lib/logger_service.dart
kevmoo: Tough one. I worry that adding in another package here would add complexity.

Might be worth mentioning it in a comment though?

kevmoo:
Added comments. Sorry for the lag here.

kwalrath:
@tomplusplus go ahead and make changes here. I'll merge them when you're ready, and then finish the docs.

kwalrath:
Closing this, as I've incorporated the files and kevmoo's comments into #460.

PR #449, opened 2015-12-07
[WIP] docs(structural-components): add chapter

wardbell:
The second half of Filipe's original "Structural Directives" chapter is here.

IMO (and Pascal's apparently), that half seemed to depart from the point of Structural Directives and,
in fact, defined a new **component** that did the job of incorporating external content by projection.

This seemed to me to be a new concept, deserving its own chapter. For lack of better terms, I dubbed it the "Structural Component".


wardbell:
Oops ... this belongs in the Ideablade/angular.io fork.  Closed.  Will resurface later when ready.

PR #450, opened 2015-12-07
Dev tooling 21 - symlink support and perf improvements

jtraband:
Adds abiiity to replace node_modules in _example subfolders with a shared symlink folder in _examples.
Fixes to other utilities to insure that they can work with symlinks.
Replaces 'globule' with 'globby' because of perf issues associated with node_modules.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -16,6 +16,7 @@ var fs = fsExtra;
 var exec = require('child_process').exec;
 var execPromise = Q.denodeify(exec);
 var prompt = require('prompt');
+var globby = require("globby");
 
 // TODO:
 //  1. Think about using runSequence
@@ -33,6 +34,7 @@ var LIVE_EXAMPLES_PATH = path.join(RESOURCES_PATH, 'live-examples');
 var docShredder = require(path.resolve(TOOLS_PATH, 'doc-shredder/doc-shredder'));
 var exampleZipper = require(path.resolve(TOOLS_PATH, '_example-zipper/exampleZipper'));
 var plunkerBuilder = require(path.resolve(TOOLS_PATH, 'plunker-builder/plunkerBuilder'));
+var fsUtils = require(path.resolve(TOOLS_PATH, 'fs-utils/fsUtils'));
 
 var _devguideShredOptions =  {
   examplesDir: path.join(DOCS_PATH, '_examples'),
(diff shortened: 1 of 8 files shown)


wardbell:
INSTRUCTIONS AND ADVICE FROM JAY IN THIS COMMENT
--

## add-examples-symlinks
`gulp add-example-symlinks` will walk every subdir of the ‘_examples’ folder and create a symlinked ‘node_modules’ folder anywhere it finds a ‘package.json’ file.  If there is already a node_modules folder in place, it and all of its children will be deleted before creating the new ‘node_modules’ folder.  The folder being ‘pointed to’ is the node_modules folder in the ‘_examples’ folder.  You will need to run ‘npm install’ in this folder at least once in order to have any node_modules to point to.  You should probably update the package.json in this directory to represent the versions of any dependencies that you want to have as the standard.  Once this is done you will never need to run ‘npm install’ in any of the subdirs, only in the _examples root. 

**Important: This gulp task requires admin access** on windows but you will probably only need to run it once. I’m not sure of the requirements on a unix system but the code is all standard node. 

That said, what is kind of neat is that if you really want a different set of dependencies for a specific _examples subdir, you can explicitly run ‘npm install’ in the specific folder.  This will in fact update the ‘node_modules’ in the _examples folder ( if you are linked) and your code will operate with the current directory’s package.json settings.  You can then later rerun ‘npm install’ in the _examples folder itself to get everything back to the way it was. 

## remove-examples-symlinks

`gulp remove-example-symlinks` removes all of the symlinks created by ‘add-example-symlinks’.  This task does not need admin access.  One side effect of running `add-examples-symlinks` followed by `remove-example-symlinks` is to remove all node_modules completely from the ‘_examples’ subtree. 

Note that neither of these tasks has any effect on the git repo because the only folders changed by any of this are all excluded ‘node_modules’ folder. 

A few other changes were needed to make this work.  First, as you noticed, when you have a lot of node_modules populated a number of our gulp tasks got very slow.  Once I ran the ‘add-example-symlinks’ I started running into this ( as well as crashing **WebStorm because it couldn’t handle the depth issue unless I told it to ignore node_modules**).  

>WebStorm users: remember or learn to exclude `node_modules`!

 I finally traced the  problem with the gulp tasks to the ‘globule’ npm package.  The real issue was that ‘globule’ was walking every file and directory in the _examples folder, including all of the nested node_module folders and then removing the ones with ‘node_modules’ in the path.  I found an alternative ‘globbing’ package (‘globby’) that has a much faster algorithm.   Some tasks went from minutes to seconds once I replaced all use of ‘globule’ with ‘globby’.  I made some of the ‘node_modules’ exclusions more explicit as well.

PR #451, opened 2015-12-07
[WIP] (docs) Dart Developer guides/attribute directives example

tomplusplus:
@kwalrath @kevmoo Please review.


kwalrath:
To make a variable private, you just name it with an underscore prefix. But it seems weird to use a constructor to initialize a class variable. Is it really an instance variable?

wardbell:
Actually, the leading underscore has no meaning to TS/JS at all. It is a convention that lets the readers (and the persons debugging the code) know that this variable is NOT part of the public API and may not be referenced externally (at your peril).

TS enforces the  `private` and `internal` access modifiers within the TS world; it has no runtime effect.

It is an instance variable, not a class variable.

tomplusplus:
Ah. Thanks @wardbell @kwalrath. I think see where my disconnect was. So if add `Renderer _renderer;` and `ElementRef _el;` then those will both be private, correct? 
Then the constructor can be `Highlight(ElementRef this._el, Renderer this._renderer);`

kwalrath:
Yep, that looks right. Dart style is to avoid abbreviations, so we'd probably use _element instead.

kwalrath:
PS to Ward: Dart has no private/internal access modifiers. Just using _ makes the variable library-private.

public/docs/_examples/attribute-directives/dart/lib/app_component.dart
```diff
@@ -0,0 +1,13 @@
+// #docregion
+library attribute_directives.app_component;
```
kwalrath: TypeScript's directives feature is called "annotations" in Dart. So...

attribute_directives -> attribute_annotations?

kwalrath: Gah, scratch that. I was getting "directive" confused with "decorator". Never mind...

public/docs/_examples/attribute-directives/dart/lib/app_component.html
```diff
@@ -0,0 +1,12 @@
+<!-- #docregion -->
+<h1>My First Attribute Directive</h1>
+<h4>Pick a highlight color</h4>
+<div>
+    <input type="radio" name="colors" (click)="color='lightgreen'">Green
+    <input type="radio" name="colors" (click)="color='yellow'">Yellow
+    <input type="radio" name="colors" (click)="color='cyan'">Cyan
+</div>
+<p><span [my-highlight]="color">Highlight me!</span></p>
+<p><span [my-highlight]="color" [default-color]="'violet'">
```
kwalrath: Does this really require a string in a string (`"'violet'"`)? That's kind of gross!

public/docs/_examples/attribute-directives/dart/web/index.html
```diff
@@ -0,0 +1,13 @@
+<!-- #docregion -->
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Attribute Directives</title>
+    <link rel="stylesheet" href="style.css">
+    <script async src="main.dart" type="application/dart"></script>
```
kwalrath: async -> defer
in these two lines.

public/docs/_examples/attribute-directives/dart/lib/app_component.html
tomplusplus: Yes it does. Anything between the double quotes is evaluated by the angular engine. If it's not in single quotes then it would look for a String object name violet

kwalrath:
LGTM. @kevmoo, any comments?

public/docs/_examples/attribute-directives/dart/lib/app_component.dart
```diff
@@ -0,0 +1,13 @@
...
+
+import 'package:angular2/angular2.dart';
+import 'package:attribute_directives/highlight_directive.dart';
+
+@Component(
+    selector: 'my-app',
+    templateUrl: 'app_component.html',
+    directives: const [Highlight])
+class AppComponent {
+  String color; //unlike TS or JS - Dart requires that the color property be added to the class object explicitly. Simply referencing it in the template will NOT dynamically add it.
```
kevmoo: Remove the comment – we don't need to highlight the diff w/ TS/JS

public/docs/_examples/attribute-directives/dart/lib/highlight_directive.dart
```diff
@@ -0,0 +1,57 @@
...
+// #enddocregion mouse-enter
+  onMouseLeave() {
+    _highlight(null);
+  }
+
+  void _highlight(String color) {
+    _renderer.setElementStyle(_element, 'background-color', color);
+  }
+
+  Highlight(ElementRef this._element, Renderer this._renderer);
```
kevmoo: Don't need to put types when using `this`

public/docs/_examples/attribute-directives/dart/lib/highlight_directive.dart
```diff
@@ -0,0 +1,57 @@
...
+  String _defaultColor = 'red';
+  // #enddocregion class-1
+  // #docregion defaultColor
+  @Input() set defaultColor(String colorName) {
+    _defaultColor = (colorName ?? _defaultColor);
+  }
+  // #enddocregion defaultColor
+// #docregion class-1
+
+  Renderer _renderer;
```
kevmoo: these both can be final

public/docs/_examples/attribute-directives/dart/lib/highlight_directive_2.dart
```diff
@@ -0,0 +1,36 @@
...
+    _highlight(null);
+  }
+  // #enddocregion mouse-methods
+
+  void _highlight(String color) {
+    _renderer.setElementStyle(_element, 'background-color', color);
+  }
+
+// #docregion ctor
+  Highlight(ElementRef this._element, Renderer this._renderer);
```
kevmoo: ditto from before

kevmoo:
See comments!

tomplusplus:
Thanks @kevmoo. I made those changes. @kwalrath updating to beta.1 broke this example. ElementRef says it no longer has a getter for 'style'. Also referenced [here](https://angular.io/docs/js/latest/api/core/ElementRef-class.html)  it suggest to use ElementRef with caution. It looks like they are still using it in TS though. Thoughts?

public/docs/_examples/attribute-directives/dart/lib/highlight_directive.dart
```diff
@@ -0,0 +1,55 @@
...
+// #docregion full
+// #docregion class-1
+// #docregion color
+  @Input('my-highlight') String highlightColor;
+// #enddocregion color
+
+  String _defaultColor = 'red';
+  // #enddocregion class-1
+  // #docregion defaultColor
+  @Input() set defaultColor(String colorName) {
```
zoechi: Wouldn't there shorthand functions be more appropriate (also below one-liners?

kevmoo: For setters, no – since `=` does return the value and `set` should be void.

zoechi: The return value is `null` anyway, but I know that setter are currently quite weird. Just wanted to get this topic some attention.

kwalrath:
@yjbanov do you know the answer to Tom's question?

> ...updating to beta.1 broke this example. ElementRef says it no longer has a getter for 'style'. Also referenced here it suggest to use ElementRef with caution. It looks like they are still using it in TS though.

public/docs/_examples/attribute-directives/dart/lib/highlight_directive.dart
```diff
@@ -0,0 +1,55 @@
...
+    _highlight(highlightColor ?? _defaultColor);
+  }
+
+// #enddocregion mouse-enter
+  onMouseLeave() {
+    _highlight(null);
+  }
+
+  void _highlight(String color) {
+    _renderer.setElementStyle(_element, 'background-color', color);
```
yjbanov: After a recent change this needs to be:

```dart
_renderer.setElementStyle(_element.nativeElement, 'background-color', color);
```

public/docs/_examples/attribute-directives/dart/lib/highlight_directive_1.dart
```diff
@@ -0,0 +1,12 @@
+// #docregion
+library attribute_directives.highlight_directive;
+
+import 'package:angular2/angular2.dart';
+
+@Directive(selector: '[my-highlight]')
+class Highlight {
+  Highlight(ElementRef element, Renderer renderer) {
+    //el.nativeElement.style.backgroundColor = 'yellow';
+    renderer.setElementStyle(element, 'background-color', 'yellow');
```
yjbanov: here too

public/docs/_examples/attribute-directives/dart/lib/highlight_directive_2.dart
```diff
@@ -0,0 +1,35 @@
...
+    _highlight("yellow");
+  }
+
+  onMouseLeave() {
+    _highlight(null);
+  }
+  // #enddocregion mouse-methods
+
+  void _highlight(String color) {
+    _renderer.setElementStyle(_element, 'background-color', color);
```
yjbanov: ...and here

yjbanov:
@kwalrath @tomplusplus 

You now need to pass `elementRef.nativeElement` to the `Renderer`. That's a recent change.

kwalrath:
Thanks very much, @yjbanov!

wardbell:
Ugh!  This is the kind of breaking change we were NOT supposed to see. It affects almost every attribute directive.  The API guide is no help on the subject as it describes the first parameter as  `renderElement: any` w/ no further comment.  

When we went to beta we promised people that we would be cautious about breaking changes. We're supposed to have a high bar for breaking changes.

**As a matter of policy, we should always update API docs when making a breaking change.**   I do not think that noting it in the CHANGES.MD is sufficient.

See issue I raised: https://github.com/angular/angular/issues/6627

wardbell:
### PLEASE DO NOT MERGE!

We cannot have samples running on different versions of A2! We have to all be on the same version across all samples and all language tracks.

I'm grateful that Tom discovered the issue though!

kwalrath:
I plan to convert this back to b0 and merge it, so we can have something to point people to.

PR #453, opened 2015-12-07
WIP: docs(*): upgrade code samples to 2.0.0-alpha-48

cironunes:
https://github.com/angular/angular-cli/issues/84

I'm not aware about the dart version so I didn't update the dart samples, but they probably should.

Should I replace `import 'package:angular2/angular2.dart';` for `import 'package:angular2/core.dart';`?


wardbell:
Thanks SO MUCH for your effort on this. I appreciate your eagerness to help.

As it happens, the alpha upgrade was just one among a larger number of upgrade steps that we've been taking. I will have to close your PR with a thank you.

We are changing these docs rapidly and there are big swings. I suggest leaving the docs alone for a week or so. Certainly don't do anything massive like this. Small typo fixes are welcome.

Best wishes, Ward

PR #456, opened 2015-12-08
docs(template-syntax): drive jade from sample code

wardbell:
Replaced all-text code with sample-driven examples.

Awaiting sanity check.  It should be ready to go.


PascalPrecht:
Quick note: the colon belongs after the parenthesis :)

wardbell:
Now THAT's feedback I can use!

PR #469, opened 2015-12-10
API Index page fix

petebacondarwin:
This PR gets dgeni to generate a different `api-list.json` file each target language, which is then displayed via a self-contained Angular directive.

Previously the code for filtering and so on was stored in the AppController and that controller always loaded the same data file on every page. This meant that we couldn't have a different API index for different languages but also that we were making an unnecessary server request on every page that was not the API index.


petebacondarwin:
This also fixes a bug where we were not displaying API docs that have docType `let`.

petebacondarwin:
Further, the barrels displayed are now entirely controlled by the top level modules that are listed in https://github.com/angular/angular.io/blob/master/tools/api-builder/angular.io-package/index.js.
So there is only one place to change when updating the barrels in the project.

petebacondarwin:
FAO: @naomiblack and @kwalrath

PR #496, opened 2015-12-14
fix(app): ensure headings are visible when navigating to in-page anchors

### LINKED ISSUE #477: Internal links hide header
Clicking internal links will hide the linked section header. This is a problem present throughout `angular.io` though, because there is no padding/margin to offset the navbar at the top of the main section.

So we get this:
image

While we should get this:
image

petebacondarwin:
Closes #477


petebacondarwin:
@filipesilva - here is a fix for #477. Can you try it out?

filipesilva:
Seems to be working fine here :D

PR #497, opened 2015-12-14
Sync API doc filtering with URL query

### LINKED ISSUE #490: Synchronize the API filtering and search with the URL query parameters
This will allow the user to use the back button from an API page to get back to the same filtered API list that they left before.

petebacondarwin:
This PR closes #490


IgorMinar:
looks reasonable to me

naomiblack:
for some reason this breaks the API docs page for me (I get no content) but I suspect that I did a bad rebase.

IgorMinar:
I rebased it here: #507 testing locally now

PR #498, opened 2015-12-14
chore(doc-gen): add createDecoratorDocs processor

petebacondarwin:
The createDecoratorDocs processor identifies docs of exports that actually
represent decorators and get useful information for them.

This commit also set up the rest of the generation to display these new
decorator docs with their own template and updates the API list so that
we can filter the docs by decorator docType.

### DIFF
--- a/public/resources/css/module/_api.scss
+++ b/public/resources/css/module/_api.scss
@@ -1,5 +1,9 @@
 @mixin icon {
   line-height: 14px;
+  &.decorator:before {
+    content: 'A';
+    background: $blueberry;
+  }
   &.directive:before {
     content: 'D';
     background: #db4437;
--- a/public/resources/js/directives/api-list.js
+++ b/public/resources/js/directives/api-list.js
@@ -27,9 +27,10 @@ angularIO.directive('apiList', function () {
 
       $ctrl.apiTypes = [
         { cssClass: 'directive', title: 'Directive', matches: ['directive'] },
+        { cssClass: 'decorator', title: 'Decorator', matches: ['decorator'] },
         { cssClass: 'class', title: 'Class', matches: ['class'] },
         { cssClass: 'interface', title: 'Interface', matches: ['interface'] },
-        { cssClass: 'function', title: 'function', matches: ['function'] },
+        { cssClass: 'function', title: 'Function', matches: ['function'] },
         { cssClass: 'const', title: 'Const or Enum', matches: ['const', 'enum'] },
         { cssClass: 'var', title: 'Variable', matches: ['var', 'let'] }
       ];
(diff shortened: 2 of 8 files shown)


IgorMinar:
@naomiblack might want to use a different icon but "A" is good enough for now.

screen shot 2015-12-14 at 8 10 42 am

and we will need to tweak the template (the export at the top is not very useful)

Otherwise this looks great!

naomiblack:
This isn't quite right. A few comments:

1) we specifically chose the red D for directives to make them the most prominent. The dark blue and placement for decorators as first in the list gives them too much visual prominence relative to the others. I can work on a better color palette consistent with material design, but probably not on short notice.
2) when I click on the decorators icon, I don't see any entries in the list.

I'm going to punt on this for 1.0, sorry.

petebacondarwin:
@naomiblack - I agree with @IgorMinar that this is a really useful addition to the docs. Aesthetics to one side, the ability to find decorators in the docs is severely limited right now. For example, imagine you want to create a pipe but you are not sure what decorator you might need. What do you look for right now? Searching for Pipe in the API list gives:

* AsyncPipe class
* COMMON_PIPES var
* CurrencyPipe class
* DatePipe class
* DecimalPipe class
* JsonPipe class
* LowerCasePipe class
* NumberPipe class
* PercentPipe class
* SlicePipe class
* UpperCasePipe class
* PLATFORM_PIPES var
* Pipe var
* PipeFactory interface
* PipeMetadata class
* PipeTransform interface

Is it obvious which of these is the place to look for the decorator?

As it turns out the `Pipe` var is the place to start and the `PipeMetadata` contains additional important information.

I expect that people who are new to Angular 2 (trying it for the first time after beta is released) will be quickly confused by this.

I am sorry that I gave decorators too much prominence in the UI - I didn't spend much time on that as I expected that you would want to tweak that anyway. But I don't think that should block it from being actioned before or near to beta release.

petebacondarwin:
Rebased on master (and fixed the capitalization issue with `function` docType in the filter list)

wardbell:
It's 2016. Time to bring this up again. I think we all agree decorators are hard to find and are worthy of distinction. They don't feel like "V-Variables" to me ... as they are clearly functions.

Perhaps the way forward is to update this idea again, lowering the positional and color  priority of the decorators (N's key objection).  

**Suggestion**: move "A" to the end (with Const or Enum) and make it gray (like Const or Enum). A distinct color is not important in this context, IMO, and we've run out of colors 

I realize "A" stands for annotation ... which few will get immediately ... but I can't think of anything better.  Can **we add tooltips** to these icons to give them some explanatory substance, at least on a desktop browser?

IgorMinar:
Ward, how about `@` instead of `A`?

wardbell:
Oooo. Great idea, Igor!

naomiblack:
I modified to make the `A` an `@`. Spacing could be tweaked, but better than prior art, so I'm pushing it live. Thanks, all. And Thanks @wardbell for re-pinging.

PR #503, opened 2015-12-15
docs(router): new chapter

wardbell:
I believe this is an acceptable publishable draft.

Let's squash it during the rebase into master. I want to keep the commits in the branch because of an intermediate state with what looks like a bug for brian.


naomiblack:
this is mostly merged with the exception of @wardbell's very last change that was force-pushed

wardbell:
closed by 728659ba71a86abcf10a153e96106fba8e69dc81

PR #507, opened 2015-12-15
feat(api-list): sync filters with URL query

### LINKED ISSUE #490: Synchronize the API filtering and search with the URL query parameters

IgorMinar:
This allows us to recover the filtering state when navigating back and
forth between the API list and individual API pages.

Closes #490


IgorMinar:
fixed the api listing breakage and renamed ai_type to apiType  - manual testing looks good. I'll merge this in.

IgorMinar:
rebased on top of master

PR #514, opened 2015-12-15
Minor fixes

robertmesserle:
Some more fixes from the running document (see commit messages for more details)


robertmesserle:
@naomiblack

naomiblack:
merged everything except demo 3 because it is broken. need a new example from misko.

PR #522, opened 2015-12-15
Changed ToDoApp to show the actual remaining todos

### LINKED ISSUE #521: Incorrect remaining count in TodoApp example
To get the remaining count the get method should be

```typescript
get remaining(): number {
    return this.todos.reduce((count, todo: Todo) => count - todo.done, this.todos.length);
}
```

aksharpatel47:
Fixes #521

### DIFF
--- a/public/docs/_examples/homepage-todo/ts/src/todo_app.ts
+++ b/public/docs/_examples/homepage-todo/ts/src/todo_app.ts
@@ -22,7 +22,7 @@ export class TodoApp {
   ];
 
   get remaining(): number {
-    return this.todos.reduce((count, todo: Todo) => count + todo.done, 0);
+    return this.todos.reduce((count, todo: Todo) => count - todo.done, this.todos.length);
   }
 
   archive(): void {
--- a/public/docs/_examples/homepage-todo/ts/src/todo_app.ts
+++ b/public/docs/_examples/homepage-todo/ts/src/todo_app.ts
@@ -21,8 +21,8 @@ export class TodoApp {
       {text:'build an angular app', done:false}
   ];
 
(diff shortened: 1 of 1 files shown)


petebacondarwin:
LGTM although it might be clearer to write:

```
return this.todos.length - this.todos.reduce((count, todo: Todo) => count + todo.done, 0);
```

aksharpatel47:
@petebacondarwin That will show the done count and not the remaining count.

wardbell:
I agree with @ntaoo; the method should do what it says its doing. I'd keep the `reduce`.  
```
  get remaining() : number {
    return this.todos.reduce((count, todo: Todo) => count + !todo.done, 0);
  }
```
No need for the explicit return type either; TS knows it's a number.

petebacondarwin:
Actually I really like the filter version!

wardbell:
@petebacondarwin why prefer filter to reduce? It's not a huge deal in practice but reduce is more to the point.

It is also more efficient because it does not create a new array, only to discard it after calculating its length (I repeat: not measurable in practice).

petebacondarwin:
@wardbell well it just matches my brain's understanding of what we are trying to do. But on second look I am happy with your suggestion too.

wardbell:
THE MOST IMPORTANT thing is that we fix the darned method soon ... because it presently returns the number of completed Todos rather than the number remaining.

With everyone's indulgence, I'm just going to git-er-done here.

aksharpatel47:
@wardbell : You just wrote my thoughts during this entire conversation.

petebacondarwin:
Indulge yourself, with post-yule fervour!

PR #537, opened 2015-12-15
reorders tabs in homepage demos

robertmesserle:
@naomiblack Changed order of demo tabs per Misko's request.


naomiblack:
merged!

PR #542, opened 2015-12-15
[WIP] Devguide display data js

wardbell:
Copied from [IdeaBlade PR 12](https://github.com/IdeaBlade/angular.io/pull/12).

Jeremy W says: This provides the JS version of the displaying data devguide. I kept as much as possible to remain consistent with the TS version, but went ahead and removed any sections that are no longer applicable (like CORE_DIRECTIVES don't need to be declared anymore).  Pushed changes to a.52


wardbell:
Closing because we are not translating to ES5 in this manner

PR #544, opened 2015-12-15
[WIP] docs(structural-components): add chapter

wardbell:
Copied from [ideablade pr 10](https://github.com/IdeaBlade/angular.io/pull/10)

**Ward says**: WORK-IN-PROGRESS! Do not merge

The second half of Filipe's original "Structural Directives" chapter is here.

IMO (and Filipe and Pascal's apparently), that half seemed to depart from the point of Structural Directives and, in fact, defined a new kind of component that did the job of incorporating external content by projection.

This seemed to me to be a new concept, deserving its own chapter. For lack of better terms, I dubbed it the "Structural Component".

So, Filipe, this is something on my plate for review in the not-to-distant future.  You may want to take it a next step before I get there. My sense is that a Tab/Tab Pane example fits perfectly in this chapter. Do you agree?

**Filipe says**:  I was trying to move away from tabs because that was Victors example for `ng-content` (which isn't a strong reason).

But to show `DynamicComponentLoader`, I think something simpler would be in order since tabs/tab-pane would require a bit more complexity in structure organization. 

For instance, the recall button component class could be provided by the parent component, and the dashboard would instantiated it. It would demonstrate that a component class could be referenced and used outside of templates, which is an amazing thing conceptually.

**Ward replies**:  I know Victor did Tabs. But that's on his blog. If that seems the most obvious example of the need to project developer content into a component's template, I'm sure he would not mind.

I would prefer NOT to lead with `DynamicComponentLoader`. It's cool but not mainstream.  Mainstream is the stuff with content children.  That said, I'd love to see what you do with it.


wardbell:
Noticed that @robertmesserle put I very nice and succinct version on the A2.io [home page](https://angular.io/)!  Source is [here](https://github.com/angular/angular.io/tree/master/public/docs/_examples/homepage-tabs/ts/src). We should steal it.

filipesilva:
It does look steal worthy! 

I'll work on this some more soon, and get back to you with a finished chapter.

filipesilva:
In a conversation with @teropa it just dawned on me that I haven't covered the simple default case of `<ng-content>` with no selector, which just transcludes everything in the components light dom. 

I need to add a section about that.

wardbell:
I agree with the direction. Carry on!

filipesilva:
@wardbell I've updated the current content and examples to beta.2, added intro, table of contents and other minor edits.

This PR at the moment does not include all the additions previously discussed, but it is functional and complete as is. We can come back to it later and add more content.

filipesilva:
@wardbell I've rolled this chapter into structural directives, as we discussed.

public/docs/ts/latest/guide/structural-directives.jade
```diff
@@ -331,6 +343,385 @@ figure.image-display
...
+  `@ContentChild` gives us the **first** component of a type inside the 
+  **shadow DOM**, whereas `@ContentChildren` would give us a list of **all**
+  of them.
+  
+  `@ViewChild` gives us the **first** component of a type inside the 
+  **light DOM**, whereas `@ViewChildren` would give us a list of **all**
+  of them.
+  
+  Lists given by `@ContentChildren` and `@ViewChildren` are of the `QueryList`
+  type, and guaranteed to always the current, up to date components.
```
zoechi: ... to always **contain**? the current ...

filipesilva: Good eye, thanks!

lazaridis-com:
@wardbell @filipesilva  - still actual?

Foxandxss:
We will revisit this kind of PRs soon.

PR #569, opened 2015-12-16
Update toh-pt2.jade

perryao:
Fixed a typo referring to the `Hero` type as a class, when it was defined in part 1 as an interface.


perryao:
I signed it!

wardbell:
We picked that up in the course of other cleanup. Thanks for reporting it!

PR #585, opened 2015-12-18
API styles redesign - for review only, do not merge

teresahu:


wardbell:
Closing because of pending redesign of angular.io

PR #593, opened 2015-12-20
chore: fix typo

wesleycho:
- Fix unintentional typo

This addresses https://github.com/angular/angular/issues/6029

PR #608, opened 2015-12-27
term correction

s9tpepper:
Fixing "faux-pas" to "faux", "faux-pas" is not the correct term here, "faux" is the intended term. Faux-pas is when you say something embarrassing or tactless in front of people, faux means artificial or imitation.


s9tpepper:
Closing, didn't see #593 which this issue duplicates.

PR #617, opened 2015-12-28
docs(quickstart): edits for brevity

naomiblack:
also note, airplane.png is on master, but commented out in jade because size/layout looked wrong.

Can you please note any changes needed on a conceptual level, as well as missing graphics? 

I noticed that angular.io/public/resources/images/devguide/architecture/module.png and angular.io/public/resources/images/devguide/architecture/overview2.png need to be updated as well.

Any others?


wardbell:
Some thoughts on the technical aspects of the new diagrams:

`databinding.png` should be rethought as the `{{}}`, `[]`, `()` are in confusingly different sizes and weights and the two-way databinding, `[(ngModel)]` isn't represented properly at all. Also, it must be `ngModel`, not `ng-model`.

`dependency-injection.png` doesn't make sense to me. The original attempts to convey the notion that the service requested by the constructor is injected ("flying") in.  That's now how I read the diagram now.

`injector-injects.png` has a similar problem. Where is that service going? Why is the word "component" so small?

`overview.png ` is based on an old overview diagram and is missing the module story in the upper left of the current diagram.  

`parent-child-binding.png` confuses me. I'm not sure the original was all that hot either. But this one I didn't get at all.

PR #543, opened 2015-12-15
[WIP] Devguide: Component Communication chapter is ready for QA

wardbell:
Copied from [ideablade pr 11](https://github.com/IdeaBlade/angular.io/pull/11).

PR #622, opened 2015-12-29
[WIP] docs(component communication): new chapter

wardbell:
WORK IN PROGRESS - do not merge
New chapter in review.
Replaces #543


public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
+include ../../../../_includes/_util-fns
+
+:marked
+  ## Component Communication
+
+  Real applications rarely use a single component to implement the entire functionality of the app. To create structured, testable
```
wardbell: This paragraph buries the lead. We should delete it.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  from them.
+  
+  Generally, an Angular app is a tree of components &mdash; with the bootstrapped component in its root. Thinking of an app's 
+  structure as a tree, there are components that are in the same branch of the component tree &mdash; these form parent and child hierarchies
+  &mdash;, and components in different branches &mdash; they are not directly related. Evidently, components need to communicate with each other
+  to carry out the tasks the app is created for.
+  
+  As we already learned, _component_ is a pivotal concept of Angular, and it was designed with _inter-component communication_ in mind.
+  We can think of the following information flows:
+  * From a parent to its children
```
wardbell: lower case "from", etc.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  * From a child to its parent
+  * Among unrelated, completely decoupled components
+  
+  In this chapter, we will learn six frequently used Angular mechanisms and techniques available for inter-component communication.
```
wardbell: Strike the future tense "will" in this para (and others).  Present tense is easier to read.  This is as much a note to myself as to you!

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  We will also get acquainted with the tradeoffs of each mechanism, and understand the scenarios best suited for using a particular method of communication.
+  Before diving deeply into the details, here is a short overview of these techniques:
+  
+  __#1: Parent to child communication with input property binding__
+  
+  The `prop-value` expression of the parent is bound to the `child-prop` input property of the child. `child-prop` is updated every time when `prop-value` changes:
+  
++makeExample('component-communication/ts/fragments/property-binding-intro.html')
```
wardbell: Ultimately, we'll have to eliminate all fragment files.

I think we can leverage your sample code to make these points rather than use difficult-to-test fragments.  It shouldn't be a problem to refer to `<hero-detail>` rather than `<child-comp>`.  

Otherwise, we have to create real samples with `ParentComponent` and `ChildComponent`.  That's possible but we have a lot of code already.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
+:marked
+  We can bind multiple properties of a single child component to the parent, and, implicitly, we can bind multiple children to the parent &mdash; 
+  ie. with `ng-for`:
```
wardbell: What does `NgFor` have to do with the point the "bind multiple properties" point? Or are there two points here ... one about multiple bindings and one about multiple children?  

IMO, the `NgFor` clouds the issue. Let's drop it. and stick w/ component communication.

p.s.:  must sweep the jade for all use of defunct kebab-cased directives.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
++makeExample('component-communication/ts/fragments/property-binding-intro-2.html')
+  
+:marked
+  [Jump to details](#-1-parent-to-child-communication-with-input-property-binding)
+
+:marked
+  __#2: Child to parent communication with event emitting__
+  
+  When a child needs to communicate with its parent, it can emit an event. The parent can define  its own event handler method , and bind this method
```
wardbell: elim space before comma

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
++makeExample('component-communication/ts/fragments/property-binding-intro-2.html')
+  
+:marked
+  [Jump to details](#-1-parent-to-child-communication-with-input-property-binding)
+
+:marked
+  __#2: Child to parent communication with event emitting__
+  
+  When a child needs to communicate with its parent, it can emit an event. The parent can define  its own event handler method , and bind this method 
+  to the child's output property:
```
wardbell: Confusing. The reader doesn't know what an `output property` is yet or how it relates to the `child-event` property.

I feel this example is confusing because of the `<button>`. We can trim it down to just
```
parent-comp>
  <child-comp (child-event)='handleMyEvent($event)'>
  </child-comp>
</parent-comp>
```

HEADS UP: Must rename all kebab-case sample component input/output properties (`child-event`) to camelCase (`childEvent`).

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+
++makeExample('component-communication/ts/fragments/event-binding-intro.html')
+  
+:marked
+  Here, the child event is emitted by the `raiseMyEvent` method. The parent declares the `handleMyEvent()` method and binds it to the `child-event`
+  output property of the child. Events can have arguments &mdash; here, `"Fired!"` is the argument &mdash;, which can be passed to the parent.
```
wardbell: Events may have **at most one** argument. That argument is an event object.

After removing the `<button>` we should not tell the reader that the event object is the `"Fired!"` string because the reader can't see that.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  At the parent side `$event` represents the argument of the emitted event.
+  
+  [Jump to details](#-2-child-to-parent-communication-with-event-emitting)
+
+  __#3: Child to parent communication with direct method calling__
```
wardbell: Let's remove this example entirely.

If we decide to discuss handing the parent to the child at all, we should do it in an appendix ... and there explain why it's discouraged ... which includes the high danger of violating unidirectional data flow when a child calls anything on a parent.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  to inject the parent into the child component's constructor.
+  The concrete technique does not matter: **Angular discourages this model, because it can sidestep the change detection mechanism, and so it may break
+  the data flow model.**
+  
+  [Jump to details](#-3-child-to-parent-communication-with-direct-method-calling)
+
+  __#4: Duplex parent/child communication through a service object__
+  
+  Parent and child components share their responsibilities. In this sharing both component types have their own responsibilities including UI logic
+  and others. The parent and its children both can manage state information about the UI logic. Often, the majority of communication is about
```
wardbell: "The parent and child delegate shared state management to the service and communicate about that state through the service."

A bit abstract. Need to think of a way to make this compelling in a few words.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+
++makeExample('component-communication/ts/fragments/event-binding-intro.html')
+  
+:marked
+  Here, the child event is emitted by the `raiseMyEvent` method. The parent declares the `handleMyEvent()` method and binds it to the `child-event`
+  output property of the child. Events can have arguments &mdash; here, `"Fired!"` is the argument &mdash;, which can be passed to the parent.
+  At the parent side `$event` represents the argument of the emitted event.
+  
+  [Jump to details](#-2-child-to-parent-communication-with-event-emitting)
+
```
wardbell: We should discuss and show the `NgModel` approach:

 `<hero-cape-picker [(ngModel)]="hero.capeColor"></hero-cape-picker>` 

Where the picked cape updates the parent property. If the caller needs to react to the change can do with (a) definedProperty and/or (b) `OnChange`.  Let's chat about that.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
+  Parent and child components share their responsibilities. In this sharing both component types have their own responsibilities including UI logic
+  and others. The parent and its children both can manage state information about the UI logic. Often, the majority of communication is about
+  managing the overall state information of this component fabric.
+  
+  In many cases the logic can be outsourced to a separate service that is the single source of truth, stores the state and provides operations to
+  access and alter the state. In such a scenario, parent and child components do not need to talk with each other directly, each of them 
+  communicates with the intermediary service:
+  
++makeExample('component-communication/ts/fragments/service-intro.ts', 'service')
```
wardbell: Again, I'd rather we find a way to leverage the real code rather than mention a `CommService`.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+
+:marked
+  This service is injected into the parent and child components via Angular dependency injection (via component constructors):
+
++makeExample('component-communication/ts/fragments/service-intro.ts', 'components')
+
+:marked
+  Both component types can access the service instance injected into the constructor. The dependency injection mechanism takes care &mdash; 
+  unless we do not configure it in another way &mdash; that the parent and its child components use the same service instance.
```
wardbell: The caveat ("unless we do not configure") is a bit confusing and unnecessary. If we have to say anything here ... and I'm not sure that we do say anything *here* ... it should be something like.

"The parent component instance and its children share a single instance of the `CommService`."

If that seems too mysterious, they can read about it in the detail.

public/docs/ts/latest/guide/component-communication.jade
wardbell: We give no hint why we would want to use a shared service instead of the earlier techniques.  I haven't read ahead yet to see if you talk about this.

If not, we need to. When is a shared state service a good idea?  I'm not sure how often it is. I think it is the right choice when 

A)  there is a complex, multi-stage, bi-directional workflow that is best represented by multiple touch points.

B) when the children should respond to messages or events from the parent.

C) to avoid complex parent/child protocols that would otherwise flow over a single property channel (corollary to A).

D) to simplify both child and parent components by delegating complex state and logic to a service (which is typically the reason to have a service in the first place). An example is a service that acquires and manage data from a remote server.

These are specialized scenarios. I suspect that the shared service is rarely one's first choice.  Particularly when you consider that it adds complexity and ... when the service exposes events to children ... requires careful `OnDestroy` logic to avoid memory leaks.

With this in mind, I'll look forward to seeing what YOU said.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  We can design and implement the approprate service methods and properties, and access them from within the components. 
+  
+  [Jump to details](#-4-duplex-parent-child-communication-through-a-service-object)
+
+  __#5: Communication among unrelated components via a message bus__
+  
+  In many inter-component communication scenario, the communicating parties are not in a perent-child relation with each other. They can
```
wardbell: "parent" not "perent".  Rewrite anyway to avoid leading with a preposition. Perhaps simply:

"The communicating components aren't always in a parent-child relationship."

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  be siblings, located in separate branches of the component hierarchy, or even be totally unrelated to each other.
```
wardbell: Actually, there are only two choices: "they can be siblings or they can be in distant branches of the application component hierarchy."  All components have a common root ancestor ... unless you're thinking about two separate applications launched from the same index.html ... a scenario I think we can safely ignore.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  unless we do not configure it in another way &mdash; that the parent and its child components use the same service instance.
+  We can design and implement the approprate service methods and properties, and access them from within the components. 
+  
+  [Jump to details](#-4-duplex-parent-child-communication-through-a-service-object)
+
+  __#5: Communication among unrelated components via a message bus__
+  
+  In many inter-component communication scenario, the communicating parties are not in a perent-child relation with each other. They can
+  be siblings, located in separate branches of the component hierarchy, or even be totally unrelated to each other. 
+  In such scenarios, components can communicate with each others directly or through intermediary object.
```
wardbell: How can these components communicate with each other (not "others") directly?  I don't know how. I don't think we show how. AFAIK, they could only communicate through an intermediary such as a "message bus".

The "message bus" is a variation of the service example we showed before. We should be clear about that.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
+  In this chapter, we will create a message bus that is used as a common communication channel for sending and listening to messages:
+  
++makeExample('component-communication/ts/fragments/message-bus-intro.ts')
+
+:marked
+  We rely on Angular's dependency injection to allow the components to use the message bus.
```
wardbell: See my earlier note about this being structurally the same as the service example.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
+  [Jump to details](#-5-communication-among-unrelated-components-via-a-message-bus)
+
+  __#6: Parent to child communication where children are transposed in &mdash; using `@ContentChildren`__
```
wardbell: The name for this isn't quite right. 

More importantly, I think you've confused two different cases. In the first, you have a directive (such as a TabControl) that does involve transposition.  We know that it does because the consumer's content is between the open and the close tags of the parent directive.  This example would look for **content** children.

The second example uses `NgFor`.  `NgFor` uses the `<template>` tag to stamp out new views that are **view** children of the parent component, `<parent-comp>`.

Notice the difference:  *content* children vs. *view* children.  They have different search and access mechanisms.

The *content* case is advanced stuff.  I wonder if we should tackle it.  The *view* children are not the same as the *content* children. The view children can be a superset of the content children ... as can be seen in this [variation of your plunker](http://plnkr.co/edit/uOjbcpqTU1z7TtXE875i).

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
++makeExample('component-communication/ts/app/take-job-contentchildren/job-board-with-content.ts', null, 'take-job-contentchildren/job-board-with-content.ts', {blk: /(HeroCrm)/g})
+
+:marked
+.alert.is-helpful
+  :marked
+    We need a component that can be used as the one to be the root component displayed by `router-outlet` we use in this example &mdash; that is why we
+    use `JobBoardWithContent`. Of course, we set up the routing configuration for this scenario path accordingly.
+:marked
+  Here, we imitate that hero information is coming from our Hero CRM system, and each hero is represented by a `hero-crm` entity. Just for the sake of
```
wardbell: What does "crm" stand for?  The CRM system? Very confusing name. You say this is an entity.  Why would a directive represent an entity? An entity is usually part of the application model ... it's not a UI artifact.

Now sometimes we create a UI wrapper around an entity ... which is widely known as a "ViewModel" or an "ObjectViewModel".  Is that what you intend? If so, I wouldn't call it an entity.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  the list of heroes to invite, so that the corresponding `HeroPanel` components could be nested into `HeroJobBoard`.
+  
+  For the sake of simplicity, we declare the list of heroes to invite as the content of `HeroJobBoard`:
+  
++makeExample('component-communication/ts/app/take-job-contentchildren/job-board-with-content.ts', null, 'take-job-contentchildren/job-board-with-content.ts', {blk: /(HeroCrm)/g})
+
+:marked
+.alert.is-helpful
+  :marked
+    We need a component that can be used as the one to be the root component displayed by `router-outlet` we use in this example &mdash; that is why we
```
wardbell: The fact that we use a router shouldn't come up at this level of detail. It's part of the sample harness ... the thing that enables us to jump among the examples within this chapter.

It plays no role in the example you're talking about.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+:marked
+  Now, we put together all pieces to check how the message bus works:
+  
+figure.image-display
+  img(src="/resources/images/devguide/component-communication/message-board-ui.png" alt="MessageBoard in action")
+  
+:marked
+  ## #6: Parent to child communication where children are transposed in &mdash; using `@ContentChildren`
+  
+  There are situations, when a parent component cannot determine the set of its child components on its own, because the collection of its children is managed
```
wardbell: I actually do understand what you're saying but I don't think the reader will get it. Who is managing externally such that the author of the component can't know the children? That would need explaining.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  by an external entity. However, even in this case, the parent needs to communicate with each children. How can the parent establish the communication channel 
+  to its children? Angular provides several mechanism to achieve this, here we show the usage of the `@ContentChildren` annotation.
```
wardbell: Several mechanisms?  What are they other than the content children example you're about to look at?  Why would I pick this mechanism over the others?

Btw, the only other mechanism that comes to mind is the shared service mentioned earlier. Why choose this approach over that one?

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
+  When the **Invite heroes** button is clicked, `inviteHeroes()` copies the content of `heroesToInvite` into `heroesInvited`.
+  
+.alert.is-helpful
+  :marked
+    The `inviteHeros()` method uses the `toArray()` method of `QueryList` to convert the `QueryList<HeroCrm>` object to a `Hero[]`. Because the shape of
+    `HeroCrm` is compatible with `Hero` &mdash; they have the same properties &mdash; TypeScript allows this implicit conversion.
+    
+    The component lifecycle has an explicit point when the content of `HeroJobBoard` is collected by the framework, and then `heroesToInvite` is updated. You can hook into
+    this lifecycle event by defining the `ngAfterContentInit()` method, and carry out any additional setup you may need to use the collection of children.
```
wardbell: You don't showed the usage of `ngAfterContentInit`. It appears no where in the sample. We can't talk about it if we don't show it.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  In the previous scenarios, the `HeroJobBoard` component managed the collection of heroes to invite this way:
+  
++makeExample('component-communication/ts/fragments/heroes-to-invite.ts', null, null, {pnk: /(invitedHeroes)/g, blk: /(Hero\.heroes)/g})
+
+:marked
+  `HeroJobBoard` obtained the collection of heroes from the `heroes` static property of the `Hero` class. This solution was pretty useful for demonstration,
+  but it does not refelect the real life. In a real application, the list of heroes is probably retrieved from the backend.
+  
+  Let's play with the idea &mdash; and it's seems realistic &mdash; that the list of heroes to invite is managed outside of `HeroJobBoard`, and even it
+  changes while we use the app. The Hero Job Board application's logic requires that by the time we click the **Invite heroes** button, we need to know
```
wardbell: As I observed above, the sample does not illustrate this case. I don't mean that it should go to a server.  I mean that the list is not determined dynamically and is in fact consumed long before the user pushes the invite button.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
+  In the previous scenarios, the `HeroJobBoard` component managed the collection of heroes to invite this way:
+  
++makeExample('component-communication/ts/fragments/heroes-to-invite.ts', null, null, {pnk: /(invitedHeroes)/g, blk: /(Hero\.heroes)/g})
+
+:marked
+  `HeroJobBoard` obtained the collection of heroes from the `heroes` static property of the `Hero` class. This solution was pretty useful for demonstration,
+  but it does not refelect the real life. In a real application, the list of heroes is probably retrieved from the backend.
+  
+  Let's play with the idea &mdash; and it's seems realistic &mdash; that the list of heroes to invite is managed outside of `HeroJobBoard`, and even it
```
wardbell: This is not the way to argue for this approach. Changes to the displayed list are not the driver.

If the list changed over time we could handle that entirely with the binding mechanisms shown earlier and with `NgFor`.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
++makeExample('component-communication/ts/fragments/no-message-bus-in-hero-job-board.ts', null, 'message-board/hero-job-board.ts (extract)', {blk: /(providers: \[JobService\])/g})
+
+:marked
+  Now, we put together all pieces to check how the message bus works:
+  
+figure.image-display
+  img(src="/resources/images/devguide/component-communication/message-board-ui.png" alt="MessageBoard in action")
+  
+:marked
+  ## #6: Parent to child communication where children are transposed in &mdash; using `@ContentChildren`
```
wardbell: We need a different name for this section. This one doesn't speak in plain language that the reader can understand.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+.alert.is-helpful
+  :marked
+    The `inviteHeros()` method uses the `toArray()` method of `QueryList` to convert the `QueryList<HeroCrm>` object to a `Hero[]`. Because the shape of
+    `HeroCrm` is compatible with `Hero` &mdash; they have the same properties &mdash; TypeScript allows this implicit conversion.
+    
+    The component lifecycle has an explicit point when the content of `HeroJobBoard` is collected by the framework, and then `heroesToInvite` is updated. You can hook into
+    this lifecycle event by defining the `ngAfterContentInit()` method, and carry out any additional setup you may need to use the collection of children.
+
+:marked
+  We do not need to change any other components to complete our task. When the app is running, and the **Invite heroes** button is clicked, we
```
wardbell: And that is part of the problem with this sample.

It doesn't demonstrate that the getting the children has anything at all to do with component communication.  You've shown how to get the children. But you're not using the children in a way that demonstrates how they offer a communication mechanism that's different than anything else you've shown.

You're using the same old binding techniques to the `HeroPanel`. The only difference is that you've found a complex and improbably way to get the heroes via shadow DOM.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
++makeExample('component-communication/ts/fragments/heroes-to-invite-2.ts', null, 'take-job-contentchildren/hero-job-board.ts (extract)', {blk: /(@ContentChildren\(HeroCrm\)|QueryList&lt;HeroCrm&gt;)/g, pnk: /(heroesToInvite)/g})
+
+:marked
+  The `JobBoardWithContent` component nests 8 `hero-crm` children: 6 coming from `Hero.heroes`, and 2 others (Dynama and Magma) added with direct markup.
+  
+  ### A few additional comments about `@ContentChildren`
+  
+  * `@ContentChildren` provides a mechanism for a parent to collect its children defined as content. We can use _all the communication techniques between the
+  parent and its children_ (in both direction) you already learned in this chapter.
+  * Components can have multiple `@ContentChildren` properties witch different child types.
```
wardbell: "which"

wardbell: The point you're making about different content children with different types won't be understood by the reader, mostly because we don't show it. I almost didn't get it myself.

No point in talking about something we don't show. FWIW, I don't think this is the place to show or talk about it.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  * `QueryList` has a `changes` property that provides a `subscribes()` method to be notified about changes in the content children.
```
wardbell: Too bad we're not making use of that.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
+  If we have have unrelated components that are in separate subtrees, they can communicate **through intermediary components**, such as services:
+  * The same intermediary component instance can be injected into the the communicating parties. 
+  * The parties utilize the intermediary component, 
+    * they can access its properties, 
+    * invoke its methods,
+    * and subscribe to its events.
+    
+  As we discussed, intermediary components are great alternatives for parent to child, and child to parent communication, too. With the help of 
+  `@ContentChildren` we can transpose children into their parent, and utilize the parent-to-child or child-to-parent communication techniques we
```
wardbell: The component children case should be in a new paragraph as it has nothing to do with "intermediary components"

wardbell: `@ContentChildren` doesn't help us "transpose" children.  We aren't "transposing" children in any case.

We are "projecting" the external content (says the Angular team) that the consumer put between tags.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
+  Here is a screenshot of our complete application:
+  
+figure.image-display
+  img(src="/resources/images/devguide/component-communication/app-ui-snapshot.png" alt="UI in action")
+  
+:marked
+  ### Live Example
+  
+  The application is available as a live example. It is a single Angular app that uses the router configured with separate routing paths
```
wardbell: I don't think that we are clear enough about what this sample is for.

We should say that we're implementing the same example 6 different ways, each way illustrating one of the aforementioned techniques.

The role of the router is to make it easy for us to hop among the 6 different illustrations.  We should also correlate the 6 techniques with the 6 illustrations and make that clear in the sample links.

Also note that we have 6 techniques and 7 sample links. That doesn't align.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  Let's start, and dive into the details!
+  
+  ## #1: Parent to child communication with input property binding
+  
+  Angular allows components to declare _data-bound input properties_. These are like regular component, but the value of an 
+  input property takes its value from the expression bound to it. The framework automatically updates data-bound input properties
+  during change detection.
+  
+  This is a _one-way binding_. As the source value changes, the destination property of the component is changed, but it does not 
+  happen in the opposite direction: altering the value of the input property within the component does not affect the source.
```
wardbell: Not precisely true.  We can modify the `Hero` object itself and that modification flows through to the parent.

What you mean to say is that we can't change the property value and thus cannot replace the hero reference.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  `HeroPanel` children. The `HeroJobBoard` component is the one that knows the circle of heroes who should be invited to listen to job requests, and it embeds 
+  a `HeroPanel` component instance for each hero.
+  
+  In this scenario, `HeroPanel` declares an input property, `hero`, as the `@Input()` decorator suggests:
+  
++makeExample('component-communication/ts/app/invite-heroes/hero-panel.ts', null, 'invite-heroes/hero-panel.ts', {pnk: /(@Input\(\))/g})
+
+:marked
+  `HeroJobBoard` can iterate through its `invitedHeroes` collection and set up its `HeroPanel` children so that a single hero is
+  assigned to each of them through the `hero` input property. As we learned before, the Angular property binding syntax
```
wardbell: Please avoid passive constructions such as "is used".

Keep it simple:

We use Angular property binding to pass the current iteration hero to the child component.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  (property name wrapped into square brackets) is used:
+
++makeExample('component-communication/ts/app/invite-heroes/hero-job-board.ts', null, 'invite-heroes/hero-job-board.ts', {pnk: /(\[hero\])/g})
+  
+:marked
+  The `Hero.heroes` static collection contains six heroes for demonstration purposes. When the application starts, `invitedHeroes` is empty, 
+  but clicking the __Invite heroes__ button invokes `inviteHeroes()` and fills up the `invitedHeroes` array. According the template definition, 
+  these bindings are established between `HeroJobBoard` and its `HeroPanel` children:
+  
+figure.image-display
```
wardbell: The diagram shows the old `ng-for` syntax.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  Instead of using the `@Input()` annotation, we can use the `inputs` annotation property of `@Component()` to specify input properties.
+
+  Using the alternative method, we could have defined the `hero` input property this way:
+  
++makeExample('component-communication/ts/fragments/using-inputs.ts', null, null, {pnk: /(inputs)/g})
+
+:marked
+.alert.is-helpful
+  :marked
+    In this chapter, we are age going to use the `@Input()` annotation.
```
wardbell: "we are age going to use the"? Do you mean "we will always use the"?

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  The Hero Job Board application uses input data binding to pass the information about an invited hero from `HeroJobBoard` (parent) to all
+  `HeroPanel` children. The `HeroJobBoard` component is the one that knows the circle of heroes who should be invited to listen to job requests, and it embeds 
+  a `HeroPanel` component instance for each hero.
+  
+  In this scenario, `HeroPanel` declares an input property, `hero`, as the `@Input()` decorator suggests:
+  
++makeExample('component-communication/ts/app/invite-heroes/hero-panel.ts', null, 'invite-heroes/hero-panel.ts', {pnk: /(@Input\(\))/g})
+
+:marked
+  `HeroJobBoard` can iterate through its `invitedHeroes` collection and set up its `HeroPanel` children so that a single hero is
```
wardbell: Not "can".  It DOES iterate through ...

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  `@Input()` always must be used with parentheses:
+
++makeExample('component-communication/ts/app/invite-heroes/hero-panel.ts', 'input-def', null, {pnk: /(@Input\(\))/g})
+
+:marked
+.alert.is-critical
+  :marked
+    It's easy to forget the parentheses. Our application will fail mysteriously if we do.
+:marked
+  Instead of using the `@Input()` annotation, we can use the `inputs` annotation property of `@Component()` to specify input properties.
```
wardbell: It is not an "annotation property".  It's a property of the `@Component `configuration object.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  and it makes displaying data simple.
+  
+  We have already implemented the functionality behind the __Invite heroes__ function with input data binding. The same mechanism allows us
+  to implement announcing jobs. As the altered definition of `HeroPanel` shows, now it has another input property, `request`, which has two
+  occurrences within the template:
+
++makeExample('component-communication/ts/app/send-job-request/hero-panel.ts', 'component', 'send-job-request/hero-panel.ts (extract)', {pnk: /(&quot;request&quot;|{{request \|\|.*\}\})/g})
+
+:marked
+  As the value of `request` changes, the value of the `class` attribute is updated, as well as the text of the `<h4>` tag through the
```
wardbell: replace the passive sentence with an active sentence.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
++makeExample('component-communication/ts/fragments/sending-request.ts', null, 'send-job-request/hero-job-board.ts (extract)', {pnk: /(\[request\]=request)/g})
+
+:marked
+  After inviting heroes, clicking the **Ask** button invokes `announceJob()` that sets the `request` property of `HeroJobBoard` &mdash; and Angular propagates this
+  change to all `HeroPanel` instances:
+
+figure.image-display
+  img(src="/resources/images/devguide/component-communication/send-job-request-ui.png" alt="Job request is sent")
+
```
wardbell: Missed an important opportunity to talk about how the child can know when an input changes.

Suppose we bound to a single component ... a message component ...
```
<message-box [message]="currentMessage"></message-box>
```
How does the `MessageBox` know when the `currentMessage` changes?

That's a really important question.

I can think of 2 ways (I learned them from Victor):
1. implement the `MessagBox.message` as a setter-only property.  The setter method is called when the message changes which is the opportunity for the message box to respond (e.g., by changing the color of the display based on whether the message is a success, warning, or error message).

1. Implement `ngOnChanges`. Angular calls this method when any input value (e.g., the message) changes. 

The first option is usually the simplest. The second option makes more sense when you want to respond to multiple input binding changes and, perhaps, coordinate among them. For example, if our component is bound to departure and arrival date properties, we'd want to know both dates before updating the display (e.g., showing an error if the user would arrive before departure).

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
+  ## #2: Child to parent communication with event emitting
+  
+  Input property binding is an easy-to-use technique that works great in most parent-to-child communication scenarios. As the expression bound to
+  the input property changes, the modification is immediately propagated to all children. At first we may think that this mechanism could be
+  applied in the opposite direction, to send data from a child to its parent.
+  
+  _Angular does not implement input data binding in this direction._
+  
+  It is a deliberate design decision. Angular was designed with _unidirectional data flow_ in mind. This approach allows better performance
```
wardbell: It's good point you make about no reverse property binding from child to parent. Enough to observe that all data flows are parent to child during a single JS turn. Flows from child to parent must occur in a subsequent JS turn. This is an aspect of unidirectional data flow (I call it uni-flow for short ;-))

Let's skip the discussion of why? Just cut these sentences and move on.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  applied in the opposite direction, to send data from a child to its parent.
+  
+  _Angular does not implement input data binding in this direction._
+  
+  It is a deliberate design decision. Angular was designed with _unidirectional data flow_ in mind. This approach allows better performance 
+  when detecting and propagating property changes, which would not be easily available with a _two-directional data flow_ mechanism. 
+  The unidirectional data flow approach allows avoiding issues coming from recursion, too, when both communicating parties watch for property 
+  changes and at the same time may change the very same property's value.
+  
+  Instead of using a duplex communication channel between parent and child components, we can use two simplex channels, one from the parent to
```
wardbell: It may be misleading to talk about property and event bindings as two simplex channels. They aren't. They are completely different mechanisms: different syntax, different timing, different binding targets.

Just say that a child can communicate with its parent via events.

Better to say that out

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+
+:marked
+  After inviting heroes, clicking the **Ask** button invokes `announceJob()` that sets the `request` property of `HeroJobBoard` &mdash; and Angular propagates this
+  change to all `HeroPanel` instances:
+
+figure.image-display
+  img(src="/resources/images/devguide/component-communication/send-job-request-ui.png" alt="Job request is sent")
+  
+:marked
+  Now, let's see how a child can communicate with its parent.
```
wardbell: Set the problem up.

Notice that in this illustration we did not provide a way for the child to accept a job. We could have bound a callback method  with a property binding
>and may we should show that?
but that's hacky.

Child components don't use property binding to communicate with their parents. They can talk to their parents with events.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  The unidirectional data flow approach allows avoiding issues coming from recursion, too, when both communicating parties watch for property 
+  changes and at the same time may change the very same property's value.
+  
+  Instead of using a duplex communication channel between parent and child components, we can use two simplex channels, one from the parent to 
+  a child (or children), and another one in the opposite direction, from the child to its parent. While we can use input property binding
+  from parent to child, Angular offers event emitting (output event binding) for child-to-parent communication.
+  
+  ### Establishing output event binding
+  
+  To set up output event emitting, we need to carry out these steps:
```
wardbell: Simplify this section. Suggestion:

>We establish an event-based channel in three steps:
>
>1. Create a child property that returns an `EventEmitter`
>2. Declare it to be a public *output* property
>3. Bind that property to a parent component event handler

Re-arrange the discourse to conform to this recipe.  Write the property first, then declare it public, then bind to it.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  It is a deliberate design decision. Angular was designed with _unidirectional data flow_ in mind. This approach allows better performance 
+  when detecting and propagating property changes, which would not be easily available with a _two-directional data flow_ mechanism. 
+  The unidirectional data flow approach allows avoiding issues coming from recursion, too, when both communicating parties watch for property 
+  changes and at the same time may change the very same property's value.
+  
+  Instead of using a duplex communication channel between parent and child components, we can use two simplex channels, one from the parent to 
+  a child (or children), and another one in the opposite direction, from the child to its parent. While we can use input property binding
+  from parent to child, Angular offers event emitting (output event binding) for child-to-parent communication.
+  
+  ### Establishing output event binding
```
wardbell: Make these sub heads imperative. Say "establish", "write", "create".  Not "establish*ing*"

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  1. Declaring this property as an `EventEmitter`
+  1. Using this property with the _output event binding syntax_ in the parent template to bind it to an event handler method
+  
+  The Hero Job Board application uses output event binding to sign that a hero undertakes the announced job. The `HeroPanel` component declares 
+  an output property as a type of `EventEmitter`, through which &mdash; as the type name suggests &mdash; it can emit an event. The parent 
+  `HeroJobBoard` component can use the output property to attach an event handler to that particular event.
+  
+  `HeroPanel` declares the `onJobTaken` output event property, as the `@Output` decorator indicates:
+  
++makeExample('component-communication/ts/app/take-job-event/hero-panel.ts', 'component', 'take-job-event/hero-panel.ts (extract)', {pnk: /(@Output\(\)|takeJob\(\)|onJobTaken\.next)/g})
```
wardbell: **Don't use the *deprecated* `next` method;  use `emit`!**

Find and fix all occurrences.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+
+:marked
+  The `takeJob()` method invokes the `next()` method of `onJobTaken`, and passes the `hero` assigned to this `HeroPanel`. 
+  The parameter of `next()` is the argument of the event emitted. Thanks to the strong type notation in TypeScript, `EventEmitter` is a generic type,
+  so we could declare `onJobTaken` as an `EventEmitter<Hero>`. When we invoke `next`, the TypeScript compiler checks whether the event argument
```
wardbell: Don't talk about it. Do it. Use the strongly typed form.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  so we could declare `onJobTaken` as an `EventEmitter<Hero>`. When we invoke `next`, the TypeScript compiler checks whether the event argument 
+  is compatible with `Hero`.
+  
+  To use `@Output()` and `EventEmitter`, `HeroPanel` needs to import them:
+  
++makeExample('component-communication/ts/app/take-job-event/hero-panel.ts', 'import', null, {blk: /(Output|EventEmitter)/g})
+
+:marked
+  Now our `HeroPanel` is ready to emit its event. `HeroJobBoard` attaches its `heroTakesJob()` event handler method to the `onJobTaken` output
+  event of `HeroPanel` using the output property binding syntax &mdash; wrapping the snake-case name of the output property, `on-job-taken`, into parentheses.
```
wardbell: Replace the "snake case" (which was actually dash- or kebab-case) with camelCase ... everywhere.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  The output property binding expression passes the `$event` identifier &mdash; that represents the argument of the emitted event &mdash; to the handler method:
+  
++makeExample('component-communication/ts/app/take-job-event/hero-job-board.ts', 'component', 'take-job-event/hero-job-board.ts', {pnk: /(\(on-job-taken\))/g, blk: /(heroTakesJob\(hero: Hero\))/g})
```
wardbell: Too much code. Can't find the relevant part of the template.

I think we're going to have to move the HTML out of the component and into its own file so that we can use `<!-- #docregion -->` to cite the appropriate fragment.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+
+.alert.is-critical
+  :marked
+    Although this technique works in many situations, it's an **antipattern**, and Angular discourages using it.
```
wardbell: WAT?  Where did you hear that emitting events is an anti-pattern?

It's highly recommended actually. What do you think is wrong with it?

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
+:marked
+  The event argument is the `hero` property value of `HeroPanel` that is passed to the `next()` method of the `EventEmitter` instance beyond
+  `onJobTaken`. `HeroJobBoard` receives the `Hero` instance that undertakes the job, and adds it to its `respondingHeroes` array.
+  
+  With output event binding, we established this communication structure from `HeroPanel` children to `HeroJobBoard`:
+  
+figure.image-display
```
wardbell: Diagram uses outdated syntax

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  img(src="/resources/images/devguide/component-communication/child-to-parent.png" alt="Output event binding")
+  
++makeExample('component-communication/ts/fragments/hero-job-board-with-parentref.ts', null, 'take-job-parentref/hero-job-board.ts (extract)', {pnk: /(\[parent\]|jobBoard\(\))/g})
```
wardbell: This discussion about `parent` binding actually ***belongs under the first technique ... Property Binding***.

It is a way to have bi-directional "communication" with property binding alone.

I don't mind mentioning this approach. We shouldn't go on about it and I do NOT think we should have a sample that does it.  That only encourages the idea.

wardbell:
### General Observations about Samples

Sweep text and samples for old syntax, especially kebab-case.

Eliminate deprecated methods such as `EventEmitter.next`

Align sub-sample links with the techniques so it is easier to correlate the text with the illustrations.

Be consistent in all bindings. Sometimes they are quoted, sometimes not as in this from a `hero-job-board.ts`
```
<hero-panel *ngFor='#hero of invitedHeroes'
  [hero]='hero'
  [request]=request
  [parent]='jobBoard'>
</hero-panel>
```
**Always quote. Always use double quotes in _templates_**
```
<hero-panel *ngFor="#hero of invitedHeroes"
  [hero]="hero"
  [request]="request"
  [parent]="jobBoard">
</hero-panel>
```
**Always use single quotes in _code_**

Move long inline templates to template files and use docregions for citations in jade.

Eliminate all fragment files.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+
+:marked
+  Now, we can invite heroes, announce job requests, and our heroes can opt to undertake the job. The `HeroJobBoard` component lists responding
+  heroes in the UI:
+  
+figure.image-display
+  img(src="/resources/images/devguide/component-communication/take-job-ui.png" alt="Hero undertakes job")
+  
+:marked
+  ### Using `@Output`
```
wardbell: I believe you have a sequence problem. The discussion of `@Output` belongs up with the event emitter property.

When you remove the `parent` example above, this discussion will fall into place.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
+.alert.is-critical
+  :marked
+    **`@Output()` always must be used with parentheses.** If we forget about it, our application will fail inscrutably.
+  
+:marked
+  Instead of decorating properties with the `@Output()` annotation, `@Component()` can be used to define output properties.
+  We could have declared that `onJobTaken` is an output property by setting up the `outputs` annotation property of `@Component`:
```
wardbell: Not an "annotation property". See my phrasing w/r/t the `inputs` array.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  
++makeExample('component-communication/ts/fragments/alternative-output.ts', null, null, {pnk: /(outputs)/g})
+
+:marked
+  ### Event arguments
+  
+  A child component can easily identify its parent during parent-to-child communication, as it has exactly one parent. Thus when a parent
```
wardbell: Too wordy. Reduce ... perhaps to this:
>We bound all `onJobTaken` child event properties to the same parent `'heroTakesJob` method. We have to tell the parent *which* hero took the job. That's why each child emits the `hero` that took the job and the binding passes that hero to the parent method via the Angular `$event` object.

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+  that the parent needs to know which of its children sends the message: nonetheless, the parent needs to know how to process the message, and
+  it may depend on its source, the child that sends it.
+  
+  When we pass arguments to events, these arguments should contain all information the parent needs to know to process the child components'
+  messages properly. We can use the non-generic `EventEmitter` type that allows using any type of event arguments. However, we'd better use
+  events with well-defined arguments that can be represented with strong TypeScript types.
+  
+.alert.is-helpful
+  :marked
+    Angular encourages us to use strong event argument types whenever it's possible, because it allows creating more maintainable and
```
wardbell: Do it, don't talk about it.

Didn't we say this already?

public/docs/ts/latest/guide/component-communication.jade
```diff
@@ -0,0 +1,707 @@
...
+    less error-prone code.
+  
+:marked
+.alert.is-critical
+  :marked
+    Nothing prevents us to create event argument types that pass back a reference to the child component instance to its parent so that the
```
wardbell: Why are we talking about this? It's not something we're showing.

To be honest, it isn't nearly as bad as passing the parent to the child.

Note also that we are passing the entire child to the parent when we use the `contentChildren` approach. If it's "bad practice" here, it must be bad practice there, right?

I think I'd just ditch this argument.

wardbell:
Closing because it is very old and too big. But we may want to revisit someday and mine it for the good stuff.

PR #628, opened 2015-12-31
Typo in TS Dependency Injection documentation

aarsan:
...if actually, should be "it actually"


aarsan:
I signed it!

- ahmet
Date: Wed, 30 Dec 2015 17:07:48 -0800

PR #647, opened 2016-01-05
TS Tutorial Services - Pt. 5 Typo, missing period

snlacks:


wardbell:
Actually, the error is in Pt. 4.  Accepting it.

PR #653, opened 2016-01-05
Change Obsevable into Observable

Gillespie59:


wardbell:
Opened to rename commit to fit standard

wardbell:
Closed again because rename is too late

PR #429, opened 2015-12-01
[WIP] docs(server-communication): adds chapter

cburgdorf:
**DO NOT MERGE**

That's a WIP in pre-alpha state. Don't dare to even look at it ;)

TODO

- [ ] upgrade to latest alpha
- [x] Intro
- [ ] Introduce a real backend (e.g. firebase)
- [ ] XHR
  - [x] Fetch data from remote location example (GET)
  - [ ] Send data to remote location example (POST)
  - [ ] Talk a little bit about REST and GET, PUT, POST, DELETE in general
- [x] JSONP
  - [x] Fetch data from wikipedias JSONP API example
- [ ] Error Handling
- [x] Taking advantage of Observables

PR #662, opened 2016-01-07
docs(guide/server-communication): adds http chapter

wardbell:
WIP - Don't merge yet

Update / replacement candidate for PR #429

TODO
- [ ] talk/show a little promises
- [ ] more on request options, especially headers
- [x] upgrade to latest alpha/beta
- [x] Intro
- [x] Introduce a real backend (e.g. firebase) [using in-mem web api instead]
- [ ] XHR
  - [x] Fetch data from remote location example (GET)
  - [x] Send data to remote location example (POST)
  - [ ] Talk a little bit about REST and GET, PUT, POST, DELETE in general
- [x] JSONP
  - [x] Fetch data from wikipedias JSONP API example
- [x] Error Handling
- [x] Taking advantage of Observables


cburgdorf:
@wardbell 

This can be reviewed again. I added bits about error handling and finished the post stuff. Also added a subnote about when to use the async pipe.

There are still some open questions:

1. Do you want the wikipedia example to have error handling to? Personally, I feel it's unnecessary noise to add error handling everywhere and I feel it's better to just show them once how it's done but otherwise keep examples more focused.

2. The POST example doesn't work for real so far since we don't have the backend yet. But I verified with an ad-hoc server that the code actually works if it had a backend to talk to.

3. Not sure what to do about Promises here. Do you think it's valuable to have the TOH example cloned for the promise approach? It's not that much different after all.

wardbell:
I updated the sample to use the in-memory web api so we can save as well as read.

I saw (working) code for the promises option but didn't see it in the text.

We're getting closer. We need at least another pass though.

PR #685, opened 2016-01-12
docs(tutorial): fix wrong text related to screenshots

Nasicus:
The text did not fit to the image in the tutorial. I changed the text a bit (tried to keep the same wording). 
It's still not 100% correct, to do that I think we'd either need to:

- introduce another image as a step between
- replace the image `heroes-list-2.png` with an image which also displays the readonly details

However this is probably a detail and I didn't know if you guys would be okay with it if I just replaced/added a new image and maybe we can also just leave it as it is.

So let me know what you think ;)

### DIFF
--- a/public/docs/ts/latest/tutorial/index.jade
+++ b/public/docs/ts/latest/tutorial/index.jade
@@ -48,14 +48,16 @@ figure.image-display
 
 :marked
   Links at the top can take us to either of the main views.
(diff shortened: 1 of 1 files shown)


wardbell:
It seems the app has changed since that description was written. Your suggestion doesn't match the new behavior either. Fixed in #716

PR #689, opened 2016-01-13
docs(quickstart): simplify System.import usage

zachlysobey:
In my opinion, the current usage of `System.import` is a bit clunky, and perhaps intimidating for novices.

__Current__
```typescript
System.import('app/boot')
        .then(null, console.error.bind(console)); 
```

__This PR__
```js
System.import('app/boot')
        .catch(err => console.error(err));
```

__Alt 1__
```js
System.import('app/boot')
        .catch(function(err) {
          console.error(err);
        });
```

__Alt 2__
```js
System.import('app/boot')
        .catch(console.error.bind(console));
```

### DIFF
--- a/public/docs/_examples/quickstart/ts/index.html
+++ b/public/docs/_examples/quickstart/ts/index.html
@@ -17,7 +17,7 @@
     <!-- #docregion systemjs -->
     <script>
       System.config({
-        packages: {        
+        packages: {
           app: {
             format: 'register',
             defaultExtension: 'js'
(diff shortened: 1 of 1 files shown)


robwormald:
The current PR state will fail in most all the browsers out there except Chrome, since it's not transpiled and it's the only one that supports arrow functions natively.

Alt2 is a reasonable middle ground.

wardbell:
Sorry. Rob's objections stand. Alt2 is modestly better but necessitates a sweeping change to all samples and tests in multiple browsers. I am unpersuaded that changing from `then(null, fn)` to `catch(fn)` does much to relieve the SystemJS intimidation factor

We'll probably do it eventually but we won't be in this PR.

Thanks for offering though.

PR #693, opened 2016-01-13
docs(tutorial): add dart version of toh tutorial

kasperpeulen:
@kwalrath This is only the end result, and only the source code of the tour of heroes tutorial.

It seems like the ts version of toh-1 up to toh-5 are not in sync with the end result yet. So I haven't ported those.


public/docs/_examples/tutorial/dart/lib/hero_detail_component.dart
```diff
@@ -0,0 +1,30 @@
...
+class HeroDetailComponent implements OnInit {
+  @Input() Hero hero;
+
+  final HeroService _heroService;
+  final RouteParams _routeParams;
+
+  HeroDetailComponent(this._heroService, this._routeParams);
+
+  ngOnInit() async {
+    int id = _routeParams.params['id'] as int;
```
kasperpeulen: For some reason angular2 assumes that this must be a `String`, but it is actually an `int`.

public/docs/_examples/tutorial/dart/web/main.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:angular2/bootstrap.dart';
+import 'package:angular2/router.dart';
+
+import 'package:angular2_tour_of_heroes/hero_service.dart';
+import 'package:angular2_tour_of_heroes/app_component.dart';
+
+main() {
+  bootstrap(AppComponent, [
+    ROUTER_PROVIDERS,
```
kasperpeulen: Use path location strategy doesn't go well with `pub serve`. See https://github.com/dart-lang/pub/issues/1374

Maybe use hash location strategy?

public/docs/_examples/tutorial/dart/.gitignore
```diff
@@ -0,0 +1,22 @@
+# Files and directories created by pub
```
kwalrath: You can delete this file, since angular2 has its own .gitignore.

public/docs/_examples/tutorial/dart/pubspec.yaml
```diff
@@ -0,0 +1,15 @@
+name: angular2_tour_of_heroes
+version: 0.0.1
+environment:
+  sdk: '>=1.13.0 <2.0.0'
+dependencies:
+  angular2: 2.0.0-beta.1
+  browser: ^0.10.0
+  dart_to_js_script_rewriter: ^0.1.0+4
```
kwalrath: We haven't specified this in other pubspecs.

public/docs/_examples/tutorial/dart/pubspec.yaml
```diff
@@ -0,0 +1,15 @@
+name: angular2_tour_of_heroes
+version: 0.0.1
+environment:
+  sdk: '>=1.13.0 <2.0.0'
```
kwalrath: We haven't specified this in other pubspecs. Do you think we should?

kwalrath:
@sanfordredlich could you please take a look?

public/docs/_examples/tutorial/dart/.gitignore
kasperpeulen: Are you sure? I saw a .gitignore in the the ts tutorial:

image

public/docs/_examples/tutorial/dart/pubspec.yaml
kasperpeulen: I think we should. Only for using caret notation in dependencies, we need 1.6.0 or higher I think. For async/await 1.9.0 or higher. Null aware operator 1.12.0 or higher (which I use). But to be safe, I always use the latest stable version.

public/docs/_examples/tutorial/dart/pubspec.yaml
kasperpeulen: I vote for including `dart_to_js_script_rewriter` in all tutorials. It also explained what this does in the quickstart.

public/docs/_examples/tutorial/dart/pubspec.yaml
kevmoo: Agreed! At this point, assuming 1.13 is goodness, me thinks.

public/docs/_examples/tutorial/dart/pubspec.yaml
kevmoo: This breaks Dartium though, right?

Having it as a one-off to show how to handle production is goodness. We shouldn't have it in every pubspec if it means Dartium won't work.

kasperpeulen: @kevmoo It doesn't break dartium for development:

> This transformer only runs when pub is running in release mode.
https://github.com/Workiva/dart_to_js_script_rewriter

So this transformer does does nothing in development. I think it actually fixes dartium in *release* mode, because otherwise `browser/dart.js` will point dartium to dart files, that don't exists in release mode.

public/docs/_examples/tutorial/dart/lib/app_component.dart
```diff
@@ -0,0 +1,32 @@
+import 'package:angular2/core.dart';
+import 'package:angular2/router.dart';
+
+import 'heroes_component.dart';
```
zoechi: Are relative imports ok now?

public/docs/_examples/tutorial/dart/lib/dashboard_component.dart
```diff
@@ -0,0 +1,31 @@
...
+    templateUrl: 'dashboard_component.html',
+    styleUrls: const ['dashboard_component.css'])
+class DashboardComponent implements OnInit {
+  List<Hero> heroes = [];
+
+  final HeroService _heroService;
+  final Router _router;
+  DashboardComponent(this._heroService, this._router);
+
+  ngOnInit() async {
```
zoechi: `async` is redundant when `await` isn't used.

public/docs/_examples/tutorial/dart/lib/hero_detail_component.dart
zoechi: I think you should ensure only strings are used in route parameters. If you navigate by clicking a bookmark you'll get a string not an int. I'm on the phone which makes reading cumbersome but i think it's cause by `gotoDetail()` where an int is passesd to `navigate()`

kasperpeulen:
It seems like I messed this PR up, see #700 for the new PR

kevmoo:
FYI: @kasperpeulen you can always `git push --force` to your PR branch if you want to clean things up.

Do you want me to close this one?

public/docs/_examples/tutorial/dart/.gitignore
kwalrath: I suppose I never noticed because I don't normally see dotfiles. I'll check in the angular author channel.

kasperpeulen:
@kevmoo ah that is a nice trick!

okay, closing this

PR #697, opened 2016-01-13
Pipe Doc Updates, clarification on impure pipes.

joeeames:


wardbell:
Todos:

- [ ] Why A2 ships without a  `FilterPipe`
- [ ] Memoization in an impure pipe improves perf (that's what the crazy JSON example does)
- [ ] Trim some of the rhetorical excess (says the man with rhetorical excess)

joeeames:
I believe the json example isn't actually memoized. it just executes once and then never executes again. memoization would involve a cache of inputs and corresponding outputs that shortcircuit the algorithm, but it doesn't do that.

RE the rhetorical excess: it definitely needs that.

wardbell:
So am I a "go" to incorporate these changes into the chapter (w/ my usual filter)?

joeeames:
yes

wardbell:
moved to PR #974 where I can work on it

PR #702, opened 2016-01-15
docs(landing): media query for mobile to fix formatting with flex style

NathanWalker:
Current state of mobile looked like this:
img_3523
img_3524

-----
# With these fixes:
img_3525
img_3527


naomiblack:
thanks! merged.

Issue #550, opened 2015-12-16
In plunkr examples, anchor tags with href=javascript:false does not work as expected in Firefox


princemaple:
They were meant to stop the clicking from doing anything, but leads to a page with "false" as the only content on it in Firefox.

Before clicking the link:
![screenshot_121615_125902_pm](https://cloud.githubusercontent.com/assets/1329716/11829951/c8cd1ffc-a3f4-11e5-9a3f-92e3c43397ac.jpg)

After clicking the link:
![screenshot_121615_125920_pm](https://cloud.githubusercontent.com/assets/1329716/11829957/d39a632c-a3f4-11e5-8b16-ec30e83337ee.jpg)

wardbell:
Fixed a different way a while back.

PR #719, opened 2016-01-18
fix: homepage example fix for firefox #550

nlmueng:
added return false; instead of false to fix issue with firefox


wardbell:
Thanks for your submission. It happens that we reworked these examples substantially so your change will no long apply. We also took a different approach, omitting the href altogether and using CSS to present the hand pointer.

See #740 

Thanks again. Keep helping.

PR #721, opened 2016-01-19
docs:add a2-in-memory-web-api support

wardbell:
Support an in-memory RESTy `HttpBackend` service for HTTP samples that can PUT/POST/DELETE
Adds optional package and plunker translation to samples
Available as [3rd party npm package](https://www.npmjs.com/package/a2-in-memory-web-api) until something like it offered by Angular
Package is MIT licensed,


wardbell:
Closed because subsumed in #662

PR #727, opened 2016-01-20
More readable

nurolopher:
Concatenated two half-repeating sentences into one.


nurolopher:
I signed it!

wardbell:
Thanks but this is a difference of literary taste not substance and doesn't significantly improve readability

PR #729, opened 2016-01-20
Update quickstart.jade

nurolopher:


wardbell:
Thanks but your suggestion is a tiny difference of literary taste not substance.

PR #730, opened 2016-01-20
Fix Docs Home layout for small devices

akmittal:
Docs Home page is not displayed correctly on mobile devices. Fix to set layout as column on small devices.
image


akmittal:
I signed it!

wardbell:
Looks like someone fixed it. Let me know if you think otherwise.  Thanks for the help.

PR #733, opened 2016-01-20
docs(guide): add dart version of lifecycle-hooks

adaojunior:
@kwalrath We may have to wait until this issue https://github.com/angular/angular/issues/6498 is solved


kwalrath:
I'm going to merge this change but add a note about angular/angular#6498.

PR #735, opened 2016-01-21
docs(tutorial): fix tutorial part 5

jechol:
Tutorial part 5 is broken now.

This fixes broken links and add part-5 to tutorial index.


wardbell:
Thanks for all of the effort but as I said many times this chapter was not broken but rather a work in progress whose pieces were committed here but not published. Sorry for the confusion.

The proper tutorial chapter is arriving any day.

PR #738, opened 2016-01-22
docs(guide): change angular2/angular2 to angular2/core

BeastCode:
issue #6462: change angular2/angular2 to angular2/core
  import {Injectable} from 'angular2/angular2'; -> import {Injectable} from 'angular2/core';


wardbell:
Thanks for the catch. This will be fixed within a much larger revision of the DI chapter, arriving in a day or two.

PR #739, opened 2016-01-22
chore(news): add ebook release news

cexbrayat:
Community news announcing the release of our ebook, as briefly talked with @bradlygreen .
@naomiblack @juleskremer tell me if that's OK ;)


naomiblack:
sorry -- I missed this back in Jan -- but would have declined. I would prefer to keep this page for news rather than promos.

cexbrayat:
No pb @naomiblack , I understand
I only submitted because it was encouraged by @bradlygreen (you were in cc of that email, but you may have missed it) so I thought it would be OK.

Anyway thanks for considering it!

PR #546, opened 2015-12-15
[WIP] feat(di-guide): move code to examples

wardbell:
Copied from [ideablade pr 5](https://github.com/IdeaBlade/angular.io/pull/5).

Pascal says: Please review this one.

I haven't moved all code snippets to examples yet since I'm not really sure how we want to handle this for simple one liners. Sometimes I also have the feeling it doesn't really make sense/gives us any additional value to move a simple snippet out to the examples folder.

Thoughts and feedback welcome!

PR #748, opened 2016-01-25
docs(dependency-injection): move code from jade to examples; add e2e tests

wardbell:
Refactor and finish what Pascal started in #546


wardbell:
I pushed it. As with all docs, it can be improved some more.

PR #757, opened 2016-01-26
docs(attribute directives/TS/Dart): remove `Renderer`; improve text

wardbell:
@kwalrath - Please confirm that my Dart changes are correct. Then we should push ASAP so we can run with latest A2 bits.


public/docs/_examples/attribute-directives/dart/lib/highlight_directive_1.dart
```diff
@@ -5,8 +5,7 @@ import 'package:angular2/angular2.dart';
 
 @Directive(selector: '[my-highlight]')
 class Highlight {
-  Highlight(ElementRef element, Renderer renderer) {
-    //el.nativeElement.style.backgroundColor = 'yellow';
-    renderer.setElementStyle(element, 'background-color', 'yellow');
+  Highlight(ElementRef element) {
+    el.nativeElement.style.backgroundColor = 'yellow';
```
kwalrath: el -> element, right?

kwalrath:
LGTM. I'll confirm that this still runs once you commit this change.

kwalrath:
What are the "latest" bits? You might need to modify pubspec.yaml to specify that version of Angular.

public/docs/_examples/attribute-directives/dart/lib/highlight_directive_1.dart
wardbell: Argh! I'll fix and force push

wardbell:
We don't have "latest bits" yet. We'll have to coordinate on the expected move to beta.2 which (if we are fortunate) will require no code changes.

PR #760, opened 2016-01-27
fix(nav): fixed tertiary nav

davidreher:
Although providing the possibility to use a tertiary nav, it is not working at the moment. Fixed it by copying the pathSuffix functionality from _primary.

By the way: This documentation tool is really great. We adopted it for our own docs, where we need the tertiary nav ;)


davidreher:
@wardbell can you take a look?

naomiblack:
I'm not going to merge this because we've explicitly decided not to provide a tertiary nav for angular.io. But I'm curious to see it on your site.

Issue #768, opened 2016-01-28
copy-button directive overlaps nav


Kevin-K:
![image](https://cloud.githubusercontent.com/assets/7290455/12645782/3ec25396-c59a-11e5-8e13-af100ca06e15.png)

![image](https://cloud.githubusercontent.com/assets/7290455/12645892/001f582c-c59b-11e5-8149-ece4c4437273.png)

The z-index on the copy-button directive (`z-index: 1000`) causes overlap onto the nav bar when scrolling (`z-index: 5`).

wardbell:
closed by #769

PR #769, opened 2016-01-28
fix(directives): set copy directive z-index below nav

Kevin-K:
Fixes: #768

Sets the z:index on the copy-button directive to 4 to be lower than the main navbar.

### DIFF
--- a/public/resources/js/directives/copy.js
+++ b/public/resources/js/directives/copy.js
@@ -25,7 +25,7 @@ angularIO.directive('copyContainer', function() {
     transclude: true,
     template:
       '<div style="position: relative">' +
-        '<copy-button style="position: absolute; top: 10px; right: 0px; z-index: 4" ></copy-button>' +
+        '<copy-button style="position: absolute; top: 10px; right: 0px; z-index: 1" ></copy-button>' +
         '<ng-transclude></ng-transclude>' +
       '</div>'
   };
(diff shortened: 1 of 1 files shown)


Kevin-K:
I signed it!

public/resources/js/directives/copy.js
```diff
@@ -25,7 +25,7 @@ angularIO.directive('copyContainer', function() {
     transclude: true,
     template:
       '<div style="position: relative">' +
-        '<copy-button style="position: absolute; top: 10px; right: 0px; z-index: 1000" ></copy-button>' +
+        '<copy-button style="position: absolute; top: 10px; right: 0px; z-index: 4" ></copy-button>' +
```
chliebel: This is not enough. Actually, there’s no reason for the z-index to be larger than 1, so you should set it to this exact value.

chliebel:
@Kevin-K Unfortunately, setting the `z-index` to 4 is not enough, because you miss the nav bar in smartphone mode:



There doesn’t seem to be any reason why `z-index` should be larger than 1. Therefore I suggest that you update your PR and set `z-index: 1`. This fixes `copy-button`.

Kevin-K:
Good catch, I didn't consider the side-nav z-indexes. 

Moved it to `z-index: 1`. While reading the tutorials I didn't notice anywhere where the copy button sits over any layered content. If it does and I missed it, setting up to `z-index: 3` will remain under the side-nav  when pushed to the `xs` layout.

PR #780, opened 2016-01-29
Update/fix url for "contribute - Angular Material"

arturparkhisenko:


DevVersion:
I'm not sure, should we really redirect the newcomers to the `material2` project? It's still in progress and `material1` is *stable*. 

So I think it would be better to create a second link. Linking to both projects. `material1` and `material2`

arturparkhisenko:
@DevVersion but first link is broken too, so i already updated it for 1st version.

DevVersion:
Okay, I created a PR because I thought you want to refer to material2. But I closed mine ;)

arturparkhisenko:
@DevVersion can you check a description for `Material2`, am i right?

public/contribute.jade
```diff
@@ -23,7 +23,14 @@
 
     p Our goal is to deliver a lean, lightweight set of Angular-based UI elements that implement the material design specification for use in Angular single-page applications (SPAs).
 
-    a(href="https://github.com/angular/material/blob/master/docs/guides/CONTRIBUTING.md" class="button" md-button) Contribute to Angular Material
+    a(href="https://github.com/angular/material/blob/master/CONTRIBUTING.md" class="button" md-button) Contribute to Angular Material
+
+  .l-sub-section
+    h3 Angular Material2
+
+    p Angular team's Material Design components built on top of Angular 2. Things are in very early stages now- check back soon for more progress.
```
DevVersion: I would rather say `Material Design Components, built on top of Angular 2 by the Angular Team is providing a huge amount of components using the official material design specification`.

But I'm not sure (I'm not a native speaker) - This should be reviewed by someone of the Angular team

arturparkhisenko: @DevVersion  i'm not  native speaker too, so we'll wait a review :smirk:

naomiblack:
I would prefer not to list Material 2 on the contribute page, as we're not set up for additional contributors, but the link fix for material 1 is good

wardbell:
Kept the link fix; ignored the rest. Thanks.

PR #790, opened 2016-02-01
Update forms.jade

techeverri:
Fix file extension typo. change "style.cs" for "style.css"


wardbell:
Thanks. Picked up your fix with PR #792

PR #797, opened 2016-02-02
docs(http): add http headers discussion/revise jsonp section

wardbell:
Two important additions/changes
1) Explain how to add headers (essential for the POST) example
2) Reword the JSONP section and re-implement the extended observables example using `Subject` instead of the `NgFormControl`.


PascalPrecht:
cc'ing @cburgdorf as he might've been interested in reviewing this.

PR #799, opened 2016-02-02
docs(http): refine observables story

wardbell:
- Abandons `Subject` in favor of `Observable.create`
- Renames chapter link in left nav to **Http Client** because no one looks for "Server Communication"
- Update the doc samples `package.json` to use `a2-in-memory-web-api` v.0.1.1, aligned with A2.b2
- Tells a better story about what `switchMap` does.

Thanks to @cburgdorf for critical feedback

### DIFF
--- a/public/docs/_examples/package.json
+++ b/public/docs/_examples/package.json
@@ -25,7 +25,7 @@
     "rxjs": "5.0.0-beta.0",
     "zone.js": "0.5.10",
     
-    "a2-in-memory-web-api": "^0.1.0",
+    "a2-in-memory-web-api": "^0.1.1",
     "bootstrap": "^3.3.6"
 
   },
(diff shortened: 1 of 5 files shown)


cburgdorf:
Looking good! I chose "server communication" since it also covers `JsonP` which uses...well not the `Http` but the `Jsonp` service. But if you can live with that I'm fine with it too ;)

wardbell:
No one could find the chapter. I can't tell you how many people said "What chapter on http?".

And since `jsonp` inherits from `http` we have a little justification there.

I decided to go with your improvements to observable before asking Rob W on the grounds that it is manifestly better guidance than what we publish now. I will get his insight and update as necessary.

cburgdorf:
:+1:

PR #800, opened 2016-02-02
docs(ts/guide/hierarchical-dependency-injection): fix typo: app/estore.service.ts -> app/restore.service.ts

michel-zimmer:


wardbell:
Thanks.

PR #809, opened 2016-02-04
docs(testing) Update Testing Pipe documentation

Gillespie59:
Pipe should implement PipeTransform interface


wardbell:
The testing chapter is slated for an overhaul soon. But I agree with this and accepting it now will remind me to get it right later.  Thanks.

PR #810, opened 2016-02-04
res.json().data doesn't work

steinsag:
I tried it locally in a slightly different context, but I could never get any result with

    res.json().data

But if I replace it with

    res.json()

everything works fine.

### DIFF
--- a/public/docs/_examples/server-communication/ts/app/toh/hero.service.1.ts
+++ b/public/docs/_examples/server-communication/ts/app/toh/hero.service.1.ts
@@ -17,7 +17,7 @@ export class HeroService {
   getHeroes () {
     return this.http.get(this._heroesUrl)
                     .toPromise()
-                    .then(res => <Hero[]> res.json().data, this.handleError)
+                    .then(res => <Hero[]> res.json(), this.handleError)
                     .then(data => { console.log(data); return data; }); // eyeball results in the console
   }
 
@@ -28,7 +28,7 @@ export class HeroService {
 
     return this.http.post(this._heroesUrl, body, options)
                .toPromise()
-               .then(res => <Hero> res.json().data)
+               .then(res => <Hero> res.json())
                .catch(this.handleError);
   }
 
--- a/public/docs/_examples/server-communication/ts/app/toh/hero.service.ts
+++ b/public/docs/_examples/server-communication/ts/app/toh/hero.service.ts
@@ -25,7 +25,7 @@ export class HeroService {
   getHeroes () {
     // #docregion http-get, http-get-v1
     return this.http.get(this._heroesUrl)
-                    .map(res => <Hero[]> res.json().data)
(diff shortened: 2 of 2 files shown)


steinsag:
I signed it!

wardbell:
What does "*I tried it locally in a slightly different context*" mean? What context is that?

The only context that matter w/r/t the `HeroService` is the chapter sample itself. I'm sorry but your claim of error and your proposed changes are simply not correct. 

The HTTP Chapter code works as demonstrated by the [plunker link](https://angular.io/resources/live-examples/server-communication/ts/plnkr.html) in the chapter itself. That plunker is generated directly from the sample

You can see that the `HeroService` code really is
```
  getHeroes () {
    return this.http.get(this._heroesUrl)
                    .map(res => <Hero[]> res.json().data)
                    .do(data => console.log(data)) // eyeball results in the console
                    .catch(this.handleError);
  }
```
If I had to guess, I'd say that you were targeting your own web api which returns response data in a different shape ... perhaps a response that lacks the wrapper object with a `data` property that *this* sample presupposes.

The chapter warns several times that you have to adjust to the actual characteristics of your API.

I'm closing until you can provide evidence that the chapter and sample are actually wrong.

steinsag:
Ok, now I see. The sample data is wrapped in a 'data' key. Mmmh, surprising and not really needed. When reading the chapter I thought I always have to call json().data so that data is not referring to the structure of the returned json.

So any reason why you got this data thing in the sample data? Why not just have:

`[
    { "id": "1", "name": "Windstorm" },
    { "id": "2", "name": "Bombasto" },
    { "id": "3", "name": "Magneta" },
    { "id": "4", "name": "Tornado" }
  ]`

PascalPrecht:
> Mmmh, surprising and not really needed. 

This is because returning `[]` directly from a JSON/REST endpoint is not valid JSON.

In other words, if you wanna be RESTful, this is needed.

steinsag:
> This is because returning `[]` directly from a JSON/REST endpoint is not valid JSON.

That's not true, see: http://stackoverflow.com/a/5034547

wardbell:
It's *valid* JSON. The problem is that it is ***insecure*** JSON on the wire.

See this [OWASP security guideline](https://www.owasp.org/index.php/OWASP_AJAX_Security_Guidelines#Always_return_JSON_with_an_Object_on_the_outside) on the subject

PascalPrecht:
Right, sorry my statement was misleading. Security is the reason. Also, see this SO post: http://stackoverflow.com/questions/3503102/what-are-top-level-json-arrays-and-why-are-they-a-security-risk

wardbell:
@steinsag Sorry for the confusion. It is easy to overlook the `data` property on the response in the sample API.

The real take-away should be: **look closely at the details of your web api *before* unpacking the result object**.

I tried to make that point. It doesn't always register.

PR #811, opened 2016-02-05
[WIP] docs(upgrade) switch from TSD to Typings

teropa:
Also switch Protractor back to JS for now. Conversion not worth it right now because
of ambient type conflicts, until Protractor ships with its own typings. (see https://github.com/angular/protractor/issues/1203#issuecomment-168777170)

Also remove some unneeded type <reference>s from examples and guide.


teropa:
Outdated

PR #820, opened 2016-02-08
docs(tutorial/toh-pt4): Remove duplicate "the"

karllindmark:


wardbell:
Thanks for your help.  We caught this in a pending round of changes so won't be needing it.

PR #821, opened 2016-02-08
Developer guides/pipes

tomplusplus:
@kwalrath and @kevmoo  Please review.

I update everything to be good with beta 3. 

The only way i could get` [(ngModel)]` to work is if I included `directives: const[COMMON_DIRECTIVES]` . I thought those were implicit when they were included in the pubspec.yaml. Thoughts?


tomplusplus:
Using `<input />` vs `<input>` didn't make a difference either way. I still had to include `COMMON_DIRECTIVES` in order for `ngModel` to work. I would point out though ( as petty as it might be ) that inputs are self closing tags so to they should have  `/>` to be clean html.

tomplusplus:
I updated it to match the plunkr version ( which didn't have any of the doc regions) I can add those in today sometime

kwalrath:
replaced by #827.

wardbell:
Are you sure about the self closing tags, @tomplusplus?

My understanding is that that is neither required nor desirable in HTML (but is required in XHTML). Picky either way

PR #823, opened 2016-02-08
docs(server-communication): e2e tests added

Dotneteer:
Here are the E2E tests added to the server-communication chapter.

### DIFF
--- a/public/docs/_examples/server-communication/e2e-spec.js
+++ b/public/docs/_examples/server-communication/e2e-spec.js
@@ -1,0 +1,173 @@
+describe('Server Communication', function () {
+
+  beforeAll(function () {
+    browser.get('');
+  });
+
+  describe('Tour of Heroes e2e tests', function () {
+    
+    var _initialHeroCount = 4;
+    var _newHeroName = 'Mr. IQ';
+    var _heroCountAfterAdd = 5;
+    
+    it('should display ' + _initialHeroCount + 'heroes after init', function () {
+      var myTohComp = element(by.tagName('my-toh'));
+      expect(myTohComp).toBeDefined('<my-toh> must exist');
+      var heroListComp = myTohComp.element(by.tagName('hero-list'));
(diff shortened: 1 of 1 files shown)


public/docs/_examples/server-communication/e2e-spec.js
```diff
@@ -0,0 +1,174 @@
...
+      var heroTags = heroListComp.all(by.tagName('li'));
+      expect(heroTags.count()).toBe(_initialHeroCount);
+    });
+    
+    it('should not add hero with empty name', function () {
+      var myTohComp = element(by.tagName('my-toh'));
+      expect(myTohComp).toBeDefined('<my-toh> must exist');
+      var addButton = myTohComp.element(by.tagName('button'));
+      expect(addButton).toBeDefined('"Add Hero" button must be defined');
+      addButton.click(function() {
```
wardbell: The following kind of test, which you use repeatedly, does not work 
```
addButton.click(function() {...})
```
The function is never executed; the test always passes.

The `click` method returns a promise. The correct syntax invokes the function within the `then` success callback:
```
addButton.click().then(function() { ... })
```

Dotneteer: Fixed. Changed all occurrences (2) to the proper pattern.

Dotneteer:
I’m the author of these changes.

wardbell:
@Dotneteer  Your claims about the empty list in wiki tests are not true and your test does not confirm it; you do not do as you say here:

>With the 'testForRefreshedResult()' method we check that the result list is immediately deleted after
      sending a key to the search field,

 I tried to come up with a test that would confirm/deny it and I can't. The old list is always populated until replaced by the new.

Therefore, I removed your comments with what seem to me to be false claims.

I agree that it is not easy to test the difference between the two wiki requests. The challenge is that we cannot know if wikipedia will respond in under 300ms.

I don't think this is worth testing right now.

PR #824, opened 2016-02-08
docs(cookbooks/component communication): new cookbook

Dotneteer:


wardbell:
I'm squashing to one commit and force pushing.

zoechi:
I saw @robwormald mentioning yesterday that `EventEmitter` should only be used for outputs. Here it's also used for events from services. Would be great to get some feedback on this. If `EventEmitter` shouldn't be used this way it shouldn't be used in the tutorials.

wardbell:
@zoechi  Correct!  This is a WIP and is not ready for press for reasons that include the one you cite.

@Dotneteer The tests were incorrect and only appeared to pass. The last two examples fail the corrected tests and are too difficult to repair today.  Feel free to try to fix them.

More important, these two recipes are too complicated and I'm not convinced they tell a good story or are good advice. Holding until I can think this through.

zoechi:
@wardbell thanks for the feedback!

Dotneteer:
I signed it!

wardbell:
@Dotneteer I really want to publish this. I made changes and am forging ahead.

 We are NOT supposed to use the `EventEmitter` for services. I replaced the EvE with the RxJS `Subject` in the *mission-control-astronaut* service example. I also show unsubscribing in the component's `ngOnDestroy` so that readers will know to beware of the memory leak potential.

I had to discard the last "message bus" recipe for three reasons:

1. It didn't adequately differentiate itself from the *mission-control-astronaut* service.

1. It didn't really demonstrate unrelated components communicating. It *looked* like a more complicated version of the previous recipe.

1. It mistakenly added the message bus to the providers array of two different components. Although it finessed the mistake by forcing both to share the same instance, this is not a good practice.

I deleted the code for this last recipe and the other disused recipes as they would otherwise be discovered and distract readers.  You can find them in your own copy of earlier commits.

Dotneteer:
Ward, thanks for your review. I will check the `EventEmitter` resolution, to see the pattern we can use in services.

wardbell:
I'm merging now. This PR will be closed.

PR #826, opened 2016-02-09
docs(dependency-injection): copy edit the TypeScript version

kwalrath:
I'll let this rest and then take another look tomorrow.

### DIFF
--- a/public/docs/ts/latest/guide/dependency-injection.jade
+++ b/public/docs/ts/latest/guide/dependency-injection.jade
@@ -1,25 +1,23 @@
 include ../../../../_includes/_util-fns
 :marked
-  **Dependency Injection** is an important application design pattern.
-  Angular has its own dependency injection framework and
+  **Dependency injection** is an important application design pattern.
+  Angular has its own dependency injection framework, and
   we really can't build an Angular application without it.
-  It's used so widely that almost everyone just calls it "DI".
+  It's used so widely that almost everyone just calls it _DI_.
 
-  In this chapter we'll learn [what DI is, why](#why-di) we want it.
+  In this chapter we'll learn what DI is and why we want it.
   Then we'll learn [how to use it](#angular-di) in an Angular app.
-  
-  All code in this chapter is available as a 
-  [live example](/resources/live-examples/dependency-injection/ts/plnkr.html)
-  on the web.
-    
+
+  [Run the live example](/resources/live-examples/dependency-injection/ts/plnkr.html)
+
 <a id="why-di"></a>
 .l-main-section
 :marked
-  ## Why Dependency Injection?
+  ## Why dependency injection?
 
   Let's start with the following code.
 
-+makeExample('dependency-injection/ts/app/car/car-no-di.ts', 'car', 'app/car/car.ts (no di)')
++makeExample('dependency-injection/ts/app/car/car-no-di.ts', 'car', 'app/car/car.ts (without DI)')
 
 :marked
   Our `Car` creates everything it needs inside its constructor.
@@ -40,12 +38,12 @@ include ../../../../_includes/_util-fns
   when the definion of `Engine` changes, our `Car` class must change.
   That makes `Car` brittle.
 
-  What if we want to put a different brand of tires on our `Car`. Too bad.
+  What if we want to put a different brand of tires on our `Car`? Too bad.
   We're locked into whatever brand the `Tires` class creates. That makes our `Car` inflexible.
 
   Right now each new car gets its own engine. It can't share an engine with other cars.
   While that makes sense for an automobile engine,
-  we can think of other dependencies that should be shared ... like the onboard
+  we can think of other dependencies that should be shared... like the onboard
   wireless connection to the manufacturer's service center. Our `Car` lacks the flexibility
   to share services that have been created previously for other consumers.
 
@@ -55,23 +53,23 @@ include ../../../../_includes/_util-fns
   Will a new instance of `Engine` make an asynchronous call to the server?
   We certainly don't want that going on during our tests.
 
-  What if our `Car` should flash a warning signal when tire pressure is low.
+  What if our `Car` should flash a warning signal when tire pressure is low?
   How do we confirm that it actually does flash a warning
   if we can't swap in low-pressure tires during the test?
 
   We have no control over the car's hidden dependencies.
   When we can't control the dependencies, a class become difficult to test.
 
-  How can we make `Car` more robust, more flexible, and more testable?
+  How can we make `Car` more robust, flexible, and testable?
 
-  That's super easy. We probably already know what to do. We change our `Car` constructor to this:
+  That's super easy. We probably already know what to do. We change our `Car` constructor to a version with DI:
 
   <a id="ctor-injection"></a>
 
 +makeTabs(
   'dependency-injection/ts/app/car/car.ts, dependency-injection/ts/app/car/car-no-di.ts',
-  'car-ctor, car-ctor', 
-  'app/car/car.ts DI (constructor), app/car/car.ts No DI (constructor)')(format=".")
+  'car-ctor, car-ctor',
+  'app/car/car.ts (excerpt with DI), app/car/car.ts (excerpt without DI)')(format=".")
 
 :marked
   See what happened? We moved the definition of the dependencies to the constructor.
@@ -84,7 +82,7 @@ include ../../../../_includes/_util-fns
 :marked
   Now we create a car by passing the engine and tires to the constructor.
 
-+makeExample('dependency-injection/ts/app/car/car-creations.ts', 'car-ctor-instantiation', 'Simple Car')(format=".")
++makeExample('dependency-injection/ts/app/car/car-creations.ts', 'car-ctor-instantiation', '')(format=".")
 
 :marked
   How cool is that?
@@ -95,10 +93,10 @@ include ../../../../_includes/_util-fns
   If someone extends the `Engine` class, that is not `Car`'s problem.
 .l-sub-section
   :marked
-    The consumer of `Car` has the problem. The consumer must update the car creation code to
-    something like:
+    The _consumer_ of `Car` has the problem. The consumer must update the car creation code to
+    something like this:
 
-  +makeExample('dependency-injection/ts/app/car/car-creations.ts', 'car-ctor-instantiation-with-param', 'Super Car')(format=".")
+  +makeExample('dependency-injection/ts/app/car/car-creations.ts', 'car-ctor-instantiation-with-param', '')(format=".")
 
   :marked
     The critical point is this: `Car` itself did not have to change.
@@ -110,10 +108,10 @@ include ../../../../_includes/_util-fns
   We can pass mocks to the constructor that do exactly what we want them to do
   during each test:
 
-+makeExample('dependency-injection/ts/app/car/car-creations.ts', 'car-ctor-instantiation-with-mocks', 'Test Car')(format=".")
++makeExample('dependency-injection/ts/app/car/car-creations.ts', 'car-ctor-instantiation-with-mocks', '')(format=".")
 
 :marked
-  **We just learned what Dependency Injection is**.
+  **We just learned what dependency injection is**.
 
   It's a coding pattern in which a class receives its dependencies from external
   sources rather than creating them itself.
@@ -125,6 +123,7 @@ include ../../../../_includes/_util-fns
   We need something that takes care of assembling these parts for us.
 
   We could write a giant class to do that:
+  <!-- PENDING: add prominent comment in code that this is a bad pattern? -->
 
 +makeExample('dependency-injection/ts/app/car/car-factory.ts', null, 'app/car/car-factory.ts')
 
@@ -137,7 +136,7 @@ include ../../../../_includes/_util-fns
   Wouldn't it be nice if we could simply list the things we want to build without
   having to define which dependency gets injected into what?
 
-  This is where the Dependency Injection Framework comes into play.
+  This is where the dependency injection framework comes into play.
   Imagine the framework had something called an `Injector`.
   We register some classes with this `Injector` and it figures out how to create them.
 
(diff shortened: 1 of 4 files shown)


public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -1,25 +1,23 @@
 include ../../../../_includes/_util-fns
 :marked
-  **Dependency Injection** is an important application design pattern.
-  Angular has its own dependency injection framework and
+  **Dependency injection** is an important application design pattern.
```
zoechi: Dependency Injection is a name of a pattern not an action. Shouldn't this be written with uppercase as it was? https://de.wikipedia.org/wiki/Dependency_Injection

kwalrath:
@wardbell this is ready to review.

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -40,12 +38,12 @@ include ../../../../_includes/_util-fns
...
   That makes `Car` brittle.
 
-  What if we want to put a different brand of tires on our `Car`. Too bad.
+  What if we want to put a different brand of tires on our `Car`? Too bad.
   We're locked into whatever brand the `Tires` class creates. That makes our `Car` inflexible.
 
   Right now each new car gets its own engine. It can't share an engine with other cars.
   While that makes sense for an automobile engine,
-  we can think of other dependencies that should be shared ... like the onboard
+  we can think of other dependencies that should be shared... like the onboard
```
wardbell: Why remove the space before the ellipsis? 
Maybe it should read "we can think of other dependencies that should be shared such as the onboard"

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -84,7 +82,7 @@ include ../../../../_includes/_util-fns
 :marked
   Now we create a car by passing the engine and tires to the constructor.
 
-+makeExample('dependency-injection/ts/app/car/car-creations.ts', 'car-ctor-instantiation', 'Simple Car')(format=".")
++makeExample('dependency-injection/ts/app/car/car-creations.ts', 'car-ctor-instantiation', '')(format=".")
```
wardbell: Why remove the title 'Simple Car'?  Doing so makes it difficult to relate this code to the derived "Super Car" coming up soon.

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -95,10 +93,10 @@ include ../../../../_includes/_util-fns
   If someone extends the `Engine` class, that is not `Car`'s problem.
 .l-sub-section
   :marked
-    The consumer of `Car` has the problem. The consumer must update the car creation code to
-    something like:
+    The _consumer_ of `Car` has the problem. The consumer must update the car creation code to
+    something like this:
 
-  +makeExample('dependency-injection/ts/app/car/car-creations.ts', 'car-ctor-instantiation-with-param', 'Super Car')(format=".")
+  +makeExample('dependency-injection/ts/app/car/car-creations.ts', 'car-ctor-instantiation-with-param', '')(format=".")
```
wardbell: Without the `Super Car` title, I can't tell that this is a _different_ class - a _derived_ class - with a different implementation.

Why do you think these particular changes are worthwhile?

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -207,83 +205,87 @@ include ../../../../_includes/_util-fns
...
   We don't have to create an Angular injector.
   Angular creates an application-wide injector for us during the bootstrap process.
 
-+makeExample('dependency-injection/ts/app/main.ts', 'bootstrap', 'app/main.ts (bootstrap)')(format='.')
++makeExample('dependency-injection/ts/app/main.ts', 'bootstrap', 'app/main.ts (excerpt)')(format='.')
 :marked
-  We do have to configure the injector by registering the **providers** 
-  that create the services we need in our application.
+  We do have to configure the injector by registering the **providers**
+  that create the services that our application needs.
```
wardbell: That's a lot of "that"s.

Maybe we can say "We do have to configure the injector by registering the **providers**  that create the services our application requires."

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -207,83 +205,87 @@ include ../../../../_includes/_util-fns
...
   We'll explain what [providers](#providers) are later in this chapter.
   Before we do, let's see an example of provider registration during bootstrapping:
-  
+  <!-- Again we show an example of bad code. Ew. I see we quickly explain
+  what you should do instead, but people might just copy the code and miss the explanation. -->
```
wardbell: If they blindly copy it, presumably they copy the comment with it. One can only hope that a code reviewer says "What is _that_?"

Unlike other examples of "bad" code, this example is merely discourage. It still works. Believe it or not, this was once the _encouraged_ practice ;-)

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -207,83 +205,87 @@ include ../../../../_includes/_util-fns
...
   The injector now knows about our `HeroService`.
   An instance of our `HeroService` will be available for injection across our entire application.
 
   Of course we can't help wondering about that comment telling us not to do it this way.
-  It *will* work. It's just not a best practice. 
+  It *will* work. It's just not a best practice.
   The bootstrap provider option is intended for configuring and overriding Angular's own
-  pre-registered services.
-  
+  pre-registered services. <!-- such as what? -->
```
wardbell: ""  ... we don't want to go down that rabbit hole now. 

If you're curious, a good example is changing the router's default location strategy:
```
bootstrap(AppComponent, [
  ROUTER_PROVIDERS,
  provide(LocationStrategy,
         {useClass: HashLocationStrategy})
])
```

public/docs/ts/latest/guide/dependency-injection.jade
kwalrath: It looked wrong to me with spaces before & after, but (upon further research) it's not.

In general, I usually use ellipses in very informal communication, and em dashes in docs. (I think ellipses sound wishy washy, which I am IRL but not when I write tech docs. :) But I'm fine with using ellipsis here, with whatever spacing you like.

I'll undo this change.

public/docs/ts/latest/guide/dependency-injection.jade
kwalrath: In all the pages I looked at before, we were using this title to specify the filename and whether it's an excerpt or a specific version.

It seemed weird to have a different convention here.

public/docs/ts/latest/guide/dependency-injection.jade
kwalrath: Your suggestion is better.

public/docs/ts/latest/guide/dependency-injection.jade
kwalrath: I like to use CAPITAL LETTERS in bad code comments. :) Because you can't shout loudly enough that this is BAD CODE.

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -336,47 +338,47 @@ include ../../../../_includes/_util-fns
...
   app/heroes/hero.service (v.1)`)
 
 :marked
   The constructor now asks for an injected instance of a `Logger` and stores it in a private property called `_logger`.
-  We call that property within our `getHeroes()` method when anyone asks for heroes.
+  We call that property within our `getHeroes` method when anyone asks for heroes.
 
   <a id="injectable"></a>
-  ### Why *@Injectable*?
+  ### Why @Injectable?
```
wardbell: I'd prefer the italics around @Injectable. No biggie

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -336,47 +338,47 @@ include ../../../../_includes/_util-fns
...
+    li <b>Future proofing:</b> No need to remember <code>@Injectable</code> when we add a dependency later.
+    li <b>Consistency:</b> All services follow the same rules, and we don't have to wonder why a decorator is missing.
 .l-sub-section
   :marked
     The `HeroesComponent` has an injected dependency too. Why don't we add `@Injectable()` to the `HeroesComponent`?
-    
+
     We *can* add it if we really want to. It isn't necessary because the `HeroesComponent` is already decorated with `@Component`.
-    TypeScript generates metadata for *any* class with a decorator and *any* decorator will do.
+    TypeScript generates metadata for *any* class with a decorator, and *any* decorator will do.
```
wardbell: You are the comma queen.  I'm not sure why you want the reader to trip over that comma on the way to "and". ;-)

public/docs/ts/latest/guide/dependency-injection.jade
kwalrath: I have no taste, but I've heard from people with taste that italics with boldface is a design no-no. Also, I've heard from real live readers who have trouble reading italics, so I tend to avoid them (italics, not readers).

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -426,40 +428,40 @@ code-example(format).
...
-  We must register a service *provider* with the injector or it won't know how to create the service. 
-  
   Earlier we registered the `Logger` service in the `providers` array of the metadata for the `AppComponent` like this:
 +makeExample('dependency-injection/ts/app/providers.component.ts','providers-logger')
 :marked
-  The `providers` array appears to hold a service class. 
-  In reality it holds an instance the [Provider](../api/core/Provider-class.html) class that can create that service.
- 
+  The `providers` array appears to hold a service class.
+  In reality it holds an instance of the [Provider](../api/core/Provider-class.html) class that can create that service.
```
wardbell: I vaguely recall that we need a more complete link to API guide chapters if this text is to be re-used in a Dart or JS document. Do you remember what the formula is? I don't.

public/docs/ts/latest/guide/dependency-injection.jade
kwalrath: The comma makes it easier to avoid having to reread the sentence, due to incorrectly combining the "and" with one of the two phrases that come before it.

It generates metadata for any class [that has a decorator and for any decorator]... no wait!

It generates metadata for any class that has [a decorator and any decorator]... no!

It generates metadata for any class that has a decorator, and any decorator will do. (harder to misread that!)

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -592,55 +595,56 @@ code-example(format).
...
+
+  In all previous examples, the dependency value has been a class *instance*, and
   the class *type* served as its own lookup key.
-  Here we get a `HeroService` directly from the injector by supplying the `HeroService` type as the key/token.
+  Here we get a `HeroService` directly from the injector by supplying the `HeroService` type as the token:
 +makeExample('dependency-injection/ts/app/injector.component.ts','get-hero-service')(format='.')
 :marked
-  We have similar good fortune (in typescript) when we write a constructor that requires an injected class-based dependency.
-  We define a constructor parameter with the `HeroService` class type and Angular knows to inject the
+  We have similar good fortune when we write a constructor that requires an injected class-based dependency.
```
wardbell: I suppose this is true in both TS and Dart. In JS we'd have to annotate to tell Angular the service/provider lookup token.

public/docs/ts/latest/guide/dependency-injection.jade
wardbell: Maybe we should go with neither ellipsis nor dash but rather the full, unbroken sentence?
See my comment.

public/docs/ts/latest/guide/dependency-injection.jade
wardbell: Good point. But this passage makes no sense at all if we can't tell that we're referring to two different classes, one inheriting from the other.

If we must we can say 'app/car/car-creations.ts (SimpleCar)' and 'app/car/car-creations.ts (SuperCar)' ... or whatever is technically correct.

public/docs/ts/latest/guide/dependency-injection.jade
wardbell: Feel free to modify the sample itself accordingly.

public/docs/ts/latest/guide/dependency-injection.jade
wardbell: I'm in big trouble then because I've adopted the practice of using italics in these headers when referring to code objects.  The <code> tags (and backtics) are less readable in headers.

I felt we should do *something* to differentiate a code symbol from mere text.

We are bike shedding ... but I'd really like to settle on something.  Do you really dislike italics?

wardbell:
LGTM (where G=Great)

public/docs/ts/latest/guide/dependency-injection.jade
kwalrath: Is it really that important to have different fonts in headers?

Most places I've worked, we use only plain old bold font in headers (no code font, italics, or other special cases). Some of that might have been due to tool limitations, but it might have been a purely esthetic decision.

In this particular case, the @ and the capital I make this unmistakably a decorator. And I tend to thing of italics as variables (e.g. <code>cp <em>file1 file2</em></code>).

All that said, I don't care too strongly about this. It's my knee-jerk reaction to avoid italics in headings and question them elsewhere... but I can control my knee.

public/docs/ts/latest/guide/dependency-injection.jade
kwalrath: Looking around in template-syntax.jade, it appears that `../api/` -> `/docs/ts/latest/api/`.

public/docs/ts/latest/guide/dependency-injection.jade
kwalrath: But it's _not_ a derived or different class. It's just two different ways of calling the same constructor.

Either way, I don't see how the title helps you grasp the difference. Maybe if I highlighted the constructor call, that would be more helpful.

public/docs/ts/latest/guide/dependency-injection.jade
kwalrath: I'm adding a tiny bit of explanation, since not having one distracted me.

kwalrath:
I'm going to merge this, but of course we can always change it in the future...

PR #834, opened 2016-02-12
chore(devguide/ts): to beta.6 + typings + related doc updates

wardbell:


alexeagle:
Thanks for the hard work on this release, Ward.
I'm a little disappointed that the "quickstart" has become much longer and less "quick" (see eg. https://github.com/angular/angular/issues/6536 where users point out that it is already taking much more than five minutes)
But I know you understand the alternatives and trust your judgement on it.

PR #838, opened 2016-02-13
[WIP] docs(quickstart): condense the QS with expandos

wardbell:
Maybe QuickStart will actually be quick
Todo (later): some professional styling and animation.

### DIFF
--- a/public/docs/ts/latest/guide/typescript-configuration.jade
+++ b/public/docs/ts/latest/guide/typescript-configuration.jade
@@ -1,0 +1,151 @@
+include ../../../../_includes/_util-fns
+
+:marked
+  TypeScript is a primary language for Angular application development.
+  
+  TypeScript is a dialect of JavaScript with design-time support for type-safety and tooling.
+  
+  Browsers can't execute TypeScript directly. It has to be "transpiled" into JavaScript with the *tsc* compiler
+  and that effort requires some configuration.
+  
+  This chapter covers some aspects of TypeScript configuration and the TypeScript environment
+  that are important to Angular developers.
+
+a(id="tsconfig")
+.l-main-section
+:marked
+  ## *tsconfig.json*
+  We typically add a TypeScript configuration file (`tsconfig.json`) to our project to
+  guide the compiler as it generates JavaScript files.
+.l-sub-section
+  :marked
+    Get details about `tsconfig.json` from the official
+    [TypeScript wiki](https://github.com/Microsoft/TypeScript/wiki/tsconfig.json).
+:marked
+  We created the following `tsconfig.json` for the [QuickStart](../quickstart.html):
++makeJson('quickstart/ts/tsconfig.1.json', null, 'tsconfig.json')(format=".")
+:marked
+  The options and flags in this file are essential for Angular 2 applications.
+  
+  <a id="noImplicitAny"></a>
+  ### *noImplicitAny* and *suppressImplicitAnyIndexErrors*
+
+  TypeScript developers disagree about whether the `noImplicitAny` flag should be `true` or `false`.
+  There is no correct answer and we can change the flag later.
+  But our choice now can make a difference in larger projects so it merits discussion.
+
+  When the `noImplicitAny` flag is `false` (the default),
+  the compiler silently defaults the type of a variable to `any` if it cannot infer
+  the type based on how the variable is used. That's what we mean by *implicit `any`*.
+
+  We initialized the `noImplicitAny` flag to `false` in the QuickStart 
+  to make learning TypeScript development easier.
+  
+  When the `noImplicitAny` flag is `true` and the TypeScript compiler cannot infer
+  the type, it still generates the JavaScript files. But it also **reports an error**.
+  Many seasoned developers prefer this stricter setting because type checking catches more
+  unintentional errors at compile time.
+  
+  We can set a variable's type to `any` even when the `noImplicitAny` flag is `true`.
+  We do so when that seems like the best choice for the situation, 
+  deliberately and explicitly, after giving the matter some thought.
+
+  If we set the `noImplicitAny` flag to `true`, we may get *implicit index errors* as well.
+  Most developers feel that *this particular error* is more annoying than helpful.
+  We can suppress them with the following additional flag.
+code-example(format=".").
+  "suppressImplicitAnyIndexErrors":true
+
+
+a(id="typings")
+.l-main-section
+:marked
+  ## TypeScript Typings
+  Many JavaScript libraries such as jQuery, the Jasmine testing library, and Angular itself,
+  extend the JavaScript environment with features and syntax
+  that the TypeScript compiler doesn't recognize natively.
+  When the compiler doesn't recognize something, it throws an error. 
+  
+  We use [TypeScript type definition files](http://www.typescriptlang.org/Handbook#writing-dts-files)
+  &mdash; *d.ts files* &mdash; to tell the compiler about the libraries we load.
+  
+  TypeScript-aware editors leverage these same definition files to display type information about library features.
+  
+  Many libraries include their definition files in their npm packages where both the TypeScript compiler and editors
+  can find them. Angular is one such library. 
+  Peek into the `node_modules/angular2/` folder of any Angular application to see several `...d.ts` files that describe parts of Angular.
+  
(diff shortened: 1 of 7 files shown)


johnpapa:
love it. much more readable and consumable. ship it!

public/docs/ts/latest/guide/typescript-configuration.jade
```diff
@@ -0,0 +1,148 @@
+include ../../../../_includes/_util-fns
+
+:marked
+  TypeScript is a primary language for Angular application development.
```
filipesilva: 'is the' instead of 'is a'

public/docs/ts/latest/guide/typescript-configuration.jade
```diff
@@ -0,0 +1,148 @@
...
+  
+  TypeScript is a dialect of JavaScript with design-time support for type-safety and tooling.
+  
+  Browsers can't execute TypeScript directly. It has to be "transpiled" into JavaScript with the *tsc* compliler
```
filipesilva: 'compiler' instead of 'compliler'

public/docs/ts/latest/guide/typescript-configuration.jade
```diff
@@ -0,0 +1,148 @@
...
+  **We do nothing to get *typings* files for library packages with bundled *d.ts* files.**
+  
+  ### Manual typings
+  Sadly, many libraries &mdash; jQuery, Jasmine, and Lodash among them &mdash; do *not* include `d.ts` files in their npm packages.
+  Fortunately, either their authors or community contributors have created separate *d.ts* files for these libraries and 
+  published them in well-known locations.
+  The *typings* tool can find and fetch these files for us.
+  
+  We installed the [typings](https://github.com/typings/typings/blob/master/README.md) tool 
+  with npm (it's listed among the *DevDependencies* in the `package.json`) and added an npm script
```
filipesilva: 'devDependencies' instead of 'DevDependencies'

filipesilva:
Looks great!

public/docs/ts/latest/guide/typescript-configuration.jade
wardbell: I really wanted it to be "a", not "the".  Don't want to offend others

wardbell:
thanks @filipesilva; updated w/ your changes (except as noted)

naomiblack:
I like where you're headed with the expandos, but I'd like to spend some design time on making them look good. The "Why" button isn't quite right. I don't want to merge this in without some professional styling and animation -- but the mechanics of it are great. Let's hold off on this PR for a bit.

The banner align on the resize is now off-kilter with respect to the sidenav buttons. The banner is already on my design to-do list. I've nudged it as far as I'm comfortable with for now. Please don't spend more time on banner adjusts.

image

public/docs/ts/latest/quickstart.jade
```diff
@@ -1,270 +1,505 @@
...
-  
-  Technically, a component is a class that controls a view template.
-  We'll write a lot of them as we build Angular apps. This is our first attempt
-  so we'll keep it ridiculously simple.
-  
-  ### Create an application source sub-folder
-  
-  We like to keep our application code in a sub-folder off the root called `app/`.
-  Execute the following command in the console window.
+button(onclick='why("what-is-a-component")' class='button button-small') Why
```
kwalrath: This "why" needs to be preceded by an action. Maybe:

Let's create a super simple Angular component.

[WHY]

public/docs/ts/latest/quickstart.jade
```diff
@@ -1,270 +1,505 @@
...
-
-.l-sub-section
-  :marked
-    Learn why we import `bootstrap` from `angular2/platform/browser`
-    and why we create a separate *main.ts* file in the [appendix below](#main).
-:marked
-  We've asked Angular to launch the app in a browser with our component at the root.
-  Where will Angular put it? 
-
+button(onclick='why("inside-main")' class='button button-small') Why
```
kwalrath: This is a *really* long "WHY" segment. Also, it's unclear what "WHY" refers to. Maybe split it up into a few WHYs, and make sure that the most important stuff is always visible. For example:

At the top of the file, you import the Component class.

[WHY]

Next, you create a class that has an @Component decorator.

[WHY]

The @Component decorator specifies a selector—the name of the HTML tag you'll use (`<my-app>`).

[WHY]
...

kwalrath:
Once we get the content and look closer to done, we should do a copy edit.

public/docs/ts/latest/quickstart.jade
wardbell: That bothered me too. I'll fix.

wardbell:
I abdicate on styling. I have spent almost no time on it and will not spend more.

My concern about delay is new-comers to A2 aren't getting the information they need. Are we letting the better be the enemy of the good?

Looking into Kathy's suggestions now.

naomiblack:
Give me a couple of days, Ward, I'm on it.

I hear you, but I don't want to just throw a button up there.

wardbell:
I'm chillin'!

wardbell:
I discarded the style changes and force pushed. Those bad boys are gone forever :-)

public/docs/ts/latest/quickstart.jade
wardbell: I've rewritten this segment in the spirit of your suggestion. Hope you think it is an improvement.

naomiblack:
Some issues that still need to be resolved:
- You're using a button here to open the expanded content, but once it's expanded, the button label is still "why" -- that doesn't make any sense with the text open. The label on the button should change when the text is expanded. or try using md-dialog for this instead? 
- instead of button, use `md-button(class='md-primary')`
- align the button left a bit farther so it is flush with the code examples
- when the text section is open, add some more padding below it. It's not clear where the expanded bit finishes and the next bit of body text begins.

wardbell:
Thanks, @naomiblack 

The most significant question is "have I gone too far"?  I don't know how to resolve that one. Maybe we can find some time to talk it through?  Or I can always unwind it completely and go back to a long, flat page.

Do you agree there is/was a problem to be solved? I was getting strong push back on how not-quick this was  (e.g.,from the visitors last week). The new material on typings that arrived last week just pushed it over the edge for me. But I confess I have no direct line to newbies who could tell me which QS style is better.

Meanwhile ... there are substantive changes here that aren't in the published QS. They need to go to press one way or the other.

What can I do to make this right and get it unstuck?

### Regarding other points

I completely rewrote the gigantic AppComponent block (another excellent @kwalrath suggestion). I think it is more clear and in more digestible pieces.

I agree with removing the tiny expandos.

I like your labels and button stylings ... let's go with those! 

Easy for me to change the button text when the expando is open (I wrote the JS for it) . Do you have a suggestion for how it should read? For example, what should "*Why do we need a main.ts?*" change to when its content is open?

I can adjust the indents and padding. I'm pretty helpless when it comes to the other styling concerns, especially colors. Not my forte (you noticed). I'm inclined to think that some of those improvements could come later when your designer becomes available.

wardbell:
### Brief/Verbose Mode?

Maybe I overdid it with the fine grained approach.  What do you think of a single switch (checkbox) - Brief/Verbose Mode - and we killed all of the little expandos.  Could fall back to the original formatting instead of tucking all expanded text in `.l-sub-section`.

That would get us unstuck and out of the styling/labeling challenges.

kwalrath:
Single switch seems fine, as long as you can see when you're missing out on a discussion. But if you can't... then you don't know what you're missing. I'd rather have a companion page than an all-or-nothing, see-it-or-don't switch.

I suppose a companion page would be an interesting test. We could use analytics to see whether people ever go there...

naomiblack:
Until I have a real web designer I can lean on to improve the interaction problems of how the expando works, I think a single switch might be the way to go for now. We could apply some light styling, maybe create a new `.l-verbose-section` class to apply some light shading or other minimal visual indicator to show the diffs. And then give readers a "jump to the brief version" escape hatch back from the verbose one.
I like the way we do this on our internal codelab docs, but I don't have the time or skill to build it. Would rather take a simple approach to the immediate need to simplify the QS, and then iterate on doing expandos later.

wardbell:
See the new single-toggle version #863

wardbell:
We will do expandos but not this way and with the latest content. Keeping the branch alive but closing this PR

PR #839, opened 2016-02-15
(bug) Code copy icon floats over website header

webmaxru:
Code copy icon floats over the website header because `z-index: 1000` is hardcoded in `<copy-button>` directive. The fix is to remove `.l-layer-5` from .main-nav and add `z-index: 1001` to it.


chliebel:
Hi @webmaxru, this unfortunately misses the nav bar in smartphone mode:

copy-button

#769 resolves this in a more elegant way (in my opinion), as it changes the copy-button’s `z-index` property only and thus is less invasive.

wardbell:
closed in favor of #769

PR #843, opened 2016-02-16
docs(cookbooks/dependency-injection): document and samples ready for QA

Dotneteer:


wardbell:
Thanks, Istvan! Closing because we have one now.

PR #849, opened 2016-02-17
Fix examples of typings npm script

joshuacc:
To pass arguments to an npm script, a `--` separator is necessary.

This is resulting in incorrect bug reports on the typings project: https://github.com/typings/typings/issues/220

### DIFF
--- a/public/docs/ts/latest/quickstart.jade
+++ b/public/docs/ts/latest/quickstart.jade
@@ -538,7 +538,7 @@ code-example(format="").
     The following command installs the typings file for the Jasmine test library and updates the `typings.config` 
     so we that we get it automatically the next time.
 code-example(format="").
-  npm run typings -- install -- jasmine --ambient --save
+  npm run typings -- install jasmine --ambient --save
 .l-sub-section
   :marked
     Learn about the features of the *typings* tool at its [site on github](https://github.com/typings/typings/blob/master/README.md).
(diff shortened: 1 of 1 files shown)


joshuacc:
I've signed the CLA.

joshuacc:
Actually, it appears that the examples were actually correct. I'd propose using the more consistent style in this PR to avoid confusion, however. Namely always putting `--` after `npm run typings`.

wardbell:
Accepting this PR would not have prevented the original error that spawned the typings issue; see my comments there. The errant command would still have failed.

There is a limit to how much we can teach the community tools such as npm. That said, I think we could do better by drawing attention to the significance of the -- argument and linking to the npm guidance.

That may be how I address the problem rather than substituting one mysterious command for another.

I do appreciate the heads up, though!

joshuacc:
@wardbell I realize that you have to balance a lot of different concerns in the Angular docs. (Great job on them, BTW.)

I do think that drawing attention to the `--` and consistently using it as a separator between the npm script and all subsequent arguments would make it easier for beginners to internalize, as @blakeembrey mentioned in the typings thread.

wardbell:
I agree and will do

wardbell:
Ok ... I've thought about it more and incorporated all of these suggestions in a new PR #849.

I also incorporated it in PR #838 which is my preferred version of the QuickStart.

These typings changes will arrive in one or the other PR.

We don't need THREE PRs on this subject so I'm closing this one.  Thanks for the help.

joshuacc:
Awesome. Thanks @wardbell!

PR #873, opened 2016-02-22
docs(upgrade) switch from TSD to Typings

teropa:
Also switch Protractor back to JS for now. Conversion not worth it because
of ambient type conflicts, until Protractor ships with its own typings.

Also remove some unneeded type <reference>s from examples and guide.


filipesilva:
If you're interested, I did a setup of protractor in TS without any ambient type conflicts.

https://github.com/angular/angular-cli/pull/228/files

PR #879, opened 2016-02-23
docs(template-syntax/Dart): TrackBy code + fixes.

thso:
Tried to stay as close as possible to the ts version, while showcasing nice features of Dart.

  - Implementation of elements change counters used for the comparison of no `TrackBy` vs proper usage of `TrackBy`,
  - Fixed issues that prevented from compiling, 
  - General style cleanup.

Can you take look @kwalrath ?

### DIFF
--- a/public/docs/_examples/template-syntax/dart/lib/app_component.dart
+++ b/public/docs/_examples/template-syntax/dart/lib/app_component.dart
@@ -19,7 +19,17 @@ enum Color { Red, Green, Blue }
       MyClickDirective,
       MyClickDirective2
     ])
-class AppComponent {
+class AppComponent implements OnInit, AfterViewInit {
+  @override
+  ngOnInit() {
+    refreshHeroes();
+  }
+
+  @override
+  ngAfterViewInit() {
+    _detectNgForTrackByEffects();
+  }
+
   String heroName;
   String help;
   String actionName = 'Go for it';
@@ -35,9 +45,18 @@ class AppComponent {
   bool isUnchanged = true;
   bool isSelected = false;
   Color color = Color.Red;
-  List<Hero> heroes = Hero.MockHeroes;
-  Hero selectedHero = Hero.MockHeroes[0];
-  Hero currentHero = Hero.MockHeroes[0];
+
+  List<Hero> heroes;
+  Hero currentHero;
+
+  // #docregion refresh-heroes
+  /// Updates [this.heroes] with fresh set of cloned heroes.
+  refreshHeroes() {
+    heroes = mockHeroes.map((hero) => hero.clone()).toList();
+    currentHero = heroes[0];
+  }
+  // #enddocregion refresh-heroes
+
   final Hero nullHero = null;
   Map product = {'name': 'frimfram', 'price': 42};
   FormElement form;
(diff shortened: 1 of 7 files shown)


public/docs/_examples/template-syntax/dart/lib/hero_detail_component.dart
```diff
@@ -56,18 +57,17 @@ class HeroDetailComponent {
...
+        <div>Last: {{hero?.lastName}}</div>
+        <div>Birthdate: {{hero?.birthdate | date:'longDate'}}</div>
+        <div>Web: <a href="{{hero?.url}}" target="_blank">{{hero?.url}}</a></div>
+        <div>Rate/hr: {{hero?.rate | currency:'EUR'}}</div>
+        <br clear="all">
+        <button (click)="delete()">Delete</button>
+      </div>''')
 class BigHeroDetailComponent extends HeroDetailComponent {
   // #docregion input-output-1
   @Input() Hero hero;
```
kwalrath: Do you think we should have `@override` here?

kwalrath:
We have some no-return methods that aren't marked void. Should we say their return type is void? I'm thinking of delete() in [Big]HeroDetailComponent, and ng*Init() in AppComponent.

kwalrath:
We'd talked about changing the `ngSwitchWhen` templates (in app_component.html) to not have the `[...]`. For example, we'd change this:

```
<template [ngSwitchWhen]="'Eenie'"><span>Eenie</span></template>
```
to this:
```
<template ngSwitchWhen="Eenie"><span>Eenie</span></template>
```

The former works, but it's ugly.

thso:
> We have some no-return methods that aren't marked void. Should we say their return type is void? I'm thinking of delete() in [Big]HeroDetailComponent, and ng*Init() in AppComponent.

Yes, we should add return types to every function definition as it default to dynamic, which is bad.

public/docs/_examples/template-syntax/dart/lib/hero_detail_component.dart
thso: Yup, I'm on it.

thso:
I agree `<template ngSwitchWhen="Eenie"><span>Eenie</span></template>` is cleaner.

wardbell:
If hand coded it would be: `<span ngSwitchWhen="Eenie">Eenie</span>`

wardbell:
The key is that one should use `foo="some-string"` when you're NOT binding and reserve `[foo]="someProperty"` for property binding scenarios in which the value must be watched.

kwalrath:
@thso: Yes, please change it to `<template ngSwitchWhen="Eenie"><span>Eenie</span></template>` is the upshot. (Right, @wardbell?)

public/docs/_examples/template-syntax/dart/lib/app_component.dart
```diff
@@ -19,7 +19,17 @@ enum Color { Red, Green, Blue }
       MyClickDirective,
       MyClickDirective2
     ])
-class AppComponent {
+class AppComponent implements OnInit, AfterViewInit {
+  @override
+  ngOnInit() {
```
zoechi: The lifecycle methods don't have return types (here and below) is this intentional?

wardbell:
Well .. that's what I did and why. I would think the same reasoning would be persuasive in Dart world but not sure. That's your judgement call

kwalrath:
LGTM. Thanks, @thso!

PR #882, opened 2016-02-24
wip: table of contents draft

PascalPrecht:


PascalPrecht:
@wardbell you'll notice that all i've done is adding a heading for the section list. I think I was lucky that the template syntax guide is already strucutred in a way that it works very well. To me it feels like add the "Table Of Contents" header makes a huge difference.

Of course, it'll take a bit more work to apply that to other chapters.

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -29,9 +31,6 @@ include ../../../../_includes/_util-fns
   * [Template expression operators](#expression-operators)
 // #enddocregion intro
 
-:marked
-  [Run the live example](/resources/live-examples/template-syntax/ts/plnkr.html)
```
PascalPrecht: Notice that I've removed the "Live example" here because.... Which live example are we even talking about? I find this rather misplaced.

wardbell: Disagree about removing the example which is enormously valuable in exploring the ideas expressed here and closely tracks the ToC here.

But I do agree that the purpose should be made clear. Doing that.

wardbell:
Incorporating this change. Opened companion issue #935

PR #886, opened 2016-02-25
Wrong folder structure

datoml:
Seems like the tabs are wrong in the folder structure after this "After these changes, the folder structure looks like this:" in https://angular.io/docs/ts/latest/guide/router.html.

app.component.ts, crisis-list.component.ts and main.ts should stay in the app folder.


datoml:
I signed it!

wardbell:
Good catch. Made the slight mod of moving the files below the folders ... as one expects in a file viewer.

PR #887, opened 2016-02-25
[WIP] chore: temporary TS-referencing chapters for gaps in JS/Dart

wardbell:


naomiblack:
I like the new switcher. Make it so!

kwalrath:
Does this point to the Dart sample when one exists? That's an additional link that's helpful to our devs.

wardbell:
The gameplan is to ensure that there is always a corresponding page in each of the language tracks, even if that page can't be navigated to from the left nav.

I wish I could do it all programmatically ... discover if the page exists; if it doesn't, generate the replacement on the fly and fool the server into serving it. That is beyond my present talents.

kwalrath:
LGTM

kwalrath:
It's great to have a Dart glossary! There are still a few minor problems, but they shouldn't necessarily hold up this PR.

Problems I noticed in the glossary:

* 3 occurrences of "decorator" (should be "annotation" in Dart); we could just have a note that explains that, I suppose (I do try to explain differences from TS in the Dart docs)
* In the #component section, links to Directive and decorator are 404s (true in the TS version, too; I suspect it's either a redirect on the server or the glossary has lots of bad links)

naomiblack:
I think this is looking good... ship and iterate!

wardbell:
@kwalrath

* replaced the word "decorator" with "annotation" in Dart ... using my new, mad javascript-in-jade skills.

* Found and fixed many (all) of those broken links (there were others ... omission of `#` was the common cause).

kwalrath:
Thanks, @wardbell!

PR #893, opened 2016-02-27
docs(cb-ts-to-js): add cookbook about applying TypeScript examples to ES5

teropa:


teropa:
@wardbell This is ready for review.

My biggest question is about presenting it: I have it under Cookbooks for JS, but I'm worried about people not being able to find it.

wardbell:
I made a bunch of mostly cosmetic changes. To your specific question, I made this cookbook available BOTH to JS and to TS readers.

It's quite good so I'm merging it in its present state. We can improve as necessary.

PR #900, opened 2016-02-29
docs: fix typos

patrickmcd:
1. *H* in *Hierarchical* is not silent, so *an* is incorrect (a higher-archical rather than an eye-archical).
2. Missing *we* between Suppose & configured.


wardbell:
Hmm. I wouldn't have bothered for the first change because [*an hierarchy* is legit](http://www.proz.com/kudoz/English/linguistics/4454507-a_vs_an_with_hierarchy_hierarchichal.html).

But thank you for the second; therefore I'll take them both ;-)

PR #905, opened 2016-03-01
First run 404 issue fixed for QuickStart

webmaxru:
There is `404 GET /app/main.js` during initial `npm start` because of this file is not yet compiled. You have to hit _Reload page_ to make the demo work. Quick and dirty workaround could be adding `"prestart": "tsc"`. **Dirty** because it causes multiple file reloads then (but fixes the initial issue).


wardbell:
Open to another solution I you have one.

PR #906, opened 2016-03-02
refactor: remove comment that lost its way

chliebel:
The removed comment obviously shouldn’t be there.

### DIFF
--- a/public/resources/js/directives/code-example.js
+++ b/public/resources/js/directives/code-example.js
@@ -28,10 +28,3 @@ angularIO.directive('codeExample', function() {
     }
   };
 });
-
-        // '<div style="position: relative">' +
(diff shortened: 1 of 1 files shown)


wardbell:
Rejecting because this is a site infrastructure file that belongs to @petebacondarwin and he undoubtedly held on to this for a reason. He can clean it up if he wants.

We appreciate the help but we'd prefer contributions that improve the documentation rather than clean up  obscure corners.

chliebel:
Hi @wardbell, alright—I see your point.

Yes, I guess @petebacondarwin should clean that up. The comment in question is a copy of the (rendered) copy-button which looks different now because of #769.

-- 
Cheers,
Chris

PR #929, opened 2016-03-08
docs(toh5): add better explanations for templateUrl

Foxandxss:
Since this tutorial is meant to explain each step, I think we should go the whole nine yards with it.

Paths are always complicated when first encountered and with the previous explanation it wasn't clear which path to use.

Also added a few blockquotes with explanations.

Perhaps the paths changes with the offline compiler, but you know what they say about beta: It worked. That day. On my computer.


wardbell:
Excellent suggestion. I've merged your change (w/ slight rewording for what I think will be a bit more clarity). Thanks.

PR #947, opened 2016-03-10
Add ngModelOptions debounce to API search input

dfsq:
Searching API with a little debounce (400ms) during typing makes for a slightly better UX by reducing flickering on each key stroke.


wardbell:
TBH,  I didn't notice any improvement with this change. It continued to search for every key stroke (if it hadn't seen that combo before).

For example, I typed "xxy" as fast as I could type and still there were three searches.

I'm disinclined to accept this until you can demonstrate actual improvement. That's why I'm closing. Sorry.

Happy to re-open if you can provide supporting evidence.

dfsq:
Hm, this strange that you couldn't see the difference, it is very obvious: for example if you type "comp" on the live site (https://angular.io/docs/ts/latest/api/) you can see flickering footer something like this:

Without debounce

Adding debounce will make flicking less frequent:

With debounce

wardbell:
Ok ... I'll look again

PR #954, opened 2016-03-11
style(dart-template-syntax): Remove redundant and unused misplaced assets under lib/

thso:


thso:
Only assets under web/ are served.

kwalrath:
LGTM

PR #956, opened 2016-03-12
docs(quickstart): make a few corrections

Foxandxss:
I am removing some dots at the end of some sentences. I don't really like that, but that is the document convention so I remove them for consistency.


wardbell:
I like the proposed changes that are actually in the changed files but I don't see how they relate to your PR comment:

>I am removing some dots at the end of some sentences.

Foxandxss:
So when creating a list:

* foo
* bar
* baz

The convention on that file is to not put a period after each sentence.

So we had something like:

* foo
* bar.
* baz.

So for consistency I deleted the "dots" after each list item in two cases.

PR #958, opened 2016-03-13
docs: add quickstart repository where needed

Foxandxss:
We created a repository at https://github.com/IdeaBlade/angular2-quickstart so instead of forcing the people to re-create the whole quickstart every time they follow a new tutorial, they can clone it and move forward.

I updated all the articles I saw that where using the quickstart as... quickstart.


wardbell:
Updated it as discussed.Merging

PR #960, opened 2016-03-14
docs(cb-rating-component): cookbook about creating a rating component

Foxandxss:
This is a cookbook about component creation that focus a bit on using the complete syntax of ngFor.


Foxandxss:
Updated for RC 1

Issue #948, opened 2016-03-10
Router Lifecycle Hooks: routerCanActivate Vs. @CanActivate


SamVerschueren:
I was looking into router lifecycle hooks and bumped upon the following unclarity.

The [Router Lifecycle Hooks](https://angular.io/docs/ts/latest/guide/router.html#!#router-lifecycle-hooks) section tells us the following

> The router calls similar hook methods, [routerCanActivate](https://angular.io/docs/ts/latest/api/router/CanActivate-var.html) and [routerCanDeactivate](https://angular.io/docs/ts/latest/api/router/CanDeactivate-interface.html), before it navigates to a component or away from a component.

With an example of a component implementing the `CanDeactivate` lifecycle hook. To me, this looks like there is a `CanActivate` lifecycle hook as well. So I went looking into the documentation and I found the [CanActivate](https://angular.io/docs/ts/latest/api/router/CanActivate-var.html) decorator. So I was wondering `Are the API docs outdated or is the Router Lifecycle Hooks paragraph 'incorrect'?`. I looked into the source code and found the `CanActivate` decorator.

It might be nice if this could be clarified more deeply in the `Router Lifecycle Hooks` section. Make it clear that `routerCanDeactivate` is really a hook into the component and `CanActivate` is a decorator accepting a function to decide if the component should be created at all.

wardbell:
You are correct that there is no `routerCanActivate` hook method. My bad. Fixing it.

This chapter is about COMPONENT lifecycle hooks. The reference to the ROUTER hooks is parenthetical.  We should talk about those in the Router chapter (that's another Todo).

wardbell:
Actually, I can't completely fix it yet because the two ways to intervene in the activation process are so broken. I'm hoping to hold out until `OnActivate` is fixed.

But I can correct my mistake ... which I'm doing in PR #973

wardbell:
Closing because this is all about the deprecated router. We're not fixing anything about the deprecated router.

PR #973, opened 2016-03-16
docs(router): correct description of router hooks

wardbell:


wardbell:
Partially addresses #948

PR #974, opened 2016-03-16
docs(pipes) clarify impure pipes; explain missing filter/orderBy pipes

### LINKED ISSUE #570: explain why angular 2 doesn't have filter or orderby pipes
This comes up in feedback often as a request. Would be good to explain it.

wardbell:
Copied from  #697.
Closes #570 

Todos:

- [x] Why A2 ships without a  `FilterPipe` and `OrderByPipe` (equivalents in A1)
- [x] Memoization in an impure pipe improves perf (that's what the crazy JSON example does)
- [x] Trim some of the rhetorical excess (says the man with rhetorical excess)
- [x] Clarify implications of Misko's minification concerns for pipe developers. How do they avoid trouble?

Regarding the first Todo, Misko writes:
>I think in NG2 we are saying that you should do more of your filtering in controller rather then in the template. The issues with `filter` and `orderBy` is that they rely on reflection, and as such will break with minification. So we have been making sure that all of our core directive/pipes are reflection free and hence minification safe.

I hadn't considered those angles.

Here is a community implementation for those who don't care: http://embed.plnkr.co/DHLVc0

Joe responded:
>I believe the json example isn't actually memoized. it just executes once and then never executes again. memoization would involve a cache of inputs and corresponding outputs that shortcircuit the algorithm, but it doesn't do that.


wardbell:
Dear @googlebot. I authored these requests.  The author in question, Joe Eames, is also covered by the IdeaBlade corporate CLA.

PR #977, opened 2016-03-17
docs: Converted divs to labels in form

thelgevold:
@wardbell converted the form headings to use labels for consistency based on suggestion from @marcysutton   

Let me know if this is correct.


marcysutton:
Yay! :tada:

public/docs/_examples/cb-dynamic-form/ts/app/dynamic-form-question.component.html
```diff
@@ -1,13 +1,13 @@
 <!-- #docregion -->
 <div [ngFormModel]="form">
-  <div class="formHeading">{{question.label}}</div>
+  <label attr.for={{question.key}}>{{question.label}}</label>
```
Foxandxss: Couldn't this be done like:

```
<label [attr.foo]="question.key"></label>
```

I think that way is more angularish2.

On the other hand, you are missing quotes in there, right?

thelgevold:
@Foxandxss Good catch on the missing quotes. I have updated. Thanks!
About the []. I thought that we only need the []'s if we want change tracking on the property? In this case I wanted it to be a more like "one-time" bind since it's a unique id that will never change. It just needs to be set once. Am I wrong here?

PascalPrecht:
Yes, `[attr.foo]="expr"` is the same as `attr.foo="{{expr}}"`. Interpolations are also just a shorthand syntax for property binding.

gkalpak:
> Interpolations are also just a shorthand syntax for property binding.

Except only for string values, right ?

wardbell:
Yes @gkalpak. And you SHOULD use the no-brackets form if the value never changes during the lifetime of the component/directive.

If that's the case here (and I think it is), I'd prefer:
```
<label attr.for="{{question.key}}">{{question.label}}</label>
...
<select id="{{question.key}}" *ngSwitchWhen="'dropdown'" 
```

Foxandxss:
If I am not mistaken (and that is how we have it documented) there is no difference on:

```
[foo]="bar"
foo="{{bar}}"
```

And technically Angular compiler will convert the second line into the first one.

We should make a convention on what to use for the entire documentation.

wardbell:
I believe you ARE mistaken. The critical difference is that Angular will continue to watch `bar` for future changes in the first case. It will NOT do so in the second; that will be a first-time binding only.

Also, you don't need the ("). This is fine: `foo={{bar}}`

Foxandxss:
Then we need to fix our doc.

wardbell:
Yes. I was just looking at Template Syntax. I _thought_ I updated it for this case but evidently not. Put it on my todo list ;-)

Foxandxss:
I am playing here: http://plnkr.co/edit/XwONaKDlpSCeqBR9AJDb?p=preview

In both cases, it gets updated, also without quotes it fails. Am I doing something wrong?

wardbell:
AAARGH!  I was wrong.  The use of interpolation at all means that A2 continues to update the value ... something I just tested.  So the special first-time-only case works only if set with an object literal.

In the following example, `<selector>` is a component with its own `id` input property. The `attr.for` setting is for demo purposes to prove a point; you'd actually just write `for="foo"`.
```
<label attr.for="foo">{{question.label}}</label>
...
<selector id="foo" ...
```

thelgevold:
Hm.. That caught my by surprise. I thought the no []'s version would not do any change tracking and behave like the :: one time in Angular 1.
How would you do one-time in A2 now? To me the no [] seems like the perfect syntax for one-time :-)

wardbell:
I agree. Now we have to be careful again. At least the change detection cycle doesn't spin over and over as in A1.  To be honest, first-time binding never made that much of a difference in any test known to me. It made us feel good though :-)

wardbell:
Please fix to consistent syntax. I personally like this:
```
<label attr.for={{question.key}}>{{question.label}}</label>
...
<select id={{question.key}} *ngSwitchWhen="'dropdown'" 
```
But the following is acceptable too:
```
<label [attr.for]="question.key">{{question.label}}</label>
...
<select [id]="question.key" *ngSwitchWhen="'dropdown'" 
```
But, please, not a mix of property binding and interpolation

Unrelated: how is the new image different from the first? I couldn't see it ;-)

thelgevold:
I just pushed an update to switch to use [] for all of them. 

Did someone at some point say that one of the reasons for the [] syntax is that it makes it easier to make tooling around the bindings? If that's the case then maybe we should opt to use [] everywhere? Given that there is no behavioral difference between the different formats.

The image does not look very different, but I figured it made sense to just regenerate it since I switched to label tags :-)

wardbell:
I haven't heard that `[]` is easier to tool than `{{}}`. I don't understand why that would be so. Handlebars/mustaches are tooled by Visual Studio just fine.

gkalpak:
> first-time binding never made that much of a difference in any test known to me

Faulty tests ?

PR #980, opened 2016-03-18
Update testing chapter

filipesilva:
This PR updates the testing chapter by adding real code examples (instead of inline code samples) and using the newest beta version.

The example files were taken https://github.com/wardbell/ng2BabyUnitTest, but the tests that were not in the guide yet made extensive use of old APIs. Those files were renamed to have `.not-yet` as a file type, to indicate that they are not yet used.

/cc @wardbell


wardbell:
I'm taking this NOW even if it isn't done ... because TODAY it is better than the existing published chapter.

Thank you for this great start!

PR #996, opened 2016-03-21
docs(template-syntax): clarify context discussions

kwalrath:
This PR is based on text provided by @montyr75.

Unfortunately, the diffs are hard to read, so I'm attaching pics of the rendered text.

TS expression context:
image

Dart expression context:
image

TS statement context:
image

Dart statement context:
image

@montyr75
@wardbell

### DIFF
--- a/public/docs/ts/latest/guide/template-syntax.jade
+++ b/public/docs/ts/latest/guide/template-syntax.jade
@@ -6,9 +6,7 @@ include ../_util-fns
 
   Many of us are familiar with the component/template duality from our experience with model-view-controller (MVC) or model-view-viewmodel (MVVM). In Angular,  the component plays the part of the controller/viewmodel, and the template represents the view.
 
-  Let’s find out what it takes to write a template for our view. We’ll cover these basic elements of template syntax.
-
-  # Table Of Contents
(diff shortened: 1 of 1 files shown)


montyr75:
I like it! This version will be much clearer for Dart devs.

wardbell:
LGTM

zoechi:
What about explicitly mention how to work around? Like a getter or function in the component that forwards to a global member?.

```dart
@Component(
  selector: 'my-component',
  template: '<button (click)="log('been clicked')">click me</button>')
class MyComponent {
  log(String value) => print(value);
}
```
Not sure if this is obvious for everyone.

montyr75:
It's subtle, but I did imply that in the very last sentence: "They can't _directly_ call `print`..."

But yes, if you wanted to be an overachiever, you could explicitly mention workarounds.

zoechi:
:D You don't want to be an overachiver? 
I just saw a few questions on SO and Gitter where people were wondering what's a valid way to do it. My example above seems to come to mind easily for them as well, but they keep wondering if this is the way to go or if they missed something. With an explicit pointer these doubts would vanish.

montyr75:
I like to overachieve, yes, but this is Kathy's call.  :)

I agree that it would be helpful.

PR #999, opened 2016-03-23
Align initial dropdown styles with .md-button styles

dfsq:
Attempt to add several rules from .md-button to the existing button styles so that they are rendered the same and no visual flickering happens.


Foxandxss:
What is the status of this PR?

juleskremer:
We are redesigning angular.io and the documentation site now; this will be addressed with that redesign so I'm closing this PR now.

PR #1002, opened 2016-03-24
(WIP) chore(toh): change inputs array for input decorator

Foxandxss:
The PR itself is complete, but there is a weird issue with the generated plunker:

http://plnkr.co/edit/fbNckRhMEitZ72XWFsaW?p=preview

That doesn't happen if we do `npm start` at the `toh-3` folder or if I replicate this on a webpack starter I have. Only plunker.

I have no idea why so I leave this here so we can debug it and not forget about this.

### DIFF
--- a/public/docs/_examples/cb-component-communication/ts/app/hero.ts
+++ b/public/docs/_examples/cb-component-communication/ts/app/hero.ts
@@ -1,4 +1,4 @@
-export interface Hero {
+export class Hero {
   name: string;
 }
 
--- a/public/docs/_examples/toh-1/ts/app/app.component.ts
+++ b/public/docs/_examples/toh-1/ts/app/app.component.ts
@@ -1,12 +1,12 @@
 // #docregion pt1
 import {Component} from 'angular2/core';
 
-// #docregion hero-interface-1
-interface Hero {
+// #docregion hero-class-1
+export class Hero {
   id: number;
   name: string;
 }
-// #enddocregion hero-interface-1
+// #enddocregion hero-class-1
 
 @Component({
   selector: 'my-app',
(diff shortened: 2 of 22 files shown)


ericmartinezr:
Maybe this [question](http://stackoverflow.com/questions/36114770/plnkr-getting-error-as-unexpected-token-export-while-export-a-typescript) has some clues? Changing `interface` to `class` removes the error message.

Foxandxss:
Interesting. We discussed it yesterday but none was really sure. Adding:

```
meta: { 'app/hero.ts': { format: 'es6' }},
```

Fixes it as seen [here](http://plnkr.co/edit/f6snQFeA5amPorU3coGM?p=preview)

Locally we configure System.js differently. I am going to research a bit to see if we can make a better `index.html` for plunker or just modify the demo for the affected ones.

wardbell:
Making a few adjustments:

1. I replaced the `Hero` interface with a class ... everywhere ... and adjusted the text accordingly. I just don't want to have to deal with this ... even though I cannot repro your plunker problem

2. You added a new snippets file. No to snippets files!  They are the last choice of a desperate author. They are not testable.  I can eliminate the need for this file.

Foxandxss:
To reproduce the problem, just open the plunker (on the first message) and look at the console:



To fix it, we just need to add a new `meta` tag to the System.js config. Problem is that is generated automatically for all files with the same content.

1) I liked our explanation about using interfaces because they add no code and we just needed it for typings and development help.

2) That is fine by me, I can remove them :)

wardbell:
Weird. I see it now. Whatever. Problem solved by making it a class. Thanks. Committing and pushing now.

wardbell:
Ooops. You missed the others in toh-4 ... tutorial. Getting them now.

Foxandxss:
Oh I see your changes. I was about to change them myself.

Did I miss the toh-4 ones? Weird, I was sure I changed 3, 4 and 5 (at least add the input, there wasn't any relevant conversation around it).

Thank you for the changes.

wardbell:
I think we're good to go. I'm merging.

PR #1004, opened 2016-03-24
docs(template-syntax): clarify [()] de-sugaring

wardbell:
also changed anchors to use id instead of name

### DIFF
--- a/public/docs/ts/latest/guide/template-syntax.jade
+++ b/public/docs/ts/latest/guide/template-syntax.jade
@@ -542,7 +542,7 @@ table
 // #docregion property-binding-7
 :marked
   ### Binding target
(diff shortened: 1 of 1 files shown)


public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -542,7 +542,7 @@ table
 // #docregion property-binding-7
 :marked
   ### Binding target
-  A name between enclosing square brackets identifies the target property. The target property in the following code is the image element’s `src` property.
+  a id between enclosing square brackets identifies the target property. The target property in the following code is the image element’s `src` property.
```
Foxandxss: an

PR #1005, opened 2016-03-25
Updated link on promises

EmmaRamirez:
The old link on the documentation lead to an article on promises that
contained outdated information, so this one leads to newer content.


EmmaRamirez:
I signed it!

wardbell:
Thanks. I added similar updates to all similar links.

PR #1007, opened 2016-03-26
docs(dart): update to b12

kwalrath:
Also update dart_to_js_script_rewriter dependency to ^1.0.1, and change most angular2.dart imports to be core.dart instead.

The pipes example broke without the angular2.dart import, so I let it be.

The server-communication sample has never worked for me, so I changed it but might have broken it further.


kwalrath:
Looks lovely. Merging!

PR #1010, opened 2016-03-26
chore: change RouteConfig for Routes

Foxandxss:
To be merged when this feature arrives.

The Dart examples / .jade are not translated. I am not sure if I have to touch it myself or leave that for someone else. I can update the PR to add those changes too.

The API / Cheatsheet I believe they are generated from the angular repo so there is nothing I can do for that and I guess it will be magically fixed when the feature arrives.


kwalrath:
@Foxandxss I'll make sure the Dart version is updated when this goes in.

The cheatsheet is generated from files under https://github.com/angular/angular/tree/master/modules/angular2/docs/cheatsheet. It looks like most of those haven't been updated for months, so don't count on it happening magically!

Foxandxss:
Sounds good to me @kwalrath. If you want me to update this PR with the Dart, I will do it right away.

Also, if you think that those docs needs to be updated, ping me as well and I update them as well, no problem.

wardbell:
Holding until A2 has this change ... which it doesn't yet.

wardbell:
I don't believe we're going with this. Switching to new router. Closing

PR #1011, opened 2016-03-28
docs(toh-pt4): Fix order of methods

JamesLaverack:
Change the ordering of the methods `getHeroes` and `getHeroesSlowly` in the "Take it slow" appendix.
The original ordering incorrectly suggested that the reader should swap the new `getHeroesSlowly`
method for the existing `getHeroes` method, rather than the other way around.

### DIFF
--- a/public/docs/ts/latest/tutorial/toh-pt4.jade
+++ b/public/docs/ts/latest/tutorial/toh-pt4.jade
@@ -384,7 +384,7 @@ code-example(format="." language="html").
   Like `getHeroes`, it also returns a promise. 
   But this promise waits 2 seconds before resolving the promise with mock heroes.
   
-  Back in the `AppComponent`, swap `_heroService.getHeroesSlowly` for `_heroService.getHeroes`
+  Back in the `AppComponent`, swap `_heroService.getHeroes` for `_heroService.getHeroesSlowly`
   and see how the app behaves.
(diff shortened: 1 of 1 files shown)


wardbell:
Why do you say "incorrectly"?  The text seems correct to me. We're asking the reader to use the "slowly" version instead of the regular version. That's what it means to

> swap `_heroService.getHeroesSlowly` for `_heroService.getHeroes` and see how the app behaves. 

The change that you are suggesting would do the opposite ... which makes no sense to me. 

Help me see what I am missing.

JamesLaverack:
I would read the phrase "...swap X for Y..." to mean "replace instances of X with instances of Y". I think on further thought that it's the use of the word "swap" that causes the confusion.

How would you feel about replacing the sentence with:

> Back in the AppComponent, replace `_heroService.getHeroes` with `_heroService.getHeroesSlowly` and see how the app behaves.

This should be obvious from context to an attentive reader, but especially from a tutorial ambiguity of any kind should be avoided.

Foxandxss:
For what is worth, I agree with James in here. Also I prefer what he said in the last message. Better to add a `with` and get rid of all confusions.

wardbell:
I like it. That I can do. I'll make that change as I merge yours (so you get credit ;-))

PR #1012, opened 2016-03-28
[WIP] docs: add DI cookbook

thelgevold:
@wardbell Adding the DI cookbook. Let me know if this is what you had in mind. 

Adding @PascalPrecht as well in case he has time to offer some feedback.


Foxandxss:
We are promoting `@Injectable()` in every service (even if it has no dependency) as the best practice.

thelgevold:
Thanks for pointing that out @Foxandxss - I will update if that's the direction we are going in. 

That said, I am not sure I love that requirement. Mainly because it adds a dependency on Angular 2 where strictly isn't necessary.

thelgevold:
Based on team feedback I have added Injectable to all services for consistency

wardbell:
Hi Tor -
I'm documenting here some of the more important changes that I'm making/recommending.

1. We recommend registering application-wide providers in the root component, not in bootstrapping.  Anything that really belongs with the app itself goes in `AppComponent`.  I'm moving your app-specific providers there. Illustrated usage by extending `HeroBiosComponent` with logger service injection.

2. We only register Angular overrides in the bootstrapper.  A good example is router `LocationStrategy` or a dev-time http backend fake like `a2-in-memory-web-api`. 

3. All component class names should end in `Component`. Fixed that.

4. We recommend that constructors be super simple and never call services that could do anything like hit a database. I revised the `SortedHeroesComponent` by moving the `HeroesService` call to `ngOnInit`.

5. I wanted to make a more dramatic argument against component inheritance so I made `HeroesBase` into a component. I also shifted the sorting logic to `SortedHeroesComponent` for ... ahem ... greater realism.

6. I think the inheritance issues are lower priority than anything else that you're showing. I shifted it to the bottom.

I'll update this comment as I make more changes.

wardbell:
Hey Tor - I've made more changes but it's in a broken state. Please don't touch at the moment.

wardbell:
AARGH.  I renamed the branch to all lowercase (our standard), and it closed this PR.

Please see continuation PR #1022

PR #1014, opened 2016-03-29
use local resource to replace google cdn

eeandrew:
Googleapis is not available in China. So we can't get these lib files directly from googleapis. Instead, we have to use local lib files. 

I create this PR so that  angular.min.js angular-animate.min.js angular-aria.min.js angular-material.min.js angular-material.min.css material-icons.min.css are loaded from local server.

Please take these into consideration. Thx.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -306,16 +306,16 @@ gulp.task('remove-example-boilerplate', function() {
     })
 });
 
-gulp.task('serve-and-sync', ['build-docs'], function (cb) {
+gulp.task('serve-and-sync', ['build-docs','_copy_angular_resource'], function (cb) {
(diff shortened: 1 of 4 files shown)


naomiblack:
Thanks -- this is a good change -- it also makes it easier to work offline. @eeandrew Could you sign the Google CLA please? This is essential in order for us to accept your PR.

Separately, I'm working on a project to localize the docs to Chinese, which would involve hosting all the content on a more China-friendly server. Would you be interested in contributing to that effort? :)

Foxandxss:
What is the status of this PR? Ping @eeandrew

naomiblack:
Please see angular.cn -- I think this addresses your concern as it's hosted in China.

PR #1018, opened 2016-03-29
Add info file for dart-doc-syncer to template-syntax.

thso:
Others will follow, but let's test on a single repository first.

ptal @kwalrath


kwalrath:
Merging now...

PR #1036, opened 2016-04-01
docs(cb-select-box-component): cookbook about creating a select box

Foxandxss:
This is quite a big cookbook, but we wanted to cover a lot here.

For the e2e tests, there is no way that I can find to assert what's the selected option in a select box. So I have an array of heroes and since I can get the selected value, I can use the array to get the value.


Foxandxss:
Updated to RC1

PR #1037, opened 2016-04-01
Update package.json

FintanK:
Keywords for package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,9 +9,17 @@
     "harp": "harp",
     "live-server": "live-server",
     "test-api-builder": "jasmine-node tools/api-builder",
-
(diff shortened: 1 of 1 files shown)


FintanK:
I signed it!

Foxandxss:
We are not releasing this repo as a package so I don't think that keywords are needed.

FintanK:
Not needed as such but metadata is never a bad idea in my opinion.

wardbell:
Thanks for your effort and kind intent. Sorry to disagree though. 

A lack of keywords is not "bad metadata" or they wouldn't be optional.

There's nothing wrong with this contribution per se, so we may merge it.

PR #1040, opened 2016-04-02
docs(style-guide): add A2 style-guide - v.1

wardbell:


wardbell:
John and I are working on this together. He's ok with this commit.

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,672 @@
...
+
+  ### Avoid Naming Collisions
+  <a id="020"></a>
+  #### Style 020
+
+    - Use unique naming conventions.
+
+    **Why?** Unique names help avoid module name collisions. We import and export modules by the file in which they are contained. Thus the name of the files and their folders is important.
+
+    **Why?** Names of folders and files should clear convey their intent. For example, `app/speakers/speaker-list.component.ts` may contain a component that manages a list of speakers. The folder and file names clearly convey the meaning, and thus the module import is clear as `import { SpeakerComponent } from './speakers/speaker-list.component'`.
```
IgorMinar: clear -> clearly

IgorMinar: shouldn't this be updated with the `index.ts` info which change the import to `from './speakers'`?

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,672 @@
...
+  **Why?** Small functions are easier to maintain.
+
+  **Why?** Small functions help avoid hidden bugs that come with large functions that share variables with external scope, create unwanted closures, or unwanted coupling with dependencies.
+
+<a href="#toc">Back to top</a>
+
+:marked
+  ## Modules
+
+  ### Avoid Naming Collisions
```
IgorMinar: I'm confused about how are you avoiding naming collisions. ES module system is designed to prevent name collisions.

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,672 @@
...
+  <a id="020"></a>
+  #### Style 020
+
+    - Use unique naming conventions.
```
IgorMinar: the conventions themselves should be unique? unique for each project or each file? that sounds a bit counter-productive to our goal. :)

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,672 @@
...
+<a href="#toc">Back to top</a>
+
+:marked
+  ## Components
+
+  ### Member Sequence
+  <a id="033"></a>
+  #### Style 033
+
+    - Place properties up top, then public functions, then private functions, alphabetized, and not spread through the component code.
```
IgorMinar: public functions -> public methods
private functions -> private methods

what does "not spread through the component code" mean?

IgorMinar: also keep in mind that ES6 doesn't have property initializers yet. So unless you are using some custom babel plugins, you'll have to initialize the properties in the constructor.

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,672 @@
...
+    /* recommended */
+    export class ToastComponent implements OnInit {
+      // private fields
+      private _defaults = {
+        title: '',
+        message: 'May the Force be with You'
+      };
+      private _toastElement: any;
+
+      // public properties
```
IgorMinar: is it intentional that public properties are below private properties? with methods the order is reversed.

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,672 @@
...
+<a href="#toc">Back to top</a>
+
+:marked
+  ### Defer Logic to Services
+  <a id="035"></a>
+  #### Style 035
+
+    - Defer logic in a component by delegating to services.
+
+    - Define a component for a view, and try not to reuse the component for other views. Instead, move reusable logic to factories and keep the component simple and focused on its view.
```
IgorMinar: this sounds weird to me. In my mind there are two kinds of components:

1/ routable components that are usually very application specific and are not meant to be reused even within the application.
2/ reusable components (application agnostic ones as buttons, or the whole material suite, and application-specific ones like profile/product widgets with pop-overs, etc).

the purpose of the service is to facilitate data delivary to these components

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,672 @@
...
+
+      **Why?** Logic may be reused by multiple components when placed within a service and exposed via a function.
```
IgorMinar: functions -> methods?

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,672 @@
...
+<a href="#toc">Back to top</a>
+
+:marked
+  ## Services
+
+  ### Singletons
+  <a id="040"></a>
+  #### Style 040
+
+    - Services are singletons and should be used for sharing data and functionality.
```
IgorMinar: well, not quite. services are singletons within the same injector - you can think about each element in the component composition tree as having it's own injector and each element inherits injector services from the parent.

but if a child element defines a provider for a service that was already defined an instantiated in the parent, the DI will create a new instance of this service visible only to this child component and its children. This is what allows you to have multiple instance of certain services (e.g. ElementRef is different for each component) and we don't need the concept of "locals" from Angular 1.

johnpapa: yep. just hadnt figured out how to articulate this, and what is applicable to the guide, if at all. because that depth belongs in the docs, so some of this may just go

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,672 @@
...
+  ```
+  
+<a href="#toc">Back to top</a>
+
+:marked
+  ### Providing a Service
+  <a id="041"></a>
+  #### Style 041
+
+    - Services should be provided to the Angular 2 injector at the top-most component where they will be shared.
```
IgorMinar: Maybe we should even before this section introduce the concept of component hierarchies as a separate recommendation. Everything in an Angular app is based on this single concept (the component composition, routing, di, directory structure, lazy-loading, compilation units....).

The general guidance for creating reusable components or defining providers should be that they should be created at the level of deepest common parent. This sets the visibility of that component or service instance within the application tree.

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,672 @@
...
+  <a id="050"></a>
+  #### Style 050
+
+    - Refactor logic for making data operations and interacting with data to a service. Make data services responsible for XHR calls, local storage, stashing in memory, or any other data operations.
+
+      **Why?** The component's responsibility is for the presentation and gathering of information for the view. It should not care how it gets the data, just that it knows who to ask for it. Separating the data services moves the logic on how to get it to the data service, and lets the component be simpler and more focused on the view.
+
+      **Why?** This makes it easier to test (mock or real) the data calls when testing a component that uses a data service.
+
+      **Why?** Data service implementation may have very specific code to handle the data repository. This may include headers, how to talk to the data, or other services such as `$http`. Separating the logic into a data service encapsulates this logic in a single place hiding the implementation from the outside consumers (perhaps a component), also making it easier to change the implementation.
```
IgorMinar: $http is no more :)

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,672 @@
...
+      
+      **Why?** A component that handles routing is known as the componenter router.
+
+      **Why?** The `<router-outlet>` indicates where the tempalte should be displayed for the target route.
+      
+      ```typescript
+      import { Component } from 'angular2/core';
+      import { RouteConfig, ROUTER_DIRECTIVES, ROUTER_PROVIDERS } from 'angular2/router';
+
+      import { SpeakersComponent, SpeakerService } from './^speakers';
```
IgorMinar: `^` -> `+`

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,672 @@
...
+      import { DashboardComponent } from './^dashboard';
```
IgorMinar: `^` -> `+`

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,672 @@
...
+    - Structure your app such that you can `L`ocate your code quickly, `I`dentify the code at a glance, keep the `F`lattest structure you can, and `T`ry to stay DRY. The structure should follow these 4 basic guidelines.
+
+      *Why LIFT?*: Provides a consistent structure that scales well, is modular, and makes it easier to increase developer efficiency by finding code quickly. Another way to check your app structure is to ask yourself: How quickly can you open and work in all of the related files for a feature?
+
+      When I find my structure is not feeling comfortable, I go back and revisit these LIFT guidelines
+
+      1. `L`ocating our code is easy
+      2. `I`dentify code at a glance
+      3. `F`lat structure as long as we can
+      4. `T`ry to stay DRY (Don’t Repeat Yourself) or T-DRY
```
IgorMinar: why are you repeating yourself? this is the same as line 548. no?

public/docs/ts/latest/guide/style-guide.jade
johnpapa: this is gone

public/docs/ts/latest/guide/style-guide.jade
johnpapa: this is removed

public/docs/ts/latest/guide/style-guide.jade
johnpapa: yes

public/docs/ts/latest/guide/style-guide.jade
johnpapa: revised. also fixed the ctor initializer.

for now just using TS

public/docs/ts/latest/guide/style-guide.jade
johnpapa: mistake. fixed ordering

public/docs/ts/latest/guide/style-guide.jade
johnpapa: revised

public/docs/ts/latest/guide/style-guide.jade
johnpapa: agreed

public/docs/ts/latest/guide/style-guide.jade
johnpapa: good catch :)

public/docs/ts/latest/guide/style-guide.jade
johnpapa: revised

public/docs/ts/latest/guide/style-guide.jade
johnpapa: revised

public/docs/ts/latest/guide/style-guide.jade
johnpapa: gone

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,651 @@
...
+
+  Define 1 component per file, recommended to be less than 400 lines of code.
+
+  **Why?** One component per file promotes easier unit testing and mocking.
+
+  **Why?** One component per file makes it far easier to read, maintain, and avoid collisions with teams in source control.
+
+  **Why?** One component per file avoids hidden bugs that often arise when combining components in a file where they may share variables, create unwanted closures, or unwanted coupling with dependencies.
+  
+  The following example defines the AppComponent, handles the bootstrapping, and shared functions all in the same file.
```
mgechev: "AppComponent" could be formatted as a keyword `AppComponent`.

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,651 @@
...
+    - Place properties up top followed by methods. Private members should follow public members, alphabetized.
+
+      **Why?** Placing members in a consistent sequence makes it easy to read and helps you instantly identify which members of the component serve which purpose.
+
+    ```typescript
+    /* recommended */
+    export class ToastComponent implements OnInit {
+      // public properties
+      title: string;
+      message: string;
```
mgechev: `title` should be after `message` (`title` > `message` in lexicographical order, not alphabetized).

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,651 @@
...
+        templateUrl: 'app/app.component.html',
+        styleUrls: ['app/app.component.css'],
+        directives: [ROUTER_DIRECTIVES, NavComponent],
+        providers: [
+          ROUTER_PROVIDERS,
+          SpeakerService
+        ]
+      })
+      @RouteConfig([
+        { path: '/dashboard', name: 'Dashboard', component: DashboardComponent, useAsDefault: true },
```
mgechev: I think it'll be good to define naming conventions for the routes. I like the current one: `componentName.replace(/Component$/, '')`;

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,651 @@
...
+            /exception
+            /logger
+          /core
+          /dashboard
+          /data
+          /layout
+          /widgets
+        /content
+        index.html
+      .package.json
```
mgechev: There's extra "." before `package.json`.

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -0,0 +1,651 @@
...
+      import { DashboardComponent } from './+dashboard';
+      import { NavComponent } from './layout/nav.component';
+
+      @Component({
+        selector: 'my-app',
+        templateUrl: 'app/app.component.html',
+        styleUrls: ['app/app.component.css'],
+        directives: [ROUTER_DIRECTIVES, NavComponent],
+        providers: [
+          ROUTER_PROVIDERS,
```
zoechi: Adding `ROUTER_PROVIDERS` on a component instead of to `boostrap(AppComponent, [ROUTER_PROVIDERS])` seems to cause issues with routing. I suggested to move them to `bootstrap()` repeatedly on StackOverflow and most of the time this was what fixed the problem. Should this work the same on the root component?

zoechi:
What about some words about whether to prefer decorators or component/directive parameters (`inputs: []`, `outputs: []`, `host: {}`)?

mgechev:
@zoechi this will follow-up once we agree on the core principles (directory structure and naming).

public/docs/ts/latest/guide/style-guide.jade
johnpapa: done

public/docs/ts/latest/guide/style-guide.jade
johnpapa: done

public/docs/ts/latest/guide/style-guide.jade
johnpapa: yes. we need a style for it

johnpapa: but i heard a rumor that route names may be going away. perhaps we avoid any route styles for now?

@wardbell @IgorMinar ????

public/docs/ts/latest/guide/style-guide.jade
johnpapa: tldr; we try not to provide anything at boostrap. 

this makes it more difficult to test and often the root component is what really needs the providers, as it is the root of the component tree. this also makes the bootstrap super simple and focuses just on the platform.

i know @wardbell had some chats with Misko on this and he can chime in too.

johnpapa:
@zoechi I agree with @mgechev ... we need to get the core guidelines written first. Then we'll keep updating it.

public/docs/ts/latest/guide/style-guide.jade
zoechi: I don't have a strong opinion in one direction or the other, but I saw this caused lots of confusion. The hint in the docs to use `provide: []` on the root component seemed to be interpreted a lot to add it on every component which breaks many things.
I just think the favored approach should be stated **very clearly**.

wardbell:
Merging ... THIS BRANCH IS NOW CLOSED

public/docs/_examples/style-guide/ts/001/app/placeholder.ts
```diff
@@ -0,0 +1,93 @@
...
+    template: `
+        <pre>{{heroes | json}}</pre>  
+      `,
+    styleUrls: ['app/app.component.css'],
+    providers: [HeroService]
+  })
+  export class AppComponent implements OnInit{
+    heroes: Hero[] = [];
+
+    constructor(private _heroService: HeroService) {}
```
choeller: Just out of curiosity: Why do you still recommend to prefix private fields with `_`. I always saw that as an ugly hack because of the lack of the private modifier, so since using TypeScript I totally dropped that.

mgechev: Mostly for debugging, especially when you don't have sourcemaps.

Also, you can access all these properties inside of the templates. Having instance members prefixed with `_` gives better semantics.

choeller: @mgechev Thanks for the quick reply! I just added a discussion-issue about this here: https://github.com/angular/angular.io/issues/1108. As this convention goes against Microsofts recommendations I think that it is at least worth to think about dropping the prefix ;)

PR #1042, opened 2016-04-02
docs(guide): fix typo

techeverri:
Missing closing parenthesis


wardbell:
We're about to rewrite most of this. Thanks, but I'm declining as this change will be lost in the wash.

PR #1043, opened 2016-04-02
Fix some errors in Jasmine Testing 101

bjwyse:
The links to node_modules need to go back one directory as the unit-tests.html file is created in the 'src' folder, not the root folder.

When we add the 1st.spec.ts file, we add it to the existing unit-tests.html file and don't create a new 1st-tests.html file. Fixed this in the doc too.


filipesilva:
Right you are, completely missed this myself. Thank you!

wardbell:
Thanks. Merged

PR #1044, opened 2016-04-02
[WIP]:Create Tour of Heroes chapter on http

thelgevold:
*** DO NOT MERGE ***
This branch very much a Work in progress and mainly here for feedback before I do the full implementation.

@wardbell 
Started looking at the Tour of Heroes chapter on Http. I am assuming this chapter will pick up where the router chapter left off and start with the application that was created there.

I have added a rough outline with some ideas in the jade file. Let's discuss if this is headed in the right direction.

Highlevel:

"Backstory" and the motivation behind the move to Http

Show how to configure a2-in-mem-web-api. This is already explained in the Dev guide appendix, but thinking we need to show how to use it in the context of our TOH AppComponent.

Convert HeroService to use http instead of hard coded response. Do we stick to promise based http calls? Could do Observables, but have to teach observables too then. Besides the existing components already expect promises from HeroService. The code in this PR uses http with promises, but I can easily change it.

Add ability to edit and add new heroes using a2-in-mem-web-api. Thinking we can reuse HeroDetailComponent to support edit and add scenarios. This includes adding a post method to HeroService. Did a quick tests and a2-in-mem-web-api seems to handle post requests without any issues.  

Are there other topics/features we should include as well?


wardbell:
I like where you're going.

Remember that the tour is a gentle introduction to A2. It's meant to reassure the reader that A2 covers the application ground that _they need it to cover_.  It's not supposed to be comprehensive. It's not supposed to be contentious or risk confusing the reader with choices.

Let's keep the http chapter in that spirit.

The chapter should also be short. Not too short but I'd definitely err on the side of brief and breezy.

---

The promise/observable trade-off is the most vexing to me. With my preamble in mind, I think we stick with promises *at the boundary* between the `HeroService` and the components. *Within* the `HeroService` we talk observables and _hint_ at what's possible there. But I don't want an essay on observables in the ToH.

This is not the place to lead readers on a scary adventure into the uncertainties of observables and other competing data strategies.

There is no reason for ToH to get into the observable weeds. We should say just enough to let the reader know that ***we know observables are there***.

---
Yes the A2-in-mem-web-api is the way to go. It handles all the CUD methods you need. No reason to go too deep thought. Remember that it exists for demos. We want them to know that we're using it to make their learning experience easier (they won't need to set up a server to learn A2's http).  But we don't want to go overboard. They should know that using a real web api involves no more than a change of  URL.

---

Please make sure that we give a nod to error handling. **We must write error catching**. We don't have to do anything grand with the errors: writing to console is probably sufficient. But we *absolutely must* have catch in the `HeroService` for all methods.

johnpapa:
*`a2-in-memory-web-api`*
I agree using this is the way to go. However ToH should just say it's here to help us avoid a backend for now. If we had a real back end, we would drop that code. If you want to learn more, check out this *future* appendix about how to provide a replacement for the backend. I would go any deeper than that. Otherwise, this is very scary at this stage.

*promises vs observables*
There are a lot of ways we can go here. I like keeping the concept count down. So let's think about where we are as we enter this chapter. We know how to use a basic promise to get mock data, but we dont yet know how to get data with *http* or an observable. I think it makes a ton of sense to introduce one of these in this chapter, but not both. Which does the user want more? I bet they want *http* as many have likely rarely heard of Rx just yet. So I think adding *http* and showing how to use *toPromise* is the way to go in this chapter.

Also add a quick note/callout pointing out that we are using toPromise because natively the `http.get` call returns something else. That something is an observable. But don;t worry, we'll cover that in the next chapter.

Then, the next chapter can show how to take http and turn things to an observable.

Following this pattern it feeds what the reader is craving .... they walk into the chapter wanting to know http, they don't realize they may want Rx later quite yet. So they learn http, see a small mention of  Rx, and then they are primed for the next chapter which will swap out promises for Rx.

*error handling*
I vote that every sample has error handling. This is a crucial and oft missing or improperly coded aspect. We should help folks see real error handling right out of the gate.

Looking forward to seeing what you do, Tor!

Foxandxss:
I totally agree with John here regarding promises. As long as we make very clear that we will cover observables in the next part or something.

filipesilva:
John's plan to introduce Observables sounds great.

Maybe the next chapter should be about application state, and shows how to change state on Component A via service calls on the unrelated Component B. Then it would make a lot of sense to change the promises to subscribes.

thelgevold:
Thanks everyone. This is great feedback! I will keep you in the loop and let you know when this is in a more finished state.

thelgevold:
Closed this one due to branch issue. Opened new PR here: https://github.com/angular/angular.io/pull/1066

PR #1048, opened 2016-04-05
docs: (template-syntax) Fixed typo

bampakoa:
Fixed typo error in Template Syntax chapter of Basics section.

### DIFF
--- a/public/docs/ts/latest/guide/template-syntax.jade
+++ b/public/docs/ts/latest/guide/template-syntax.jade
@@ -542,7 +542,7 @@ table
 // #docregion property-binding-7
 :marked
   ### Binding target
-  a id between enclosing square brackets identifies the target property. The target property in the following code is the image element’s `src` property.
+  An element property between enclosing square brackets identifies the target property. The target property in the following code is the image element’s `src` property.
 
 // #enddocregion property-binding-7
 +makeExample('template-syntax/ts/app/app.component.html', 'property-binding-1')(format=".")


bampakoa:
I signed it!

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -542,7 +542,7 @@ table
 // #docregion property-binding-7
 :marked
   ### Binding target
-  a id between enclosing square brackets identifies the target property. The target property in the following code is the image element’s `src` property.
+  An element property between enclosing square brackets identifies the target property. The target property in the following code is the image element’s `src` property.
```
Foxandxss: I see this, but I would say `An attribute between enclosing...`.

But that is my opinion. Let see what @wardbell has to say.

bampakoa: I thought to add the **attribute** word but then it came to my mind that Angular sets DOM properties and not attributes, thus it will be more clear to the user reading this.

Foxandxss: Yeah, is a confusing point. I still think that what you see in the HTML is an attribute and then it becomes a property when it gets processed by the browser. So it is an attribute with square brackets.

Or perhaps I am wrong. Let's get another opinion in here.

wardbell:
The word "***property***" is correct. Attribute would actually be wrong ... an important point that I belabor earlier in the chapter. 

We must distinguish between the _property_ binding and attribute _position_; the former is *what it is*, the latter is *where* it is.

Evidently I have yet to firmly establish this essential point.

Thanks for the fix!

Foxandxss:
You are right, I just didn't read the context properly it seems :)

PR #1050, opened 2016-04-06
todo app remaining tasks reducer func fixed and clarified

sulmanen:
Reduce func in get remaining() looked funny in the example. Maybe this is clearer.

### DIFF
--- a/public/docs/_examples/homepage-todo/ts/app/todo_app.ts
+++ b/public/docs/_examples/homepage-todo/ts/app/todo_app.ts
@@ -23,7 +23,7 @@ export class TodoApp {
   ];
 
   get remaining() {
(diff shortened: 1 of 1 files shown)


sulmanen:
I signed it!

gkalpak:
Indeed `count + +!todo.done` was a little too much :)
Even simpler would be:

```js
this.todos.filter((todo: Todo) => !todo.done).length
```

sulmanen:
why yes, much better.

wardbell:
I agree that it reads better. The `reduce` would be more efficient if we had a gazillion todos ... an absurdity in itself. Accepting ... after taking it the last step (because of type inference):
```
return this.todos.filter(todo => !todo.done).length;
```

PR #1053, opened 2016-04-06
docs(pipes): Fixed typo

bampakoa:
Fixed typo in **pipes** section of guide.


wardbell:
Good catch on the "If", I prefer "With that concern in mind, ..." to "It is therefore understood,..."

That's how I'm going to accept your PR. Thanks!

PR #1062, opened 2016-04-08
fix(quickref-cookbook): fix broken date test

filipesilva:
The date test was failing in different timezones due to dates being localized.

This fix makes the test slightly less robust, but guarantees that it will work in every timezone irrespective of differences between the node env timezone and browser timezone.

/cc @wardbell

### DIFF
--- a/public/docs/_examples/cb-a1-a2-quick-reference/e2e-spec.js
+++ b/public/docs/_examples/cb-a1-a2-quick-reference/e2e-spec.js
@@ -7,19 +7,19 @@ describe('Angular 1 to 2 Quick Reference Tests', function () {
   it('should display no poster images after bootstrap', function () {
     testImagesAreDisplayed(false);
   });
-  
+
   it('should display proper movie data', function () {
     // We check only a few samples
     var expectedSamples = [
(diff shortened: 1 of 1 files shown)


public/docs/_examples/cb-a1-a2-quick-reference/e2e-spec.js
```diff
@@ -7,30 +7,30 @@ describe('Angular 1 to 2 Quick Reference Tests', function () {
...
   });
-  
+
   it('should display proper movie data', function () {
     // We check only a few samples
     var expectedSamples = [
       {row: 0, column: 0, element: 'img', attr: 'src', value: 'images/hero.png', contains: true},
       {row: 0, column: 2, value: 'Celeritas'},
-      {row: 1, column: 3, value: 'Dec 17, 2015'},
+      {row: 1, column: 3, value: '2015', contains: true},
```
wardbell: So it will only fail in Europe one day of the year. LOL. I can live with that.

wardbell:
Please verify that my amendment works in your timezone. Thx.

PR #1071, opened 2016-04-11
docs(dart): convert toh-3 to Dart

kwalrath:
Based on #700 by @kasperpeulen.

### DIFF
--- a/public/docs/dart/latest/tutorial/toh-pt3.jade
+++ b/public/docs/dart/latest/tutorial/toh-pt3.jade
@@ -1,14 +1,252 @@
-include ../_util-fns
+include ../../../../_includes/_util-fns
 
 :marked
-  We're working on the Dart version of this case study.
-  In the meantime, please see these resources:
+  Our app is growing.
+  Use cases are flowing in for reusing components, passing data to components, and creating more reusable assets. Let's separate the heroes list from the hero details and make the details component reusable.
 
-  * [Multiple Components](/docs/ts/latest/tutorial/toh-pt3.html):
-    The TypeScript version of this chapter
+  [Run the live example for part 3](https://tour-of-heroes.firebaseapp.com/toh3/)
 
-  * [Dart source code](https://github.com/angular/angular.io/tree/master/public/docs/_examples/toh-5/dart):
-    A preliminary Dart version of the Tour of Heroes app,
-    featuring the hero editor, a master/detail page,
-    multiple components, services, and routing
+.l-main-section
+:marked
+  ## Where We Left Off
+  Before we continue with our Tour of Heroes, let’s verify we have the following structure. If not, we’ll need to go back and follow the previous chapters.
 
+.filetree
+  .file angular2_tour_of_heroes
+  .children
+    .file lib
+    .children
+      .file app_component.dart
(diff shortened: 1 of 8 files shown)


public/docs/dart/latest/tutorial/toh-pt3.jade
```diff
@@ -1,14 +1,251 @@
...
+
+    <!-- TODO
+    .l-sub-section
+      :marked
+        Learn more about naming conventions in the chapter [Naming Conventions]
+    :marked
+    -->
+:marked
+  We begin by importing the Angular `core.dart` file,
+  so that we can use common types like `@Component` when we create
```
zoechi: Hmm, the type is actually `Component` instead of `Component`. If you use it with `@` it would probably be better to use `@Component()` (with `()`)

public/docs/dart/latest/tutorial/toh-pt3.jade
```diff
@@ -1,14 +1,251 @@
...
+  our component.
+
+  We create metadata with the `@Component` annotation where we
```
zoechi: same

public/docs/dart/latest/tutorial/toh-pt3.jade
```diff
@@ -1,14 +1,251 @@
...
+  Angular insists that we declare a ***target*** property to be an ***input*** property.
+  If we don't, Angular rejects the binding and throws an error.
+.l-sub-section
+  :marked
+    We explain input properties in more detail [here](../guide/attribute-directives.html#why-input)
+    where we also explain why *target* properties require this special treament and
+    *source* properties do not.
+:marked
+  There are a couple of ways we can declare that `hero` is an *input*.
+  We'll do it the way we *prefer*, by annotating the `hero` property with the `@Input` decorator.
```
zoechi: I guess `@Input()` (with `()`) would be better here. This is a mean bug, which at least in TS doesn't produce an error but  still breaks the code.

zoechi: IMHO in Dart it's called annotation instead of decorator. TS calls it decorator. They also have annotations AFAIK but I can't tell when it's the one and when the other.

public/docs/dart/latest/tutorial/toh-pt3.jade
kwalrath: I know what you mean, but it's pretty standard to not include the `()`. (E.g. https://www.dartlang.org/docs/dart-up-and-running/ch03.html refers to "a `@MirrorsUsed` annotation". Then again, https://www.dartlang.org/articles/reflection-with-mirrors/ just says "MirrorsUsed annotation".

I'll keep an eye out for whether people are confused by this. I like the clue that `@` provides, but resisted putting more characters after the classname...

public/docs/dart/latest/tutorial/toh-pt3.jade
kwalrath: Oops, I thought I'd caught all the "decorat*"s. (It's on my list of [common differences](https://github.com/angular/angular.io/wiki/Converting-Pages-to-Dart).)

Good catch.

PR #1073, opened 2016-04-11
Small typo fixes (missing dashes, and double "the")

ajimenez2712:
'We spell our file names in lower dash case (AKA "kebab case") so we don't worry about' should read 'We spell our file names in lower dash case (AKA "kebab-case") so we don't worry about'

### DIFF
--- a/public/docs/dart/latest/tutorial/toh-pt3.jade
+++ b/public/docs/dart/latest/tutorial/toh-pt3.jade
@@ -62,7 +62,7 @@ code-example(format="." language="bash").
 
     All of our component names end in "Component".  All of our component file names end in "_component".
 
-    We spell our file names in lower dash case (AKA "kebab case") so we don't worry about
+    We spell our file names in lower dash case (AKA "kebab-case") so we don't worry about
     case sensitivity on the server or in source control.
 
     <!-- TODO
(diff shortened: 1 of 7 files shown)


wardbell:
Oh my my. That's a lot of the the's.  Thanks.

public/docs/dart/latest/tutorial/toh-pt3.jade
```diff
@@ -62,7 +62,7 @@ code-example(format="." language="bash").
 
     All of our component names end in "Component".  All of our component file names end in "_component".
 
-    We spell our file names in lower dash case (AKA "kebab case") so we don't worry about
+    We spell our file names in lower dash case (AKA "kebab-case") so we don't worry about
```
kwalrath: Oops, this isn't actually true, so I'll need to fix it. (Dart uses lower_with_underscores, which has no cool name that I know of.)

You can either drop this file from this PR, or I'll wait for it to clear and fix the file later.

Foxandxss: That is snake_case. Also the PR is merged already :P

kwalrath: Yeah, I just saw that. :) snake_case. Cool.

PR #1074, opened 2016-04-12
docs(quickstart/js): add missing systemjs dependency

chalisegrogan:
The [5 Min Quickstart js version](https://angular.io/docs/js/latest/quickstart.html#!#package-json) is missing a required dependency (`systemjs`) that _is_ present in the [ts version](https://angular.io/docs/ts/latest/quickstart.html#!#package-json).

### DIFF
--- a/public/docs/_examples/quickstart/js/package.1.json
+++ b/public/docs/_examples/quickstart/js/package.1.json
@@ -8,6 +8,7 @@
   "license": "ISC",
   "dependencies": {
     "angular2": "2.0.0-beta.14",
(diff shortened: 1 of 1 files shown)


Foxandxss:
The JS version is not using System.js. Are you getting any error?

chalisegrogan:
@Foxandxss when I'm including `Rx.umd.js` & `angular2-all.umd.js`, I'm getting `Uncaught ReferenceError: System is not defined`. 
screen shot 2016-04-11 at 5 24 07 pm

Maybe it's something else, then.

PR #1079, opened 2016-04-12
chore(glossary): fix broken links

### LINKED ISSUE #1078: Broken links on website documentation
Broken links on website documentation

To reproduce go to:

https://angular.io/docs/ts/latest/glossary.html#!#A

Scroll down to:

Data Binding

All of those links you can see are broken links 404

Foxandxss:
I guess that the glossary was moved outside the guide so that explains the broken links.

PS: Didn't we have a tool to verify broken links?

Fixes #1078

### DIFF
--- a/public/docs/ts/latest/glossary.jade
+++ b/public/docs/ts/latest/glossary.jade
@@ -170,16 +170,16 @@ include _util-fns
     operations and supporting declaration syntax.
 
      The many forms of binding include:
-     * [Interpolation](template-syntax.html#interpolation)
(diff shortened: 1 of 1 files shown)


wardbell:
I don't think we've run it recently. Must check on that.

I caught a couple more.

There is a remaining problem.  We have hidden glossary "guard" pages within the `guide` directories. If someone had an old link to them there (and some of the docs pages might!), these repaired links would be broken. There is no easy way to make them right in both places.

I'm tempted to delete those guard pages but not sure what I'd break. Look at this another day.

Merging this.

PR #1093, opened 2016-04-15
styles.css missing from quickstart for js

stphivos:


stphivos:
I signed it!

wardbell:
Thanks ... but I'm declining this PR for reasons that you may regard as incorrect. They are reasons nonetheless.

We omitted the styles even though they are part of the QS application when you look at it in detail. We also omitted mention of the test configuration files even though _those_ are part of the [`angular.quickstart` repo](https://github.com/angular/quickstart).

***These omissions are deliberate.***  There are already too many steps and we didn't want to burden the text with anything we didn't consider essential. We don't think there is anything learned by adding css. And while there is much important to be learned by adding testing support, we didn't think that was essential to _starting quickly_ either.

The additional files have been include silently. Maybe that is a mistake. It's one we'll live with for now.

PR #1095, opened 2016-04-15
Update question-control.service.ts - to initialize the controller value

rexebin:
controller value now will be initialized with 'question.value', regardless if the controller has validation of 'required' or not.

### DIFF
--- a/public/docs/_examples/cb-dynamic-form/ts/app/question-control.service.ts
+++ b/public/docs/_examples/cb-dynamic-form/ts/app/question-control.service.ts
@@ -11,7 +11,7 @@ export class QuestionControlService {
     let group = {};
 
     questions.forEach(question => {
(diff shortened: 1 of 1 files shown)


rexebin:
I have signed.

thelgevold:
@wardbell This fix looks good to me.

bampakoa:
@rexebin I think you have to comment just 

> I signed it!

 for the bot to set the CLA correctly. :smile:

rexebin:
I signed it!

rexebin:
😁 the bot could be a little more flexible here😂

bampakoa:
It does not seem to respond so i guess there must be something else that needs to trigger :SMI

wardbell:
I'm convinced.  Merged.

PR #1096, opened 2016-04-16
HACK: tagDiagnostics processor

petebacondarwin:


Foxandxss:
What's the status of this PR? 👍

petebacondarwin:
This was only here to provide @naomiblack  with a quick and dirty look at what tags were being used and abused in the docs. I don't have any plans to implement it further nor incorporate it permanently.
Unless @naomiblack says otherwise I would just close this.

PR #1089, opened 2016-04-14
Restyle api docs for classes and directives

ericjim:
There is a live test link if you'd like to check it out: https://api-doc-styling-1.firebaseapp.com/

This PR is work towards closing https://github.com/angular/angular.io/issues/964, as it stands, there are already quite a bit of changes in the codebase. This is in an effort to make code review easier. I will continue fulfilling the proposal in https://github.com/angular/angular.io/issues/964 in a subsequent PR.

PR #1104, opened 2016-04-18
Change margin to example title style

bampakoa:
Increment bottom margin in example title so that the first line of the
example code is completely visible.

image

### DIFF
--- a/public/resources/css/module/_example-title.scss
+++ b/public/resources/css/module/_example-title.scss
@@ -9,7 +9,7 @@
   background: #1976D2;
   box-shadow: none;
   // temporary hack to remove space between example title and code-example
-  margin-bottom: -18px;
+  margin-bottom: -15px;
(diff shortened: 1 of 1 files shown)


Foxandxss:
I noticed that as well, that is a recent change or something.

ericjim:
I suspect it had something to do with the changes made here https://github.com/angular/angular.io/pull/1089, in particular this file https://github.com/ericjim/angular.io/blob/749a8cd3a6fbb0c4dd7d4e4a9a496220bf916407/public/resources/css/module/_code-box.scss

naomiblack:
thanks for the PR -- we have a bunch of CSS work ongoing as we improve the API doc styles. It's more helpful in this case to file an issue rather than a PR. Closing without merging, but we'll correct it in the next set of CSS changes. appreciate the report!

PR #1127, opened 2016-04-20
docs(toh): remove @Input decorator in part 5

### LINKED ISSUE #1126: ToH Routing leaves behind @Input (and possibly selector)
In the routing section, you start with the ``selectedHero`` passed in as an ``@Input``:

```html
<my-hero-detail [hero]="selectedHero"></my-hero-detail>
```

```typescript
export class HeroDetailComponent {
  @Input() hero: Hero;
}
```

You then show changing to use of a route:

> We will no longer receive the hero in a parent component property binding.

I _believe_ this means the ``@Input`` is no longer needed. But it still appears at the end in ``app/hero-detail.component.ts (latest)``. It also appears in the plunkr.

Also, though I'm not sure about this...I think the ``my-hero-detail`` selector should be removed from the component, as ``<my-hero-detail>`` is no longer used in a routing scenario.

Foxandxss:
Removes the @Input decorator in part 5 since it is not used and was added by mistake.

Fixes #1126

### DIFF
--- a/public/docs/_examples/toh-5/ts/app/hero-detail.component.ts
+++ b/public/docs/_examples/toh-5/ts/app/hero-detail.component.ts
@@ -2,7 +2,7 @@
 // #docregion
 // #docregion v2
 // #docregion import-oninit
-import { Component, Input, OnInit } from 'angular2/core';
+import { Component, OnInit } from 'angular2/core';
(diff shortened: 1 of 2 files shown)


naomiblack:
merged! thank you!

PR #1132, opened 2016-04-20
docs(cookbook): Added router cookbook

brandonroberts:


Foxandxss:
Ill leave this blocked because this needs update when we land new new router.

PR #1139, opened 2016-04-22
chore: rename Elvis to safe navigation operator

Foxandxss:
Here is the needed rename. I decided not to capitalize "Safe" in any case because I don't like how the text flows in that case.

@kwalrath I checked the Dart part, I don't see any difference on both TS and Dart guides after the changes.

Closes angular/angular#6387

### DIFF
--- a/public/docs/dart/latest/guide/template-syntax.jade
+++ b/public/docs/dart/latest/guide/template-syntax.jade
@@ -366,11 +366,11 @@ table
 :marked
   ## Template expression operators
   The template expression language employs a subset of Dart syntax supplemented with a few special operators
-  for specific scenarios. We'll cover two of these operators: _pipe_ and _Elvis_.
+  for specific scenarios. We'll cover two of these operators: _pipe_ and _safe navigation operator_.
 .callout.is-helpful
   header Dart difference: ?. is a Dart operator
   :marked
-    The Elvis operator (`?.`) is part of the Dart language.
+    The safe navigation operator (`?.`) is part of the Dart language.
     It's considered a template expression operator because
     Angular 2 supports `?.` even in TypeScript and JavaScript apps.
 +includeShared('{ts}', 'expression-operators-pipe-1')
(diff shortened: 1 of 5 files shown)


Foxandxss:
I also updated the cheatsheet, better now than forget later :)

https://github.com/angular/angular/pull/8196

public/docs/dart/latest/guide/template-syntax.jade
```diff
@@ -366,11 +366,11 @@ table
 :marked
   ## Template expression operators
   The template expression language employs a subset of Dart syntax supplemented with a few special operators
-  for specific scenarios. We'll cover two of these operators: _pipe_ and _Elvis_.
+  for specific scenarios. We'll cover two of these operators: _pipe_ and _safe navigation operator_.
 .callout.is-helpful
   header Dart difference: ?. is a Dart operator
   :marked
-    The Elvis operator (`?.`) is part of the Dart language.
+    The safe navigation operator (`?.`) is part of the Dart language.
```
kwalrath: That's not the name we've used in Dart, but I can live with it.

(For the curious: The Dart spec & language tour tend to avoid using a name for it, although the language tour does once call it _conditional member access_.)

kwalrath:
LGTM!

public/docs/dart/latest/guide/template-syntax.jade
Foxandxss: I don't mind changing it for something different, not a problem :)

Foxandxss: Still better name than no name I guess.

naomiblack:
Can you provide context for why we are making this change? @IgorMinar prefers the shorter "Elvis operator", so I'd like to understand the change.

IgorMinar:
Where does the need to rename this operator come from?

Foxandxss:
My bad. Ward asked me to, but there are the issues for it:

https://github.com/angular/angular/issues/6387

And here: https://github.com/angular/angular/issues/791#issuecomment-207197430

IgorMinar:
I see. Yeah. let's go ahead and rename. Thanks for making the PR. Can you please reference the other issues via `Closes #..`. thanks!

Foxandxss:
Can I reference PRs in other repo here with that? On the other hand I created a PR at angular repo (to rename the cheatsheet too) which closes the open issue there.

wardbell:
Thanks Jesus. Merging.

PR #1145, opened 2016-04-23
docs:(cookbook showing how to create a TreeView)

thelgevold:
@wardbell 
Here is the TreeView cookbook.

I decided to add an http request as part of this. Mainly because it's the number one ask from people who have been commenting on other TreeViews I've made in the past. It also makes it easier to vary the data since it's relatively painful to hand code tree objects. I linked to the Http Dev guide for the http configurations.


thelgevold:
Updated outdated PR to rc 1

thelgevold:
Updated to use core-js shim

thelgevold:
rebased latest master

Foxandxss:
Update to this too. @thelgevold could you please update the prose now that there is no `directives` array so the paragraph where you mention it (for the recursive thing).

I can do it, but I prefer that the original author does it :)

thelgevold:
Thanks! @Foxandxss
I tweaked the point about the recursive template.

PR #1133, opened 2016-04-21
chore(ts): updating to beta.16

filipesilva:
WIP - DO NOT MERGE

Examples still broken:
- pipes

PR #1178, opened 2016-04-26
chore(ts): updating to beta.16

wardbell:
Changes necessary to adapt doc samples to beta.16

Filipe did most of the work in #1133 while waiting for beta.16 to be ready. 
Ward just nursed it the last few inches.


wardbell:
Chill out, googlebot. This is Filipe's commit with a little rub-a-dub-dub from me.

PR #1181, opened 2016-04-26
WIP: A11ydoc - copyedits

naomiblack:
This is still too long... I'm concerned that developers who need to learn about accessibility (but maybe don't share our passion for the topic) won't have the patience to read a 30-page document.

@AlmeroSteyn -- I did a quick copyedit to make the tone a little more dry so it matches the other cookbooks, and to try to keep it more tightly focused to just the essentials. I cut the length by about a third, but I'm worried that it's still quite long.

@wardbell have a look and see if you can tighten it up further?


AlmeroSteyn:
This PR is outdated and should be closed for preference of https://github.com/angular/angular.io/pull/1625

Issue #1176, opened 2016-04-25
badges are too low in api doc


ericjim:
Just a reminder for myself to fix.

PR #1187, opened 2016-04-26
fix(badge align): fix alignment of badges on smaller screens

ericjim:
### Changes
* Badges should align more with the hero title of every api doc page
* Badges should re-align properly when viewed on smaller screens
* Reduce hero title in smaller screens to prevent cut off text

### Preview
Try this on your browser, and resize the the window for the full effect. It should work in both Chrome and Safari.
* https://badges-everywhere.firebaseapp.com/docs/ts/latest/api/common/NgSwitch-directive.html (Stable for demonstration purposes only, not actually in the docs)
* https://badges-everywhere.firebaseapp.com/docs/ts/latest/api/common/NgPluralCase-directive.html (Single badge)
* https://badges-everywhere.firebaseapp.com/docs/ts/latest/api/common/FORM_BINDINGS-let.html (two badges)

### Comments
I thought I had this fix from before, but some change somewhere affected the badges, hopefully this is a more permanent solution. @petebacondarwin @naomiblack 

CLOSES:
https://github.com/angular/angular.io/issues/1176

### DIFF
--- a/public/_includes/_hero.jade
+++ b/public/_includes/_hero.jade
@@ -8,8 +8,9 @@ if current.path[4] && current.path[3] == 'api'
   - var textFormat = 'is-standard-case'
 
 header(class="hero background-sky")
-  h1(class="hero-title text-display-1 #{textFormat}") #{headerTitle}
-  if useBadges
+  div(class="inner-header")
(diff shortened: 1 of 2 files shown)


naomiblack:
merged and live!

PR #1193, opened 2016-04-26
First Draft of JavaScript tutorial - Part 1

danielzen:
This is my first pull request. Please let me know if I need to create a new branch, and I will.

I have been waiting for the JavaScript Tutorial, and decided to start writing it myself. I have take care so that all the snippets & the plunker for toh-1 are working well, and all references to TypeScript have been purged from this version. Note that I have elected to use Template Literals which are now available in all the major browsers. I could refactor away from them if necessary.

I have not yet created the completed toh app in JS which will be referenced in the Introduction, but I say that it is coming soon, since I will create it as I will continue to work on the JS tutorials if this first part is accepted. Thanks!


wardbell:
The branch doesn't bother me (although I'd have created a branch for myself). We can work around it.

It is so generous of you to work on this contribution. Unfortunately, I fear that you'll have to rework it entirely as we are about to refresh the entire ToH series to align with the CLI and I'm afraid of what that will do to your flow. 

The CLI is typescript only. In a sense, that leaves the JS version free to remain as is. But I'm not sure that is a good idea and it will complicate our end-to-end testing.

No matter what, you will want to update for RC1 as well.

Because there is so much that you'd have to redo ... I'm going to close this. We can revisit when things settle down. 

Please don't take that as an adverse judgment. We're just not in a good position to correct and absorb this right now.

Please hang on to your thoughts and revisit in a few weeks.   All the best, Ward

danielzen:
Thanks for getting back to me. Now that a few weeks have passed, I'm curious to follow-up.

I have no problem reworking all of this. I have been working on a course that is now Angular 2 for ES6 (formerly ES5y). The real question: Should the JS version of the documentation on the Angular.io site be about ES6 or ES2015?

Most of what I have been doing has been as an exercise for me, but I would love for it to help others. And, I figure expanding the JS documentation would be best. 

Finally, is the thought that there should be no official JavaScript documentation until ng2 launches so that the direction will be clear by then?

PR #1199, opened 2016-04-27
docs(cb-a1-a2): make custom pipe example compatible with beta.16 changes

teropa:


wardbell:
Incorporated already in a different, equally satisfying way. Closing. Thanks though

PR #1201, opened 2016-04-27
docs(lyfecycle-hooks-dart): Update Lifecicle hooks dart docs and Sources

luisvt:
* add lifecycle-hook.jade
* correct dart and html files
* add sample.css


chalin:
Sorry, again because we've adopted a new approach to generating the Dart versions of the dev guide chapters, we won't be merging this PR.

@luisvt / @kwalrath: please close.

PR #1203, opened 2016-04-28
WIP docs(dgeni): trying to figure out why Location-class isn't in API list

naomiblack:


tools/api-builder/angular.io-package/index.js
```diff
@@ -50,6 +50,7 @@ module.exports = new Package('angular.io', [basePackage, targetPackage, cheatshe
     'angular2/instrumentation.ts',
     'angular2/platform/browser.ts',
     'angular2/router.ts',
+    'angular2/router/location/location.ts',
```
petebacondarwin: The Location classes have been moved to `platform/browser` : https://github.com/angular/angular/blob/master/modules/angular2/src/platform/browser/location/location.ts

PR #1206, opened 2016-04-28
chore(tslint): remove component & directive prefix rules

mgechev:


Foxandxss:
As a note for this, our guide has different prefixes to show different concepts so this rule makes the lint to fail on those cases.

Rule is better for a concrete project than this site.

wardbell:
It appears that this was incorporated in some other way. Closing

Issue #1121, opened 2016-04-19
Create and Implement Press Kit 


amitafr:
The press kit will be a new page with a link located in the footer under Resources
![angular io press kit](https://cloud.githubusercontent.com/assets/10601477/14679934/f37c8f1e-06e9-11e6-991c-3063830c921d.jpg)

Downloadable assets will be sent directly to @ericjim.

naomiblack:
I'd like some text that makes it more explicit that the images on this page are licensed under CC BY 4.0. Maybe just body text under the top heading, saying
`The logo graphics available for download on this page are provided under CC BY 4.0` with a link to the license.

amitafr:
@naomiblack I like that text a lot! Will update the mock to reflect that

Issue #1120, opened 2016-04-19
Update Footer Organization and Links


amitafr:
Update Footer Organization & Links

Section 1: Resources 
- Downloads
- Libraries
- Press Kit
- Etc. 

Section 2: Learn
- Docs
- Books
- Videos
- etc 

Section 3: Help & Support
- Support Page
- File a Ticket
- Product Feedback

Section 4: Community 
- Events
- Twitter
- Facebook 
- Google+ 
- Github
- etc.

ericjim:
@alexwolfe I'll be working off these for the footer

PR #1210, opened 2016-04-28
feat(presskit): angular logo press kit

ericjim:
### Changes
* Add a press kit page to angular.io

### Preview
* Only way to access for now: https://angular-io-presskit.firebaseapp.com/presskit.html

### Known issues
* Currently not accessible via navigation, this will be fixed with https://github.com/angular/angular.io/issues/1120 which is next.
* If there is a place to put the presskit with the current navigation, I can make the addition, however, that is very soon to change. @naomiblack 

CLOSES:
https://github.com/angular/angular.io/issues/1121

@naomiblack @petebacondarwin


naomiblack:
merged and live!

PR #1217, opened 2016-04-29
fix(footer): use cursor pointer on feedback in footer

ericjim:
### Changes
* Use a cursor pointer on cover when hovering on `Product Feedback`

### Preview
* https://footer-and-nav.firebaseapp.com/

@naomiblack


naomiblack:
live and in master!

PR #1228, opened 2016-05-01
docs(style-guide): hide .avoid boilerplate

Foxandxss:
/cc @johnpapa @wardbell 

This PR will hide all the boilerplate needed to make the .avoid to compile but that we don't need to see on the guide.


Foxandxss:
Closing as I merged this to the PR6.

PR #1229, opened 2016-05-01
docs(style-guide): make examples comply with the linter

Foxandxss:
/cc @johnpapa @wardbell 

For this I had to remove two rules from our linter. They don't make sense for the style-guide.

All the issues where "This is imported, never used" so I used my imagination to use those stuff but still hide it from the guide. So there are no visual changes.

I put some providers and stuff in components too. When we do testing here, we will see if we need to move this providers around, but for now this works.


johnpapa:
LGTM

Foxandxss:
Merged then.

PR #1230, opened 2016-05-01
docs(style-guide): fix 06-03

Foxandxss:
/cc @johnpapa @wardbell 

Somehow the .avoid was the good one and the good one was the avoid. Also the guide was showing only one as good and bad.


Foxandxss:
Closing as merged in the PR 6.

PR #1241, opened 2016-05-02
docs(cli-quickstart): add cli-quickstart chapter

filipesilva:
This PR includes both the Angular-CLI quickstart chapter and e2e test support for CLI applications. 

It does not add the CLI-Quickstart to the sidebar menu though, nor live example via plunker.

To run e2e tests for CLI applications, you must have Angular-CLI installed globally and using the same version as specified in `public/docs/_examples/package.json`.

When adding a new CLI-based chapter, create a project using `ng new chapter-name` and copy the contents of the `chapter-name` dir onto `public/docs/_examples/chapter-name/ts/`.

Afterwards, add an `example-config.js` in the example project root, rename `package.json` to `package.1.json`, remove `declare var module: { id: string };` from `src/typings.d.ts` (angular.io typings already include `node`).

Finally, add `public/docs/_examples/chapter-name/e2e-spec.js` as usual.


filipesilva:
@wardbell I incorporated your changes and removed the `verbose` sections.

Before this is merged, the CLI should be updated to new world and the example app re-generated. After that, docregions need to be re-added.

wardbell:
Still holding out for your approval

filipesilva:
@wardbell PR is updated and ready. Description of PR also updated to reflect all changes and CLI chapter creation instructions.

public/docs/ts/latest/cli-quickstart.jade
```diff
@@ -0,0 +1,261 @@
...
+figure.image-display
+  img(src='/resources/images/devguide/cli-quickstart/app-works.png' alt="Our app works!")
+  
+.alert.is-important
+  :marked
+    _Windows Developers_: perform these commands in a console window opened as an administrator.
+    The Angular-CLI build steps create symlinks which require administrator rights under Windows.
+
+:marked
+  Now we're ready to write some code... or are we? We certainly could start writing code right now, but our tests would break.
```
Foxandxss: Are we or are we? Sounds weird for me.

public/docs/ts/latest/cli-quickstart.jade
```diff
@@ -0,0 +1,261 @@
...
+  It's a good idea to run tests before and after making changes to confirm that our app behaves as we expect it to.
+
+  *At last* we are ready to write some code.
+      
+a(id="edit")
+.l-main-section
+:marked
+  ## Edit the app
+  
+  We open the `angular2-cli-quickstart` projecd folder in our favourite editor.
```
Foxandxss: projecd typo

public/docs/ts/latest/cli-quickstart.jade
```diff
@@ -0,0 +1,261 @@
...
++makeExample('cli-quickstart/ts/src/app/angular2-cli-quickstart.component.css', null, 'src/app/angular2-cli-quickstart.component.css')(format=".")
+  
+:marked
+  Looking good!
+  
+figure.image-display
+  img(src='/resources/images/devguide/cli-quickstart/my-first-app.png' alt="Output of QuickStart app")
+
+:marked
+  ### Re-test and fix
```
Foxandxss: I have an issue with this label, the way it looks in the final product, seems like a footer for the image instead of a new section title.

Foxandxss:
I reviewed it. All the steps works good on a new node environment, including the deploy thing.

Leaving this aside, I thought this cookbook was a replacement for the actual 5 min quickstart and not just a quickstart **on** CLI.

As a cookbook in how to get started on the CLI is good, but we still need a replica of the "5 min quickstart" but using the CLI. Same content, same things to add, but using the CLI.

I find a bit confusing (or verbose) having to update both e2e and unit tests (which makes sense, but still verbose for the reader). Would be good if different changes could be made so both tests would need different set of changes.

On the other hand, do we really need a whole set of instructions of how to set github? I would personally add a subsection on "Learn more in how to set a token here"

Can't wait to have the CLI in more places, I love how simple it is.

Foxandxss:
Won't close this because I still need to extract a CB from here and I don't want the branch to be removed by mistake.

PR #1242, opened 2016-05-02
chore: make protractor report failed suites

Foxandxss:
So I broke a test, so sorry, protractor deceived me with:

So I debugged it and we were giving the "ok" for each suite, no matter what, so the failed suite array was never used. The function was already receiving the result, I just needed to return it. Now we get:

/cc @naomiblack @wardbell


filipesilva:
Nice. Really useful!

wardbell:
Thanks! Merged.

PR #1256, opened 2016-05-02
docs: add RxJs operators to hero.service.ts example

jakegarelick:
Proposed change adds RxJs operator import to hero.service.ts example - my motivation for this pull request was this issue: https://github.com/angular/angular/issues/5632#issuecomment-200444642.

Since RxJs operators are not mentioned until after the code example, I figure that this would be useful for anyone skimming the docs.

### DIFF
--- a/public/docs/_examples/server-communication/ts/app/toh/hero.service.ts
+++ b/public/docs/_examples/server-communication/ts/app/toh/hero.service.ts
@@ -10,7 +10,8 @@ import {Headers, RequestOptions} from 'angular2/http';
 // #enddocregion import-request-options
 // #docregion v1
 import {Hero}           from './hero';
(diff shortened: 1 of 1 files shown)


jakegarelick:
I signed it!

wardbell:
Interesting suggestion, Jake. I certainly appreciate the feedback.

However, I'm not ready to agree.  Yes, we do apply rxjs operators before we talk about them. Maybe you have a better idea for teaching this.  But I don't think the solution is to add an import that you don't actually need.

We do call out the need to [*Enable RxJS Operators*](https://angular.io/docs/ts/latest/guide/server-communication.html#!#enable-rxjs-operators) in the http chapter.  

I'm closing your PR with a tinge of regret.  Thanks for helping.

PR #1257, opened 2016-05-03
feat(docs landing): new docs landing

robertmesserle:
Continued work on #1255 and added fix for page flicker on docs landing on load.

@naomiblack @ericjim


naomiblack:
merged and live.

PR #1264, opened 2016-05-03
docs: change packageNames in systemjs.config.1.js

sealday:
Change @angular/router to @angular/router-deprecated to make tutorial work.


Foxandxss:
Thank you, forgot to update that bit.

wardbell:
Thanks, but that is incorrect. We need *both* package names as both packages exist. This is being handled in a forthcoming docs update to router documentation

PR #1265, opened 2016-05-03
docs(server-communication): remove unreachable and confusing code chunk

akozhemiakin:
Maybe I misunderstood something and the proposed change is entirely wrong, but I decide to create a pull request instead of an issue in case if I am right.

I think the removed (in this the proposed commit) chunk of code is unreachable because the callback passed to the Observable#map method will never be called, if the response status code is <200 or >=300. Its due to the current implementation of the backend (see here https://github.com/angular/angular/blob/master/modules/%40angular/http/src/backends/xhr_backend.ts).
If the status code is not within this "successful" range it will call observable error method anyway and the callback passed to the Observable#map will not be fired. So there is no sense (and it was confusing to me) in trying to handle non successful responses in the callback passed to the Observable#map.

### DIFF
--- a/public/docs/_examples/server-communication/ts/app/toh/hero.service.1.ts
+++ b/public/docs/_examples/server-communication/ts/app/toh/hero.service.1.ts
@@ -34,9 +34,6 @@ export class HeroService {
   }
 
   private extractData(res: Response) {
-    if (res.status < 200 || res.status >= 300) {
-      throw new Error('Bad response status: ' + res.status);
-    }
     let body = res.json();
     return body.data || { };
   }
--- a/public/docs/_examples/server-communication/ts/app/toh/hero.service.ts
+++ b/public/docs/_examples/server-communication/ts/app/toh/hero.service.ts
@@ -59,9 +59,6 @@ export class HeroService {
 
   // #docregion extract-data
   private extractData(res: Response) {
-    if (res.status < 200 || res.status >= 300) {
-      throw new Error('Bad response status: ' + res.status);
-    }
     let body = res.json();
     return body.data || { };
   }


Foxandxss:
Anything that could be potentially wrong deserves a pull request / issue.

We will take a peek to see what is going on.

Thank you.

wardbell:
>the callback passed to the Observable#map method will never be called, if the response status code is <200 or >=300. 

Sadly I believe you are misinformed. The current http and backend pass responses with status codes outside the 200 range as SUCCESSFUL results. I read the source code you cited and it confirms my understanding. I've also tested it. Where do YOU see that it redirects the observable on to the failure path?

It is perfectly reasonable for these Angular widgets to behave this way. The server did not fail in its response. It delivered a response ... that happens to be a 404 or whatever. It's up to the caller to determine whether that was the desired result (maybe you wanted a 404?) or not.

A future version of http may behave differently. But today's version does exactly what I said.  It is the responsibility of the app to weed out responses with "bad" response codes ... exactly as the sample shows.


We make that point [_explicitly_ in the Http chapter](https://angular.io/docs/ts/latest/guide/server-communication.html#!#bad-status-codes).

Thanks for contributing.

akozhemiakin:
And what about this chunk?
```typescript
if (isSuccess(status)) {
  responseObserver.next(response);
  // TODO(gdi2290): defer complete if array buffer until done
  responseObserver.complete();
  return;
}
responseObserver.error(response);
```
And this is the isSuccess function implementation itself
```typescript
export const isSuccess = (status: number): boolean => (status >= 200 && status < 300);
```

wardbell:
Hmmm. I can assure you that was NOT there when this chapter was written. 

Moreover, while I anticipate seeing such logic in a future version of http, it is definitely an error (I will report it) for the low level `XhrConnection` to do this. It is up to the the higher level module to determine how to handle a non-200 response.

Thanks for forcing the issue. Looks like we'll have to fix the chapter.

PR #1266, opened 2016-05-03
docs: update all docs to partially comply the style-guide

Foxandxss:
With this PR, all the most important changes of the style guide are applied. I talk about how we manage imports (spaces between keys, alphabetical order, leaving a new line between angular imports and app code...) and I also removed the underscore.

Also checked that the documents themselves are not mentioning the old variables with underscore.

Only missing part here is the testing source and guide, I need to figure out some things before.


Foxandxss:
All tests are passing too (e2e).

PR #1267, opened 2016-05-03
docs(server-communication): fix broken example

Foxandxss:
Welp, this region was hiding from me (They are no that obvious).

I updated it in the same way I did with rxjs. Also I saw that we changed from in memory library so I updated that as well. Please double check that because I didn't know that bit.


wardbell:
Thanks but this isn't right either.  There is no reason to be loading anything. No reason for a local system.config.js.  I'm keeping this open as a reminder to fix this chapter ... and _every_ chapter that tries to load anything.

Foxandxss:
My reasoning behind this change was to leave it with the same content as before. There are **3** cases in this file where we give such instructions. 

I can modify later today this PR to get rid of all of them in a good way. 

PS: there are no other file with instructions as far as I  can remember. ToH routing **had** one.

Foxandxss:
@wardbell I updated the PR. All 3 cases where any installation is performed is now gone without affecting the lecture.

in the Rxjs part I left some parts of the old explanation but I think they sense in that context.

PR #1270, opened 2016-05-03
docs(quickstart): update to new world

Foxandxss:
Please review and reword as needed.


Foxandxss:
Hold. Forgot two things. There is a #tbd link to system.js config part. And a sub section ( the one with //TODO) that needs to be removed or moved into "configuring System.js".

Will update later

Foxandxss:
This is good for review now.

Issue #1289, opened 2016-05-05
StyleGuide Style 07-03 BAD FILENAME


blackpr:
The file app/heroes/hero-list/hero-list.component.ts shows a BAD FILENAME error:

`BAD FILENAME: ../../../_fragments/style-guide/ts/07-03/app/heroes/hero-list/hero-list.component.ts.md   Current path: docs,ts,latest,guide,style-guide PathToDocs: ../../../
`

wardbell:
It isn't now. The `hero-list.component.ts` is also a direct child of `heroes` and there is no `heroes/hero-list` folder. Maybe this was fixed before I looked.

wardbell:
You're write. Merged in PR #1290

PR #1290, opened 2016-05-05
Fix  issue #1289 BAD FILENAME

blackpr:
I think that this fix https://github.com/angular/angular.io/issues/1289

### DIFF
--- a/public/docs/_examples/style-guide/ts/07-03/app/app.component.ts
+++ b/public/docs/_examples/style-guide/ts/07-03/app/app.component.ts
@@ -1,7 +1,7 @@
 // #docregion
 import { Component } from '@angular/core';
 
-import { HeroListComponent } from './heroes/hero-list.component';
+import { HeroListComponent } from './heroes/hero-list/hero-list.component';
 import { HeroService } from './heroes/shared/hero.service';
 
 @Component({
--- a/public/docs/_examples/style-guide/ts/07-03/app/heroes/hero-list/hero-list.component.ts
+++ b/public/docs/_examples/style-guide/ts/07-03/app/heroes/hero-list/hero-list.component.ts
@@ -1,8 +1,8 @@
 // #docregion
 import { Component, OnInit } from '@angular/core';
 
-import { HeroService } from './shared/hero.service';
-import { Hero } from './shared/hero.model';
+import { HeroService } from '../shared/hero.service';
+import { Hero } from '../shared/hero.model';
 
 @Component({
   selector: 'toh-heroes',


Foxandxss:
Yes, that does fix it. Thanks.

blackpr:
I signed it!

mrkiffie:
Thanks @blackpr, this issue was bugging me too. 😄

I think this change addresses the symptom rather that the issue. I'd like to propose a change that moves the file to the expected location and fixes the references in the related files.

I have made a [patch](https://github.com/mrkiffie/angular.io/commit/5e0e873b488c75ef4ecf2856b3f916dc45180d0b) that would adjust the structure to match [style 04-06](https://angular.io/docs/ts/latest/guide/style-guide.html#!#style-04-06). 

I am happy to submit it as a separate pull request if you would prefer doing it that way.

Foxandxss:
Now that you mention it @mrkiffie, you are right, this only fixes the symptom. Let see if @blackpr wants to update his PR or you can create a new one.

public/docs/_examples/style-guide/ts/07-03/app/app.component.ts
```diff
@@ -1,7 +1,7 @@
 // #docregion
 import { Component } from '@angular/core';
 
-import { HeroListComponent } from './heroes/hero-list.component';
+import { HeroListComponent } from './heroes/hero-list//hero-list.component';
```
Foxandxss: one `/` please.

blackpr:
Sorry about that.  It is my first PR

wardbell:
Strange but I don't see the problem for which this is a cure.

The `hero-list.component.ts` is currently a direct child of `heroes` and there is no `heroes/hero-list` folder. Maybe this was fixed before I looked.

mrkiffie:
Hi @wardbell. Thanks for all your great work on the docs. They have been extremely helpful. 😄 

Unfortunately the second tab of [Style 07-03](https://angular.io/styleguide#!#style-07-03)'s example is still currently broken.

Here is a screen shot of what I'm seeing
BAD FILENAME on second tab of style 07-03's example

It is the `hero-list.component.ts` that is missing in the second tab of the example. It is referencing `app/heroes/hero-list/hero-list.component.ts` but as you say that component file is a direct child of `heroes`.

The initial fix only changed the reference to the file, which resolved the `BAD FILENAME`. I had then suggested to update it according to [Style 04-06](https://angular.io/styleguide#!#style-04-06) as the tab's title was using that same structure. 

I hope this provides enough information to clarify. Could we possibly reopen it?

wardbell:
You're right! Thanks ... and thanks for persisting. Merged.

PR #1291, opened 2016-05-05
docs(router-deprecated): link doc to deprecated examples

Foxandxss:
Some of the style guide changes were lost on the migration to deprecated router but also the guide was pointing to the new router code instead of deprecated one.

/cc @wardbell


Foxandxss:
I also added plunker support (it was pointing to new router demos too)

Foxandxss:
Also the API to point to deprecated.

PR #1285, opened 2016-05-05
docs(dev guide): lifecycle-hooks - updated dart/ts code and new dart prose

chalin:
**TS** `guide/lifecycle-hooks` updates:
- example code clean-up; **e2e tests pass**
- `.jade` file:
  - mainly copyedits, but also
  - prep for extension by Dart .jade file

**Dart** guide/lifecycle-hooks:
- Significant update of example code (so it matches TS)
- Dart `.jade` extends TS `.jade` file with minor overrides

PR #1299, opened 2016-05-05
fix(lifecycle-hooks): fix typo and bad import

wardbell:
We love workshops!


chalin:
FYI, both these issues were fixed in #1285.

PR #1306, opened 2016-05-06
Fix incorrect gulp task name in README

sturdynut:
On line 65 of the README it references the `serve-and-sync-api-docs` gulp task.

Looking at the `gulpfile.js` there is a task named `serve-and-sync-api`, so I am assuming the desired change would be to update the README and not the gulp task name.

This PR updates the reference in README from `serve-and-sync-api-docs` to `serve-and-sync-api`.


wardbell:
you are correct. Thanks!

PR #1308, opened 2016-05-06
doc(quickstart): add bootstrap warning

AlmeroSteyn:
While going through the quickstart as a refresher I noticed that `Twitter Bootstrap` is being loaded in the `package.json`.

There is no further mention or use of it in the rest of the document **BUT** an explanation follows later on the `bootstrap` function of Angular 2.

I think beginners may link up the two and think that loading `Twitter Bootstrap` is mandatory.

So I added a small message to clarify it.


wardbell:
Like it

PR #1312, opened 2016-05-06
chore: add custom typings for examples

Foxandxss:
Currently `component-style` is broken because it can't find the `module` thing.

I added a custom.d.ts that will be added to typings and then it will be available everywhere.

### DIFF
--- a/public/docs/_examples/custom.d.ts
+++ b/public/docs/_examples/custom.d.ts
@@ -1,0 +1,1 @@
+declare var module: {id: string};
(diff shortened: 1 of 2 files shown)


Foxandxss:
As @chalin pointed out, this won't work on the live plunker, so I am open to ideas here.

wardbell:
I thought I HAD added this typings file (as `typings.d.ts` rather than `custom.d.ts`) but somehow failed to commit it. Could have been impeded by the .gitignore.

Thanks for figuring this out. But I took a different approach to the problem and described it in the documentation.

See PR #1355.

We are **TOTALLY HOSED** in plunkers for the moment because CLI and Style Guide are committed to the commonjs `module.id` and I can't find any transpile options that will make it work in plunker.

I tried the following and it doesn't work:
```
  typescriptOptions: {
      emitDecoratorMetadata: true, 
      module: 'commonjs' // should have done the trick!
    },
``

PR #1313, opened 2016-05-06
docs(dev guide): attribute-directives - move html excerpt to .html file

chalin:
Jade can’t parse inlined Angular adorned HTML properly, so it is best to put such source in a file and include with `makeExample`.


Foxandxss:
I won't accept this PR unless the background is #BADA55 (is green too)

just kidding.

chalin:
@thso: thx!
@Foxandxss : lol

PR #1322, opened 2016-05-07
docs(dev guide): cleanup of deprecated Jade mixin

chalin:
Replaced uses of `liveExLinks` mixin and then deleted its definition.


thso:
LGTM

PR #1325, opened 2016-05-08
docs(glossary): no RouteConfig

fmalcher:
Decorator @RouteConfig -> @Routes


wardbell:
Thanks. I'm going to decline for now because this just scratches the surface. Everything here describes the Beta Deprecated router.  We're staying with that until we're ready to talk about the Release Router which will be substantially different/improved.

PR #1326, opened 2016-05-08
docs(glossary): export in barrels

fmalcher:
The example is about exports, so `import` statements are wrong here, I suppose


wardbell:
How did I blow that? Good catch. Thanks

PR #1335, opened 2016-05-09
fix(docs landing): make docs landing responsive

ericjim:
### Changes
* Arrange doc landing cards in a column layout whenever the row runs out of space

### Preview
Live Preview: https://col-fix-docs.firebaseapp.com/docs/ts/latest/

#### Before
Before

#### After
After

@naomiblack


naomiblack:
merged and live!

PR #1339, opened 2016-05-09
docs(style-guide): make each style clickable for easier referencing

olore:
This will make it easier to capture the link to a specific style so that it can be shared.


KyleMcNutt:
Just noticed this pull request, there is other discussion on this topic going on in #1397 and #1400, for those interested.

olore:
Thanks @KyleMcNutt. Our commits are very similar. One thing mine does different (which I prefer) is that I moved the anchor to the name of the rule (i.e. "Rule of One") That way when the page loads, it shows the rule name at the top, then the rule number below it. I find the name very useful.

KyleMcNutt:
@olore I can see that perspective, I was going for the minimal change just to make it more obvious what can be linked to. It looks like from discussion in the other issues that an on-hover icon seems to be the preferred approach, but I'll leave that to the Angular teams designers to decide.

johnpapa:
@wardbell I am OK with this.. Are you?

wardbell:
Thanks. Merged as ce05f9b4f

PR #1343, opened 2016-05-10
docs(forms): Now word is interpreted as keyword

bampakoa:
The word `Now` is interpreted as keyword and does not render correctly with its style.


wardbell:
Good catch. Thanks.

PR #1352, opened 2016-05-10
docs(toh): put a bit of clarification

Foxandxss:
So @johnpapa got some feedback on the ToH and asked me to try to improve it.

Let me go step by step:

_Multiple components_

>Let's add that array property to the bottom of the @Component configuration object, immediately after the template and styles properties.
>>**Which file?**
>>**Also, does it matter where in @Component configuration this is put?**

I put a label in the example to specify which file and also reworded the sentence a bit to tell that where to put the array is a matter of choice than a fixed place.

>Let’s verify that we have the following structure

>>**You might want to say “.js and .map files not shown**

This part refers to the filetrees. I personally don't think that we need to explicitly say that the .js and .map files are not shown.

---

_Services_

>The Injector has a container of previously created services. Either it finds and returns a pre-existing HeroService from its container or it creates a new instance, adds it to the container, and returns it to Angular.

>>**Say a bit more about the container. Is this an actual object I could interact with?**

After this bit, we have a link about "Learn more about dependency injection". Personally I don't think we can or even if we should talk more about containers in the tutorial. Truth being said, we never ever say the word "container" anymore in either "Dependency Injection" or "Hierarchical injectors" guides. Perhaps we could do some work on "Dependency Injection" to name those containers so the reader won't get confused.

>Recall that the AppComponent creates an instance of HeroDetail by virtue of the <my-hero-detail> tag at the bottom of its template. That HeroDetail is a child of the AppComponent.
If the HeroDetailComponent needed its parent component's HeroService, it would ask Angular to inject the service into its constructor which would look just like the one for AppComponent:

>>**This may require a bit of rewriting into smaller conceptual bites.**

>constructor(private heroService: HeroService) { }

>>**I don't think you told me to add the import statement first.**

This is a blue box on the guide that explains that if we want to inject `HeroService` in the child component, we shouldn't use the providers array. What I did here is to update the code example label to say that that is an example. It is easy for the reader to get confused and think that the snippet is something that they should code. So yes, no import statement because that is just an example.

That being said, I think that we could get rid of the box. That is just like an excerpt of the `Dependency Injection` guide and we cannot rewrite it into smaller conceptual bites, because that is what the `Dependency injection` guide does. I suggest removing the blue box or making it more explicit that this is an excerpt of the `Dependency Injection` guide and that they can learn more there.

>Appendix: Shadowing the parent's service

>>**It isn’t clear what creates the parent-child relationship**

This appendix is a continuation from the blue box and again, something that is already explained elsewhere and confuses the reader. I would drop it too.

---

So I updated the parts I see value in and if someone has a good feedback on the other points, I will update this PR.

### DIFF
--- a/public/docs/ts/latest/tutorial/toh-pt3.jade
+++ b/public/docs/ts/latest/tutorial/toh-pt3.jade
@@ -171,7 +171,7 @@ code-example(format=".")
 :marked
   The `AppComponent`’s template should now look like this
 
-+makeExample('toh-3/ts/app/app.component.ts', 'hero-detail-template', 'app.component.ts (Template)')
++makeExample('toh-3/ts/app/app.component.ts', 'hero-detail-template', 'app.component.ts (Template)')(format='.')
 :marked
   Thanks to the binding, the `HeroDetailComponent` should receive the hero from the `AppComponent` and display that hero's detail beneath the list.
   The detail should update every time the user picks a new hero.
@@ -189,8 +189,7 @@ code-example(format=".")
   
   We tell Angular about it by listing it in the metadata `directives` array. Let's add that array property to the bottom of the
   `@Component` configuration object, immediately after the `template` and `styles` properties.
-+makeExample('toh-3/ts/app/app.component.ts', 'directives')
-
++makeExample('toh-3/ts/app/app.component.ts', 'directives', 'app/app.component.ts (Directives)')
 
 :marked
   ### It works!
@@ -215,7 +214,7 @@ code-example(format=".")
       .file hero-detail.component.ts
       .file main.ts
     .file node_modules ...
-    .file typings ...      
+    .file typings ...
     .file index.html
     .file package.json
     .file tsconfig.json
--- a/public/docs/ts/latest/tutorial/toh-pt4.jade
+++ b/public/docs/ts/latest/tutorial/toh-pt4.jade
@@ -167,7 +167,7 @@ code-example(format="." language="bash").
 
   ### Inject the *HeroService*
   
-  Two lines replace the one line of *new*:
+  Two lines replace the one line that created with *new*:
   1. we add a constructor.
   1. we add to the component's `providers` metadata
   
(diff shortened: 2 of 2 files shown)


public/docs/ts/latest/tutorial/toh-pt3.jade
```diff
@@ -187,10 +187,9 @@ code-example(format=".")
   
   We've imported `HeroDetailComponent`, we've used it in the template, but we haven't told Angular about it.
   
-  We tell Angular about it by listing it in the metadata `directives` array. Let's add that array property to the bottom of the
-  `@Component` configuration object, immediately after the `template` and `styles` properties.
-+makeExample('toh-3/ts/app/app.component.ts', 'directives')
-
+  We tell Angular about it by listing it in the metadata `directives` array. Let's add that array property to the `@Component` configuration object,
```
johnpapa: I prefer the original wording, just with the name of the file in it.

johnpapa:
I don't see the changes to the "services" that you mention above. Where are they?

Foxandxss:
> I don't see the changes to the "services" that you mention above. Where are they?

I just changed the excerpt to say that it is an example so it shouldn't be coded in the real app. Apart from that, I exposed there my personal opinion on what we should do, but didn't do the changes.

So the idea here is to agree on something so I can update this PR with those changes.

johnpapa:
I'm not sure I see a problem with leaving that part as it is. But if you feel strongly, let's get some other feedback too.

Foxandxss:
The person that gave the feedback things that the gray box (that explains hierarchical injection) is a bit confusing and should be explained more "step by step". I can understand that the gray box is a bit "complicated", but that gray box is just an excerpt of another guide we have, so instead of rewording the gray box, we can point people to the original guide.

There is also another part in services that talks about injections and containers and that person is confused about the word "containers". Containers is a DI concept but our DI guides never say "containers", that could be confusing.

wardbell:
@Foxandxss I agree with ***all*** of your points including deleting all the complicated stuff about containers and what might happen if we did x, y, or z.

Please make it so in this branch/PR and let me know when you want me to review it.

Foxandxss:
@wardbell Ready. I got rid of all those excerpt that tries to explain too much about DI and hierarchical DI and replaced them with links (or nothing for the appendix).

PR #1353, opened 2016-05-11
[review-pending] docs(dev guide): pipes - new dart prose; updated code; ts updates

chalin:
Note that the Dart source is still at beta.15 (due to https://github.com/angular/angular/issues/8258). The Dart example has been updated to have the same functionality as the TS example.


Foxandxss:
TS changes works for me.

thso:
LGTM

PR #1355, opened 2016-05-11
docs(typings): add typings.d.ts for module.id and describe it

wardbell:
Supersedes PR #1312.

### DIFF
--- a/public/docs/_examples/.gitignore
+++ b/public/docs/_examples/.gitignore
@@ -20,7 +20,7 @@ _test-output
 _temp
 **/ts/**/*.js
 **/ts-snippets/**/*.js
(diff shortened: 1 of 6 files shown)


wardbell:
We are **TOTALLY HOSED** in plunkers for the moment because CLI and Style Guide are committed to the commonjs `module.id` and I can't find any transpile options that will make it work in plunker.

I tried the following and it doesn't work:
```
    transpiler: 'typescript',
    typescriptOptions: {
      emitDecoratorMetadata: true, 
      module: 'commonjs' // should have done the trick!
    },
```

wardbell:
We are looking into reverting from commonjs modules to systemjs modules at which time this problem goes away.

PR #1358, opened 2016-05-11
feat(dgeni audit): add more keys to the api-list

ericjim:
#### Changes
* Add more keys to the api-list, this exposes the status of all api elements.
* The output is located in `ts/api-list.json` and `js/api-list.json`, this only happens when a new build is made.

#### Notes
`api-list.json` is the file that's used to create the list located at https://angular.io/docs/ts/latest/api/ (and in javascript). Since I've only made additions (and not changed previous key-value pairs), the api doc should remain unaffected. Specifically `api-list.js` will use the key-value pairs to generate the list for the api doc.

@naomiblack @petebacondarwin


naomiblack:
merged, live!

PR #1366, opened 2016-05-12
docs(style-guide): add the image width attribute

not-for-me:
Adding the width attribute to the gif image of visual studio extension for supporting mobile phones.


wardbell:
Why? Why is it a problem for this image and not every image in the docs?

not-for-me:
@wardbell because every other image's width is smaller than mobile screen size, but the size of this image is bigger than 700px.
I think that the "%" width attribute of it for supporting mobile is need.

Thanks.

johnpapa:
LGTM

wardbell:
Thanks. Merged.

Issue #1331, opened 2016-05-09
Vertical scrollbar in single line example code


bampakoa:
A vertical scrollbar appears in some example code that is single line.

![image](https://cloud.githubusercontent.com/assets/7170892/15111361/361fafcc-15ef-11e6-9fc0-bf5fc52fc478.png)

ericjim:
![screen shot 2016-05-11 at 1 10 12 pm](https://cloud.githubusercontent.com/assets/1329167/15194923/69d6dcdc-1793-11e6-908f-7f1b1f7ec9bf.png)

PR #1370, opened 2016-05-12
fix(copy btn): remove overflow from copy button

ericjim:
#### Changes
* Refactor inline styles from directive to a SASS file
* Remove overflow from copy button which was causing scroll bars to appear
* Adjust button to the top right of the `.copy-container-template`

#### Preview
* Live: https://copy-btn-no-scroll.firebaseapp.com/docs/ts/latest/quickstart.html

CLOSES:
* https://github.com/angular/angular.io/issues/1331

@naomiblack


naomiblack:
merged. live!

PR #1378, opened 2016-05-13
docs(router): Fix broken router test

jmcooper:
The tests were negatively impacting each other. Changed beforeAll to beforeEach

### DIFF
--- a/public/docs/_examples/cb-barrels/e2e-spec.js
+++ b/public/docs/_examples/cb-barrels/e2e-spec.js
@@ -1,0 +1,110 @@
+describe('Tutorial', function () {
+
+  beforeAll(function () {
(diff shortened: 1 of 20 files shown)


wardbell:
Jim - I realize that this step is safest but it really slows things down. Would you please reconsider and group as many of the tests as you can under a single `getAll` browser load?

jmcooper:
I accidentally opened this against my master branch on my fork.  I'll re-open it with the changes you suggested Ward.

jmcooper:
Actually, Ward, I decided to look deeper into the reason for failure and the test is actually exposing a problem with the app.  You can see the problem by following these steps:running the router example, clicking on a crisis, changing the name and saving

PR #1383, opened 2016-05-13
docs(style-guide): avoid I prefix for interface naming convention

TTTee:
The `style-guide` first points out that you should consider to not use an I prefix for interfaces and 2 lines later says you should avoid it. This should be consistent. (Also 'Consider' has a blue mark on the left side instead of a green one). 

image

### DIFF
--- a/public/docs/ts/latest/guide/style-guide.jade
+++ b/public/docs/ts/latest/guide/style-guide.jade
@@ -688,9 +688,9 @@ a(href="#toc") Back to top
   :marked
     **Do** name an interface using upper camel case.
 
-.s-rule.do
+.s-rule.avoid
   :marked
-    **Consider** naming an interface without an `I` prefix.
+    **Avoid** prefixing interfaces with an `I`.
 
 .s-why.s-why-last
   :marked


wardbell:
I get what you mean. We're trying to make a **soft** recommendation to steer clear of the `i` prefix but we don't have a corresponding code representation. 

A greater objection, in my mind, is that the "Why" makes no sense in this context. 

>That "Why" belongs to a different consideration: _consider using a class instead of an interface_, a recommendation that is strictly situational and difficult to explain.

TTTee:
I see, why this is only a consideration. Indeed the 'why' is a bit misleading. 
Your recommendation is to add the additional consideration you provided and keep the two code blocks below? Also one could add a code block showing an interface declaration without an `I` prefix. Should I create a new pull request/ change this one?

Besides the css-class is not correct:
https://github.com/angular/angular.io/blob/master/public/docs/ts/latest/guide/style-guide.jade#L691
```jade
.s-rule.do
  :marked
    **Consider** naming an interface without an `I` prefix.
```

johnpapa:
I see the point, but I still think it is valid. It just needs more description substance.

Using a class is a valid alternative. It is related to this and if you truly can use a class instead of an interface, then you certainly don;t even need the I.

We have reported the consider needs to be green, not blue. I think @Foxandxss is on this.

johnpapa:
I propose this be a `Consider`, not an avoid. If someone wants to use `I` as a prefix, the world won't end. :)

for this i feel we close it. @wardbell

TTTee:
@johnpapa `I` agree (pun intended). Can be closed.

PR #1390, opened 2016-05-13
docs(style-guide): push main.ts to src folder

### LINKED ISSUE #1361: StyleGuide - Having the root /app folder export a barrel
The CLI has a barrel in `/src/app` where they export the app.component.ts and some environment. Also the `main.ts` is at `/src` and not `/src/app`.

So we have to move `main.ts` up one level and then see if we want the `app` folder to export a barrel and if so, what should it export, only app.component or each subfolder (features and shared).

Foxandxss:
/cc @johnpapa  @wardbell  @Brocco @mgechev

We agreed on pushing `main.ts` to the src folder so here it is. We also talked about adding a barrel in `/src/app/` with a single export, but that is a no-no for @wardbell.

So with this PR and Ward's opinion, we can close the next issue:

Closes #1361

### DIFF
--- a/public/docs/ts/latest/guide/style-guide.jade
+++ b/public/docs/ts/latest/guide/style-guide.jade
@@ -965,7 +965,7 @@ a(href="#toc") Back to top
       .children
         .file ...
       .file app.component.ts|html|css|spec.ts
-      .file main.ts
+    .file main.ts
     .file index.html
     .file ...
 :marked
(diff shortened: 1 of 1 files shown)


johnpapa:
Cool. Has CLI adjusted too?

Foxandxss:
CLI was always having main.ts in `src/`.

About the barrel, they export more than one thing, so that is fine for now I guess.

mgechev:
LGTM

wardbell:
The move is OK with me. As @Foxandxss says, I am intolerant of a barrel at the root with only `AppComponent` in it.

PR #1391, opened 2016-05-13
feat(resources db): feed resources page from firebase

ericjim:
#### Changes
* Remove unused resource pages
* Remove resource page which was static jade
* Template resource page using ng1 and firebase

#### Preview
Live: https://resources-db.firebaseapp.com/resources

#### Notes
I am using the raw data from https://angularresources.firebaseio.com/, the best way to change the entries would be to modify the values in the database.

Also, it would be great if we could prioritize a new design 😉 

@naomiblack @juleskremer


public/resources/index.ejs
```diff
@@ -0,0 +1,19 @@
...
+<div class="js-resources-ctrl resources" ng-controller="ResourcesCtrl as ctrl">
+    <div class="section" ng-repeat="(category, categoryContent) in ctrl.fbObject">
+        <div>
+            <h1 class="capitalize">{{category}}</h1>
+            <div ng-repeat="section in categoryContent">
+                <div class="grid-fluid" ng-repeat="(subsectionName, subsectionArray) in section">
+                    <div>
+                        <h2 class="capitalize">{{ subsectionName }}</h2>
+                        <ul ng-repeat="resource in subsectionArray.resources">
+                            <li ng-if="resource.rev"><a target="_blank" href="{{resource.url}}">{{ resource.title }}</a></li>
```
ericjim: @juleskremer I am making an explicit check for approved entries of the resource page.

ericjim:
@juleskremer it would be great if we could add an entry in firebase which hasn't been approve; and leave it there for testing purposes. Just as a way of doing integration testing.

naomiblack:
live and in master.

PR #1405, opened 2016-05-16
feat(badging): Enum only matches enums, const matches const, var and let

ericjim:
#### Changes:
* Everything dgeni previously labeled const or enum should display as enum
* Everything that dgeni was outputting as var/let should display as const

#### Preview
* https://rebadging-refilter.firebaseapp.com/docs/ts/latest/api/

##### Var, Let ---> Const
* (Before) https://angular.io/docs/ts/latest/api/#!?apiType=Variable
* (After) https://rebadging-refilter.firebaseapp.com/docs/ts/latest/api/#!?apiType=Const

##### Const and Enum ---> Enum
* (Before) https://angular.io/docs/ts/latest/api/#!?apiType=Const%20or%20Enum
* (After) https://rebadging-refilter.firebaseapp.com/docs/ts/latest/api/#!?apiType=Enum

### DIFF
--- a/public/_includes/_hero.jade
+++ b/public/_includes/_hero.jade
@@ -4,6 +4,12 @@
 - var capitalize = function capitalize(str) { return str.charAt(0).toUpperCase() + str.slice(1); }
 - var useBadges = docType || stability;
 
+// renamer :: String -> String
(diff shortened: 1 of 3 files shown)


naomiblack:
live and in master.

PR #1407, opened 2016-05-16
docs(glossary): update the example paths for the glossary

makearl:
The previous paths caused this error on the glossary page:
```
BAD FILENAME: ../../_fragments/../docs/_fragments/quickstart/ts/app/app.component-import.ts.md   Current path: docs,ts,latest,glossary PathToDocs: ../../
```

### DIFF
--- a/public/docs/ts/latest/glossary.jade
+++ b/public/docs/ts/latest/glossary.jade
@@ -84,7 +84,7 @@ include _util-fns
 // #enddocregion b-c
 :marked
   That's why we can write this:
-+makeExample('../docs/_fragments/quickstart/ts/app/app.component.ts', 'import')(format=".")
++makeExample('quickstart/ts/app/app.component.ts', 'import')(format=".")
 // #docregion b-c
 
 :marked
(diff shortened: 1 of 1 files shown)


makearl:
I signed it!

wardbell:
Sorry but this is a no-win situation because of the double location of the file. There is not actually a problem when you navigate to the glossary [from the Basics link](https://angular.io/docs/ts/latest/guide/glossary.html).  You can only create the problem if you [navigate to this](https://angular.io/docs/ts/latest/glossary.html) and I believe that no link on the site does that.  Not quite sure how you stumbled into it.

We're going to fix it with a Firebase re-write.

makearl:
Good catch, I didn't realize there was another way to get there. I stumbled on this by clicking a reference to `barrel` in the [Architecture overview](https://angular.io/docs/ts/latest/guide/architecture.html) which brought me to this link: https://angular.io/docs/ts/latest/glossary.html#!#barrel

Maybe the correct change is updating the link in the Architecture overview?

PR #1410, opened 2016-05-16
docs(dart): update BASICS intro

kwalrath:
Also copy edited the TS version a bit.

NOTE: The TS version (and thus the Dart version) talks about "Reference" and "Resources" sections that don't exist. (API Reference does, but not Reference, as such.)

### DIFF
--- a/public/docs/dart/latest/guide/index.jade
+++ b/public/docs/dart/latest/guide/index.jade
@@ -1,93 +1,18 @@
-include ../_util-fns
+extends ../../../ts/latest/guide/index.jade
 
-:marked
-  This Developers Guide is a practical guide to Angular for experienced programmers who
-  are building client applications in HTML and Dart.
-figure
-  img(src="/resources/images/devguide/intro/people.png" alt="Us" align="left" style="width:200px; margin-left:-40px;margin-right:10px" )
-:marked
-  We are on a journey together to understand how Angular works and, more importantly,
-  how to make it work for us. We look at our application requirements and we see problems to solve.
-  <br clear="all">
+block includes
+  include ../_util-fns
 
-  * How do we get data onto the screen and handle user interactions?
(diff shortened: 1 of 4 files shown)


kwalrath:
@wardbell do you want me to make the section descriptions match what's really there?

wardbell:
@kwalrath Please do :-)

public/docs/dart/latest/guide/index.jade
```diff
@@ -1,93 +1,18 @@
...
+block example-links
+  :marked
+    Look for a link to that sample near the top of each page.
+    For example, the sample repo https://github.com/angular-examples/architecture
+    contains the code
+    for the [Architecture](architecture.html) chapter's sample.
+    A running version of that sample is at
+    https://angular-examples.github.io/architecture/.
+
+    A few early chapters are written as tutorials and are clearly marked as such.
```
chalin: Why not include this paragraph (L15-18) in the TS version?

chalin:
Other than the one inlined comment made about moving a paragraph to the TS version, it LGTM.

public/docs/dart/latest/guide/index.jade
kwalrath: It's the "live link" vs. "full repo" difference. I could've made it a variable and probably should.

chalin: Ah, right. I just feel that we should minimize differences between the language versions. The added remark will be useful to TS and JS users too :). If you don't want to define a variable, you could just drop the last sentence (L18) and put L15-17 in the TS version. It's up to you.

kwalrath: It actually _is_ in the TS & JS pages, but they have an all-plunkr solution where they can change the example. So the text needs to be different.

kwalrath:
About to merge...

PR #1419, opened 2016-05-17
docs(cb-barrels): Add barrels cookbook recipe

jmcooper:
Add the cookbook recipe demonstrating how to create barrels to simplify module imports.


Foxandxss:
I like it, but I would make it follow the style-guide. That means having `/heroes/heroes.component.ts|html|css`, `/heroes/hero-detail/hero-detail.component.ts|html|css`. Service renamed as `hero.service.ts`, model renamed to `hero.model.ts` and both put into a `shared` folder. That way you will have more barrels too so the cookbook would be more interesting.

Leaving that aside, I have no problem with the prose :)

Foxandxss:
Oh, final call here is from @wardbell :)

jim-cooper:
I like your suggestions @Foxandxss.  I'll make those changes.

jim-cooper:
There is one problem I just realized. Adding the SystemJs package in the index.html works great for running the example, locally, but it doesn't work on plunker. Unfortunately, the main needs to be index.js locally, but index.ts (notice the js vs the ts) on plunker.  Not sure the best way to solve that.  Seems like something the plunker builder would have to handle.

Foxandxss:
I see what you mean. I will give it a thought.

jim-cooper:
@Foxandxss I've made the edits you recommended and I think it's much better now, thanks for the suggestions. Let me know if you have any thoughts on the plunker issue.

public/docs/ts/latest/cookbook/barrels.jade
```diff
@@ -0,0 +1,97 @@
...
+    .file heroes.component.ts
+    .file hero-detail
+    .children
+      .file hero-detail.component.css
+      .file hero-detail.component.html
+      .file hero-detail.component.ts
+    .file shared
+    .children
+      .file hero.service.ts
+      .file hero.ts
```
Foxandxss: typo.

Foxandxss:
The hero-list is also a barrel, so it needs to export too :)

Foxandxss:
Could you also please fix the indentation? You are mixing spaces with tabs. We use 2 spaces.

jmcooper:
Not sure what you mean by the hero-list component. If you mean the heros.component, it is being exported in the barrel.  Can you clarify?  :)

Foxandxss:
If we see the style-guide, each folder is a barrel. That means that `app/heroes` exports a barrel, `app/heroes/hero-detail/` exports a barrel, `app/heroes/shared` another barrel, etc.

jmcooper:
Ahh, I see what you're saying.  You said `hero-list` is a barrel, I think you meant `hero-detail`. But I wasn't under the impression from the "whys" listed in the style guide that we were suggesting creating barrels for a single import, but rather to aggregate multiple imports.  `hero-detail` is being aggregated at a higher level with the other imports, should we really create an `index` file for a single import?

Foxandxss:
We are promoting that on the style guide, but I will double check first.

Foxandxss:
After a conversation, yes, we add that even if it is one export.

Foxandxss:
Plunker fix is on the way. Could you please update with the barrel stuff? :)

jmcooper:
Ok, I've added the barrel for hero-detail and made the other changes. 

However, it looks like the Plunker change isn't quite right.  It appears to be only updating the first package in the list from .js to .ts and leaving the others. Once that's fixed I'll rebase again to grab it. 

Btw, sorry for the "Merge branch..." commit, accidentally did a pull instead of a rebase on that last one.

jmcooper:
Also, not sure why the CLA changed to no. It's because of the commits from other authors that I pulled in. Was I not supposed to rebase those changes into my branch?

PascalPrecht:
@jmcooper you've pushed commits from other ppl.

You should rebase your changes so that they don't appear here.

wardbell:
Jim ... I'll look soon. No you're not supposed to merge which is why it looks like you changed 177 files.

But we understand. Git is PITA. We'll get you straightened out.

wardbell:
Closing because (a) it's in Jim's fork and I can't help, (b) it's a tangle of unrelated commits, and (c) I think I was able to straighten this out in an IdeaBlade fork of angular.io.

I believe the proper contents are in PR #1458

PR #1420, opened 2016-05-17
chore(docs): revise makeExcerpt mixin for DRYer ref to examples

chalin:
Avoiding `ts` in the file path will make it a bit easier to construct the path to the Dart version (by simply appending to the path rather than having to massage the path).

This update allows us to be more DRY when referencing examples and excerpts. E.g.,
```jade
+makeExcerpt('quickstart/ts/app/app.component.ts', 'import', 'app/app.component.ts (import)')
```
can now be just
```jade
+makeExcerpt('app/app.component.ts', 'import')
```

Also defined new mixin named `makeExample1`, for examples, using this new argument scheme. The original `makeExample` has been left untouched.

Applied new mixins to quickstart.

### DIFF
--- a/public/_includes/_util-fns.jade
+++ b/public/_includes/_util-fns.jade
@@ -68,13 +68,43 @@ mixin makeExample(_filePath, region, _title, stylePatterns)
       .example-title #{title}
   code-example(language="#{language}" format="#{format}")
     != styleString(frag, stylePatterns)
-    
-//- Like makeExample, but doesn't show line numbers and
-//- title is appened with `(excerpt)` if it doesn't already
-//- end with a parenthetical remark.
-mixin makeExcerpt(_filePath, region, _title, stylePatterns)
-  - if (_title && !_title.match(/\([\w ]+\)$/)) _title = _title + ' (excerpt)';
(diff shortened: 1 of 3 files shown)


chalin:
@kwalrath can you have a look?

kwalrath:
LGTM. @wardbell do you have an opinion? I wonder if the name, makeExample1, could be clearer. makeExample2? makeSimpleExample?

Foxandxss:
Or perhaps `makeLocalExample`? I personally don't like `example1`, `example2` kind of names because you can easily forget which is which.

wardbell:
I like but the name has to change.  `makeLocalExample` isn't bad. Or `makeProjectExample`.  Either works for me ... anything except `example1`.

Happy to accept after that change.

chalin:
I agree about the name, but figured I run by the basic idea first. 

I originally thought of: `makeEg` as a short name (to represent that is takes the short argument list).
Otherwise, `makeProjExample`? (I'm trying to keep the name reasonably short.)

chalin:
Then `makeProjExample` it is.
@kwalrath @wardbell : ready to merge.

kwalrath:
OK, I'll merge a little later this morning. Thanks, Patrice!

PR #1422, opened 2016-05-18
docs(DI articles - showing providers using object literals)

thelgevold:
@wardbell I have extended both the DI cookbook and the dev guide to include the new object literal based syntax for defining providers. 

Is this what you had in mind?

The cookbook was definitely easier to update since all the variations on ```provide(..)``` are lumped together in the same component. This made it possible to show both versions in all cases with little effort.

In the dev guide the components are much more granular, so I think it would sacrifice code readability if we go through and show all the samples using both flavors. Instead I focused on explaining the new syntax as a yet another alternative (there are three alternatives now).  

cc: @PascalPrecht


wardbell:
Made minor tweaks. Good to go!

PR #1439, opened 2016-05-19
chore: translate inline system config for plunker

### LINKED ISSUE #1424: Custom system.js config not working for Plunker
**Problem:**

If we need to define custom barrels in our examples, we will need to extend the `systemjs.config.js` (via another file or inline in the `index.html`).

That won't work on Plunker because we load `.js` locally and `.ts` in plunker.

**Solutions:**

We could either offer a different `index.html` for plunker (if we inline the config). That is easy to do, doesn't need changes.

We could have a `systemjs.custom.js` for this new custom stuff and offer a `systemjs.custom.plunker.js` that plunker would use. Downsides, we will need to write the same thing twice.

A third solution would be generating the  `systemjs.custom.plunker.js` on the fly by replacing the extensions at `systemjs.custom.js`. It could be potentially brittle, but shouldn't give any troubles.

---

I have no problem in fixing this, but first I want to see your ideas or see if I am missing an obvious solution.

/cc @wardbell

Foxandxss:
Put it in two separate options so you can translate one, both or none in no specific order.

Fixes #1424

### DIFF
--- a/tools/plunker-builder/indexHtmlTranslator.js
+++ b/tools/plunker-builder/indexHtmlTranslator.js
@@ -18,6 +18,14 @@ var _rxRules = {
   link: {
     from: '/<link rel="stylesheet" href=".*%tag%".*>/',
     to: '<link rel="stylesheet" href="%tag%">'
+  },
+  system_extra_main: {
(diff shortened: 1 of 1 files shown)


wardbell:
Excellent! Thanks

PR #1450, opened 2016-05-20
docs(style-guide): add testing harness

Foxandxss:
Testing harness for the style guide (no actual tests just yet).

The philosophy in here was to have each style to "output" an `app.component.ts` file that will be used to "route" to that style.

Here, each route renders one style.

The style pages are simply "printing out" what the style code does. So there are a lot of pages that only prints `...` because that is what the style's component has as a template.

I had to modify some style codes to at least print "something", but those changes won't appear on the style guide. Also added some extra code to make the examples work (some `moduleId`, fix some template / style urls...)

So as TL;DR, made lot of changes, but they don't have impact on the guide itself (except if I fixed typos). Each style page prints "something" and that is enough. On the test we test that we have that "something" on the screen. We don't need more, that will assert that our example codes compile successfully and are able to run and print something.

/cc @wardbell @johnpapa

### DIFF
--- a/public/docs/_examples/style-guide/ts/.gitignore
+++ b/public/docs/_examples/style-guide/ts/.gitignore
@@ -1,1 +1,2 @@
 *.js
+!systemjs.custom.js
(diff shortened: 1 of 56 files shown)


wardbell:
Is it ready to merge?

Foxandxss:
Yes. All examples (except routing, waiting for a new new router) show something on the screen that makes them testable.

Next step is test them / fix some problems with barrels to make the guide more accurate.

PR #1452, opened 2016-05-20
docs(style-guide): fix example on 02-04

### LINKED ISSUE #1254: StyleGuide - 02-04
Here we recommend to suffix services with `Service` when it is not clear what they are. Then we put `logger` as an example of what **not** to suffix but then we suffix it on the demo below.

Foxandxss:
It wasn't clear before.

Fixes #1254


johnpapa:
LGTM

PR #1461, opened 2016-05-21
docs: add recommended node and npm versions

Foxandxss:
Other versions (4.x and 6.x) are problematic as today so we will recommend those versions for the time being.


wardbell:
Merged by 81ebdfb

PR #1465, opened 2016-05-21
docs(template-syntax): review and update Dart to new doc design

### LINKED ISSUE #1460: guide/template-syntax: port to doc design scheme using Jade extends/blocks
Use Jade `extends` with `block`s rather than copying prose.

For the record, here are the differences in the HTML: [diff-ts-dart-html.txt](https://github.com/angular/angular.io/files/275856/diff-ts-dart-html.txt)

chalin:
Fixes #1460.

- Dart
  - update prose to new doc design.
  - enable e2e tests. Note: **tests pass**.
- TS: copyedits and changes to support Dart doc.

### DIFF
--- a/public/docs/_examples/template-syntax/dart/lib/app_component.html
+++ b/public/docs/_examples/template-syntax/dart/lib/app_component.html
@@ -31,11 +31,11 @@
 </div>
 <br>
 <a href="#star-prefix">* prefix and &lt;template&gt;</a><br>
-<a href="#local-vars">Template local variables</a><br>
+<a href="#ref-vars">Template reference variables</a><br>
 <a href="#inputs-and-outputs">Inputs and outputs</a><br>
 <a href="#pipes">Pipes</a><br>
 <a href="#safe-navigation-operator">Safe navigation operator <i>?.</i></a><br>
-<!--<a href="#enums">Enums</a><br>-->
+<a href="#enums">Enums</a><br>
 
 <!-- Interpolation and expressions -->
 <hr><h2 id="interpolation">Interpolation</h2>
@@ -105,9 +105,9 @@
 <!-- #docregion event-binding-syntax-1 -->
 <button (click) = "onSave()">Save</button>
 <hero-detail (deleteRequest)="deleteHero()"></hero-detail>
-<div (myClick)="clickity=$event">click me</div>
+<div (myClick)="clicked=$event">click me</div>
 <!-- #enddocregion event-binding-syntax-1 -->
-{{clickity}}
+{{clicked}}
 <br><br>
 
 <div>
@@ -184,15 +184,14 @@ button</button>
 <!-- #enddocregion property-binding-5 -->
 
 <!-- #docregion property-binding-6 -->
-  <!-- BAD! HeroDetailComponent.hero expects a Hero object,
-  not the string "currentHero".
-
-  <hero-detail hero="currentHero"></hero-detail> -->
+<!-- ERROR: HeroDetailComponent.hero expects a
+     Hero object, not the string "currentHero" -->
 <!-- #enddocregion property-binding-6 -->
-
-<!--  In checked mode, uncommenting the hero-detail above causes this:
-    EXCEPTION: type 'String' is not a subtype of type 'Hero' of 'value'. -->
-
+<div *ngIf="false">
+<!-- #docregion property-binding-6 -->
+  <hero-detail hero="currentHero"></hero-detail>
+<!-- #enddocregion property-binding-6 -->
+</div>
 <!-- #docregion property-binding-7 -->
 <hero-detail prefix="You are my" [hero]="currentHero"></hero-detail>
 <!-- #enddocregion property-binding-7 -->
@@ -202,7 +201,7 @@ Interpolated: <img src="{{heroImageUrl}}"><br>
 Property bound: <img [src]="heroImageUrl">
 
 <div>The interpolated title is {{title}}</div>
-<div [textContent]="'The [textContent] title is '+title"></div>
+<div [innerHTML]="'The [innerHTML] title is '+title"></div>
 <!-- #enddocregion property-binding-vs-interpolation -->
 
 <a class="to-toc" href="#toc">top</a>
@@ -258,9 +257,7 @@ Property bound: <img [src]="heroImageUrl">
 <!-- #docregion class-binding-2 -->
 <!-- reset/override all class names with a binding  -->
 <div class="bad curly special"
-     [class]="badCurly">Bad curly</div>
-<p><b>Note:</b> "Bad curly" should be smaller but isn't, due to
-  <a href="http://github.com/angular/angular/issues/6901">issue #6901</a>.</p>
+     [className]="badCurly">Bad curly</div>
 <!-- #enddocregion class-binding-2 -->
 
 <!-- #docregion class-binding-3 -->
@@ -374,15 +371,15 @@ bindon-ngModel
 <br>
 <!-- #docregion NgModel-3 -->
 <input
-    [ngModel]="currentHero.firstName"
-    (ngModelChange)="currentHero.firstName=$event">
+  [ngModel]="currentHero.firstName"
+  (ngModelChange)="currentHero.firstName=$event">
 <!-- #enddocregion NgModel-3 -->
 (ngModelChange) = "...firstName=$event"
 <br>
 <!-- #docregion NgModel-4 -->
 <input
-    [ngModel]="currentHero.firstName"
-    (ngModelChange)="setUpperCaseFirstName($event)">
+  [ngModel]="currentHero.firstName"
+  (ngModelChange)="setUpperCaseFirstName($event)">
 <!-- #enddocregion NgModel-4 -->
 (ngModelChange) = "setUpperCaseFirstName($event)"
 <br>
@@ -443,7 +440,7 @@ bindon-ngModel
 <!-- #enddocregion NgIf-1 -->
 
 <!-- #docregion NgIf-2 -->
-<!-- not displayed because nullHero is false.
+<!-- because of the ngIf guard
     `nullHero.firstName` never has a chance to fail -->
 <div *ngIf="nullHero != null">Hello, {{nullHero.firstName}}</div>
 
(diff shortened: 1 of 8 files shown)


chalin:
@Foxandxss, @wardbell, @kwalrath, @thso : this is ready for review. As usual, it might be easiest to [review while ignoring differences in whitespace](https://github.com/angular/angular.io/pull/1465/files?w=1).

public/docs/_examples/template-syntax/dart/lib/app_component.html
```diff
@@ -487,33 +484,33 @@
...
+      <!-- #docregion NgSwitch -->
+      <span *ngSwitchWhen="'Eenie'">Eenie</span>
+      <span *ngSwitchWhen="'Meanie'">Meanie</span>
+      <span *ngSwitchWhen="'Miney'">Miney</span>
+      <span *ngSwitchWhen="'Moe'">Moe</span>
+      <span *ngSwitchDefault>other</span>
+      <!-- #enddocregion NgSwitch -->
+
+      <!-- with <template> -->
+      <template [ngSwitchWhen]="'Eenie'"><span>Eenie</span></template>
```
thso: Why this change?

thso:
LGTM for Dart code.

public/docs/_examples/template-syntax/dart/lib/app_component.html
chalin: The objective was to show the result of desugaring the `*ngSwitchWhen` **without** any _optimizations_. The chapter material is complex enough, so I preferred that the **reader not be forced to recall** that if the template expression is a string constant then:
  `<foo [x]="'s'">`
is equivalent to 
  `<foo x="s">`
Besides, this is what matches the TS example. Let me know if you agree with this reasoning.

Foxandxss:
The TS part is good for me.

chalin:
@kwalrath : any updates on a review of the Dart-side prose?

kwalrath:
Taking a look now...

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -1,8 +1,13 @@
...
+  - var _JavaScript = 'JavaScript';
+  //- Double underscore means don't escape var, use !{__var}.
+  - var __chaining_op = '<code>;</code> or <code>,</code>';
+  - var __new_op = '<code>new</code>';
+  - var __objectAsMap = 'object';
 
-// #docregion intro
 :marked
-  Our Angular application manages what the user sees and does through the interaction of a Component class instance (the *component*) and its user-facing template.
+  Our Angular application manages what the user sees and it achieves this through the interaction of a Component class instance (the *component*) and its user-facing template.
```
kwalrath: It's good to change the original sentence, since it was imprecise, but the new version is rather run-on. How about:

Our Angular application manages what the user sees and can do, achieving this through the...

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -107,39 +98,40 @@ include ../_util-fns
...
-// #enddocregion template-expressions-1
 
-// #docregion template-expressions-2
 :marked
-  We write template expressions in a language that looks like JavaScript.
-  Many JavaScript expressions are legal template expressions, but not all.
-  JavaScript expressions that have or promote side effects are prohibited,
+  We write template expressions in a language that looks like #{_JavaScript}.
+  Many #{_JavaScript} expressions are legal template expressions, but not all.
+  #{_JavaScript} expressions that have or promote side effects are prohibited,
```
kwalrath: Add a blank line above this, so people are less likely to miss the context for the list.

(It originally had this blank line, but sometime before your changes the blank line disappeared.)

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -194,16 +184,15 @@ include ../_util-fns
 
   In Angular terms, an idempotent expression always returns *exactly the same thing* until
   one of its dependent values changes.
-// #enddocregion template-expressions-guidelines
-// #docregion template-expressions-guidelines-2
 :marked
   Dependent values should not change during a single turn of the event loop.
   If an idempotent expression returns a string or a number, it returns the same string or number
-  when called twice in a row. If the expression returns an object (including a `Date` or `Array`),
+  when called twice in a row. If the expression returns an object (including #{_a} `#{_Array}`),
```
kwalrath: Why leave out `Date`?

If there's a good reason to leave it out, change `_a` to `_an`.

public/docs/ts/latest/guide/template-syntax.jade
chalin: done

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -738,22 +679,20 @@ code-example(format="", language="html").
...
 .l-sub-section
   :marked
     While this is a fine way to set a single style,
     we generally prefer the [NgStyle directive](#ngStyle) when setting several inline styles at the same time.
-// #enddocregion other-bindings-style-3
 
-// #docregion event-binding-1
+.l-sub-section
+  :marked
+    Note that the _style property_ name can be written in camcel case, such as `fontSize`.
```
kwalrath: camcel -> camel

Or maybe camel case -> camelCase? (What do we use elsewhere?)

public/docs/ts/latest/guide/template-syntax.jade
chalin: done

public/docs/ts/latest/guide/template-syntax.jade
chalin: Good catch re. the article. Fixed.

Yes, I removed `Date` because the class name is `DateTime` in Dart and didn't want to define a new conditional variable just for that. I think that the point of that phrase remains clear even without Date/DateTime mentions (actually, I'm not even sure that the parenthetical remark is needed _at all_, but it is certainly ok with just mentioning Array/List IMHO).

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -107,39 +98,40 @@ include ../_util-fns
...
+  Many #{_JavaScript} expressions are legal template expressions, but not all.
+  #{_JavaScript} expressions that have or promote side effects are prohibited,
   including:
 
-  * assignments (`=`, `+=`, `-=`)
-  * the `new` operator
-  * chaining expressions with `;` or `,`
+  * assignments (`=`, `+=`, `-=`, ...)
+  * !{__new_op}
+  * chaining expressions with !{__chaining_op}
```
kwalrath: I'm not sure this applies to Dart. (Maybe it does. But I think I left it out for a reason...)

public/docs/dart/latest/guide/template-syntax.jade
```diff
@@ -1,407 +1,125 @@
...
-    instead of `"'The title is $title'"`, you must use
-    `"'The title is ' + title"`
-  * no support for the bitwise operators `|` and `&`
-  * new [template expression operators](#expression-operators), such as `|`
-+includeShared('{ts}', 'template-expressions-context')
-+includeShared('{ts}', 'template-expressions-guidelines')
-:marked
-  Dependent values should not change during a single turn of the event loop.
-  If an idempotent expression returns a string or a number, it returns the same string or number
-  when called twice in a row. If the expression returns an object (including a `DateTime`, `Map`, or `List`),
```
kwalrath: Why leave out DateTime & Map?

public/docs/dart/latest/guide/template-syntax.jade
```diff
@@ -1,407 +1,125 @@
...
-  header Dart difference: No truthy values
+    Perhaps more surprising, template expressions can’t refer to static
+    properties, nor to top-level variables or functions, such as `window` or
+    `document` from `dart:html`. They can’t directly call `print` or functions
+    imported from `dart:math`. They are restricted to referencing members of
+    the expression context.
+
+block array-is-list-in-Dart
+  .callout.is-helpful
+    header Dart difference: Arrays are lists
```
kwalrath: Or we could just use "list" everywhere, since that's easy now. We'd still want to call out that difference somewhere (perhaps in this chapter), but maybe we don't need to say "array" anywhere else in the Dart docs.

public/docs/ts/latest/guide/template-syntax.jade
kwalrath: __chaining_op is inconsistent between the TS & Dart versions. TS says the full name with the op in parentheses, while Dart just says the op name.

public/docs/ts/latest/guide/template-syntax.jade
chalin: Both are used elsewhere, about equally, in the docs. I'll use "camelCase". Let me know if you have objections.

public/docs/dart/latest/guide/template-syntax.jade
```diff
@@ -1,407 +1,125 @@
...
-+includeShared('{ts}', 'property-binding-8')
-+makeExample('template-syntax/dart/lib/app_component.html', 'property-binding-5')(format=".")
-+includeShared('{ts}', 'property-binding-9')
-+makeExample('template-syntax/dart/lib/app_component.html', 'property-binding-3')(format=".")
-+includeShared('{ts}', 'property-binding-10')
-+makeExample('template-syntax/dart/lib/app_component.html', 'property-binding-4')(format=".")
-+includeShared('{ts}', 'property-binding-11')
-+makeExample('template-syntax/dart/lib/app_component.html', 'property-binding-6')(format=".")
-.callout.is-helpful
-  header Dart difference: Type exceptions
```
kwalrath: Did you intentionally leave this out?

kwalrath: Oops, sorry, it's in there.

public/docs/dart/latest/guide/template-syntax.jade
chalin: Actually, come to think of it, _everything_ is an object in Dart, so the parenthetical remark is actually potentially confusing for Dartisans.

public/docs/dart/latest/guide/template-syntax.jade
```diff
@@ -1,407 +1,125 @@
-include ../_util-fns
+extends ../../../ts/latest/guide/template-syntax.jade
 
-+includeShared('{ts}', 'intro')
+block includes
+  include ../_util-fns
+  - var _JavaScript = 'Dart';
+  - var __chaining_op = '<code>;</code>';
+  - var __new_op = '<code>new</code> or <code>const</code>';
+  - var mapApiRef = 'https://api.dartlang.org/stable/1.16.0/dart-core/Map-class.html';
```
kwalrath: I guess this is what we need for now, but once https://github.com/dart-lang/api.dartlang.org/issues/62 is fixed let's update this link!

kwalrath:
LGTM, once you fix the minor problems I noted.

public/docs/ts/latest/guide/template-syntax.jade
chalin: Looking into this a bit further ... interestingly, it _is_ possible to chain expressions with side-effects in Dart, using _both_ `;` **and** `,`.

- Using `;` comes from the fact that an expression is a statement -- see [Line 577 in the Dart 1.10 grammar](https://github.com/chalin/dart-spec-and-grammar/blob/master/doc/grammar-AUTOGENERATED-DO-NOT-EDIT.txt#L577).
- Using `,` is possible in the context of `for` loops; see the for-loop production rule on [Line 593](https://github.com/chalin/dart-spec-and-grammar/blob/master/doc/grammar-AUTOGENERATED-DO-NOT-EDIT.txt#L593). The [`expressionList`](https://github.com/chalin/dart-spec-and-grammar/blob/master/doc/grammar-AUTOGENERATED-DO-NOT-EDIT.txt#L199) at the end is a comma-separated list of expressions.

But, comma-separated lists of expressions is not permitted as a statement (in contrast to other C-like languages), so I believe that the prose at line 110 that you refer to above is a good enough short qualification of the situation. Do you agree?

public/docs/dart/latest/guide/template-syntax.jade
chalin: > Or we could just use "list" everywhere

Agreed.

> We'd still want to call out that difference somewhere (perhaps in this chapter)

Agreed. I don't think that it should be nestled in a central chapter of the basic dev. guide. So I'll just cut the callout out and save it for inclusion elsewhere.

public/docs/dart/latest/guide/template-syntax.jade
chalin: Yes, thanks for the reference to https://github.com/dart-lang/api.dartlang.org/issues/62.

chalin:
I believe that I have addressed all of your concerns. (See my replies to specific line comments; in particular about expression chaining.)

public/docs/ts/latest/guide/template-syntax.jade
chalin: Just a follow-up on this. There is actually a glossary entry for "camelCase", so I think that that is the better choice.

chalin:
@kwalrath : I pushed a new separate commit (adding a Dart difference call out), now that @vicb can confirmed the recommended style property naming scheme.

chalin:
@kwalrath : can this be merged, if not what can I do to help make that happen? (I'd like to get the changes to `_util-fns.jade` for work on other chapter).

public/docs/ts/latest/guide/template-syntax.jade
kwalrath: An expression plus a `;` is a statement, to be precise. The reverse isn't true.

I don't think you could legitimately have `,` chaining in any context other than a `for` loop, so I'm excluding it from consideration. :)

But niggling precision aside... I think my comment was referring to the 4th line, which doesn't exist in the online version.

kwalrath: Actually, it _did_ refer to the 3rd line. For some reason, I didn't have "chaining expressions with ;" in the old version of the Dart TS page.

PR #1469, opened 2016-05-22
docs(style-guide): add e2e tests

Foxandxss:
Tests!

I had to change the router to use Hash based routing (so protractor won't complain) but that is not a biggie.

Thanks to the new improved e2e task and this tests, we can be sure that all the style compiles and what they show on the screen is correct.


wardbell:
Why did protracted complain? Because http-server failed?

Foxandxss:
Yes, not a html 5 mode lover

PR #1478, opened 2016-05-23
docs(toh): add instructions to add the goBack method

### LINKED ISSUE #1476: ToH Ch7:HTTP - missing instructions to edit method goBack() in hero-detail.component.ts
At the end of Ch6, the goBack method in _app/hero-detail.component.ts_ looks like:
```
goBack() {
      window.history.back();
}
```

Instructions should be added to modify this method to:
```
goBack(savedHero: Hero = null) {
    this.close.emit(savedHero);
    if (this.navigated) { window.history.back(); }
  }
```

so that it can be used as shown in the save method:
`this.goBack(hero);`

Foxandxss:
The guide was never explaining the new changes to `goBack` and the given explanation was a bit confusing. It talks of a previous usage of "save" and there is none.

Fixes #1476


johnpapa:
LGTM

PR #1480, opened 2016-05-23
docs(style-guide): move IHero to hero.model

Foxandxss:
The avoid example was showing an import like:

```
import { IHero } from './hero.model.avoid';
```

When possible, I prefer the avoid example to not reference .avoid files.


public/docs/_examples/style-guide/ts/03-03/app/shared/hero.model.ts
```diff
@@ -1,7 +1,9 @@
-// #docregion
-// #docregion example
 export class Hero {
   name: string;
   power: string;
 }
-// #enddocregion example
+
```
johnpapa: If we are going to have an interface for the class, I suggest it should be listed above the class and the class should implement it

PR #1481, opened 2016-05-23
docs(style-guide): a service doesn't have a folder

Foxandxss:
A service doesn't need a folder.


johnpapa:
Right. I think this came from my other examples which has a toast component and a toast service together.

LGTM

PR #1485, opened 2016-05-23
docs(style-guide): use input at 05-12 for consistency

Foxandxss:
05-13 does this too so I updated this one for consistency.

We may want to use the output too, but that involves more code.


johnpapa:
LGTM

PR #1486, opened 2016-05-23
docs(style-guide): rename toast selector

Foxandxss:
The same file was using a different selector to grab the toast so for consistency, I changed it.


johnpapa:
LGTM

PR #1492, opened 2016-05-23
chore: put webpack guide in a ordered position

Foxandxss:
I never realized we had the guides in a ordered position. This fixes it.


wardbell:
I don't think it should be sorted under introduction. So I kept it in position and re-titled: "Webpack: an introduction".

PR #1498, opened 2016-05-24
docs(style-guide): make style guide code compliant with itself

Foxandxss:
Well, I added barrels, barrels everywhere.

I treat each style folder (`01-01`, `02-08`) as if they were the `src` folder.

I haven't add an `index.ts` to export the `app.component.ts`. @wardbell is not too comfy with the idea and I am not sure if @filipesilva convinced him or not. So I leave that hanging to see Ward's opinion.

I am exporting all folders, even if they have no files. That is an special case for the style guide, in the real world all folders would have code.


wardbell:
No I am not happy. I am not happy with empty barrels. I am not happy with style guidance that recommends them.

This is nuts. Can we please stop?

Attn: @johnpapa, @foxandxss, @filipesilva

johnpapa:
We shouldn't have empty barrels. 
I can't look right now but I think we miscommunicated somewhere. 

John Papa




On Tue, May 24, 2016 at 8:36 AM -0700, "Ward Bell" <notifications@github.com> wrote:












No I am not happy. I am not happy with empty barrels. I am not happy with style guidance that recommends them.



This is nuts. Can we please stop?



Attn: @johnpapa, @foxandxss, @filipesilva

johnpapa:
I do believe the CLI should create a barrel at the root for exporting `app.component.ts` and `environment.ts`. A barrel at that level should be *consider* in the style guide.

I do believe all folders that have any content (a folder with content or direct content) should have a barrel (e.g. an `index.ts` that exports that content). This should be in the style guide as a *do*. This makes it significantly easier to import things. The one time setup of a barrel (especially via CLI) is so tiny and similar to making fully spelled file names, and that minor cost is well worth the easier and shorter imports.

Empty folders with only an `index.ts` with no exports *should not exist*. That is a *Avoid*

A folder with only a folder in it (and that child folder has content) is not an *Empty* folder.

wardbell:
Where are we on this? Is it ready to be accepted? @johnpapa ?  @Foxandxss ?

Foxandxss:
I broke this, yay.

Foxandxss:
Lets continue at #1578

PR #1509, opened 2016-05-25
Fix typos

vikashvverma:
There are so many typos.
This PR fixes that.


Foxandxss:
Thank you. Big problem here is that we get no grammar correction on .jade files :/

wardbell:
Well Done Sir!  Thank you much. Gratefully accepted.

PR #1511, opened 2016-05-25
[SPIKE] docs: add real quickstart example DO NOT MERGE

Foxandxss:
**DO NOT MERGE, THIS IS ONLY A DEMO TO SEE HOW THIS IDEA WOULD LOOKS LIKE**

Don't mind the prose, it is practically copy & paste from the quickstart. The idea in here is to see how a quickstart tutorial with embedded plunkers look like.

### DIFF
--- a/public/docs/_examples/real-quickstart/ts/.gitignore
+++ b/public/docs/_examples/real-quickstart/ts/.gitignore
@@ -1,0 +1,1 @@
+!systemjs.config.1.js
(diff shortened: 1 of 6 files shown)


wardbell:
I don't think it should be sorted under introduction. So I kept it in position and re-titled: "Webpack: an introduction".

Foxandxss:
Wrong PR :P

wardbell:
Oops.

Foxandxss:
Closing this. We can still check it closed and we can do a better demo with the new tooling too.

PR #1518, opened 2016-05-25
Remove redundant double <br>

rexebin:


wardbell:
Thanks but they aren't redundant. They provide the necessary vertical spacing of the bullets.

rexebin:
In the process of localisation, I find that removing them resulting the same space, so I thought they can be there but serve no purpose.

PR #1534, opened 2016-05-26
chore: move from typings to @types

Foxandxss:
**AS WIP AS IT CAN GET**

The difference in here is that we need to list all the new @types directly into tsconfig.json. I want to believe that the need is going to be removed for Typescript 2.0.

There is much more to do yet, stay tuned.


Foxandxss:
This was just a proof of concepts, we will wait for TS 2, so bye bye.

PR #1536, opened 2016-05-27
docs: Update architecture.jade link on 'Testing' page

raucher:
Old one was pointed to [https://angular.io/docs/ts/latest/testing/index.html](url), this page doesn't have info about each step, only lists them.
New page [https://angular.io/docs/ts/latest/guide/testing.html](url) has the same information + description for almost each step. So I think link should point to latter.


wardbell:
Thanks. Added a redirect for anyone who goes to `~/docs/ts/latest/testing`.

PR #1545, opened 2016-05-29
chore(tests): add TS support for e2e

filipesilva:
This PR enables angular.io's e2e test suite to be written in Typescript, and updates all `e2e-spec.js` to `e2e-spec.ts`. It doesn't update their code though, but rather simply renames the extension.


public/docs/_examples/cb-dependency-injection/e2e-spec.ts
```diff
@@ -1,4 +1,4 @@
-describe('Dependency Injection Cookbook', function () {
+describe('Dependency Injection Cookbookaaa', function () {
```
Foxandxss: whaaat :P

filipesilva: my baaad :D

wardbell:
I'm excited to have this. But unfortunately,  it breaks VS Code intellisense for ​*everything*​, not just within the `e2e.spec.ts` files.  As soon as I roll back to master, intellisense returns.  Do you see that on your machine?

filipesilva:
Closed in favour of #1556

PR #1551, opened 2016-05-29
docs(toh-pt5): change getHero to match toh-pt6 getHero

cthorne01123:
Fixes a discrepancy between the `getHero` functions in the Tour of Heroes tutorial part 6 and 7.

Part 6 getHero: https://github.com/angular/angular.io/blob/master/public/docs/_examples/toh-5/ts/app/hero.service.ts#L21
Part 7 getHero: https://github.com/angular/angular.io/blob/master/public/docs/_examples/toh-6/ts/app/hero.service.ts#L34

Reasoning: [part 7 advises the reader to remove the mock `HEROES` data](https://github.com/angular/angular.io/blob/master/public/docs/_examples/toh-6/ts/app/hero.service.ts#L34) but does not tell the reader to update `getHero` accordingly, resulting in an error. Rather than requiring the reader to rewrite `getHero` to use `getHeroes` in part 7, it seems to make more sense to have them write it this way from the start.

### DIFF
--- a/public/docs/_examples/toh-5/ts/app/hero.service.ts
+++ b/public/docs/_examples/toh-5/ts/app/hero.service.ts
@@ -19,9 +19,8 @@ export class HeroService {
 
   //#docregion get-hero
   getHero(id: number) {
-    return Promise.resolve(HEROES).then(
-      heroes => heroes.filter(hero => hero.id === id)[0]
-    );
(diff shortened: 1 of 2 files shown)


wardbell:
Very nice. I agree. Merging.

PR #1553, opened 2016-05-30
docs(router): fix missing docregion

sarunint:


Foxandxss:
How did you reach the router page? I mean, that page is hidden, so I guess we have some document pointing at it which is wrong. 

Your PR is good, but that document is half done so missing regions are expected. We probably need to rewrite the chapter, so spending time on it, is wasting time. 

Thanks.

sarunint:
I reach it by googling "Angular 2 router" and it's the first item.

Good to know that it's still hidden.

wardbell:
Thanks.  I'm going to close w/o merging because that chapter is not even close to ready and shouldn't have been visible at all. Sorry to have troubled you.

PR #1557, opened 2016-05-30
chore: kill http-server on failed tests

Foxandxss:
Instead of making a check to see which OS are you using (needs an extra lib and extra code), I just kill it twice, better be sure it is dead.

No problem on Mac, but let see Windows.


Foxandxss:
Apparently the Linux on travis doesn't like it at all, so I will keep investigating.

Foxandxss:
Fixed it at last, thanks to Filipe.

Foxandxss:
Fixed at #1561

PR #1558, opened 2016-05-30
[WIP] docs(cookbook - key bindings) Adding key-binding cookbook.

bennadel:
This cookbook demonstrates how to use the key-combination support provided by
Angular's `KeyEventsPlugin`. This allows developers to tap into element-level
events like `keydown.Escape` and `keydown.Meta.Enter`.


bennadel:
@wardbell I removed the `WIP` notation from the title (re: our meeting on Monday).

Foxandxss:
e2e won't work, I think it is due our issue with protractor & chrome  (sendKeys).

PR #1556, opened 2016-05-30
chore: update gulpfile and project to add a tsconfig to protractor test folders

wardbell:
Replaces #1545

PR #1560, opened 2016-05-31
Protractor tsconfig jesus

Foxandxss:
This is my proposed solution for #1556 

Basically it is Ward's idea but instead of ts-node to compile the test, I compile every test before hand (and only the filtered ones).

Works perfect.

I would need to remove the need of a global typescript, but I will do that if we end merging this.


Foxandxss:
googlebot, those commits are ok to be contributed.

Foxandxss:
Added that feature because travis won't pass without that.

wardbell:
Superseded by PR #1561

PR #1561, opened 2016-05-31
Protractor tsconfig filipe

filipesilva:
This is my proposed solution for #1556. See also #1560.

It removes the boilerplate `protractor.config.js` from each folder and runs all protractor tests from the `_protractor/`, all using the same config and sending in the spec files via the `--specs=` argument to protractor.

It currently fails on some tests due due to `implicitAny` issues and undeclared variables, but those errors are on the test itself.


filipesilva:
These commits were authored by me and @wardbell, who is ok with his commits being on this PR.

wardbell:
Great job. I made one final change to capture the final suites passed/failed status report inside `protractor-results.txt`.  This picks up the test suites that _were missed_ because protractor couldn't compile them (the specs, not the app).

Many such un-compilable tests remain because we haven't swept them and added proper type annotations. That's the next todo.

PR #1564, opened 2016-05-31
they did not do the same thing.

rexebin:
now they do.


wardbell:
Actually, you misunderstood the example. The one you "corrected" is meant to demonstrate property binding. 

I think I can clarify the label so that is more obvious.

Closed by PR #1568

rexebin:
My brain wasn't working logically at the time. I can see now, they indeed do the same thing. However, it can be clearer so no one have a change to misunderstand like I did.

wardbell:
And I think I have improved it. Thanks for bringing it to my attention.

PR #1572, opened 2016-06-01
docs(pipes): add safari compatibility

Foxandxss:
The plunker in safari was broken because it needs that concrete polyfill.

You can read more at https://github.com/angular/angular/issues/3333#issuecomment-219561520

### DIFF
--- a/public/docs/_examples/pipes/ts/index.html
+++ b/public/docs/_examples/pipes/ts/index.html
@@ -12,6 +12,7 @@
     <script src="node_modules/zone.js/dist/zone.js"></script>
     <script src="node_modules/reflect-metadata/Reflect.js"></script>
     <script src="node_modules/systemjs/dist/system.src.js"></script>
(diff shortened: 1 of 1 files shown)


wardbell:
This is incomplete in several respects
* [ ] Let's keep shims together. The new polyfil should be moved up near (just below?) the other browser shim. If it must be where it is, explain why.

* [ ] We can't get this from npm? Why. We strive to have development work entirely offline w/o any 404s.  

* [ ] If we can install with npm, must update plunker tool to map back to the web (using npmcdn preferably).

* [ ] Explain why this shim is needed in this example and not everywhere. If only here, must update the chapter so that the reader knows to add this shim.

* [ ] If should add this shim everywhere, then explain in the Quickstart and add to all `index.html`.

* [ ] If should add this shim everywhere, reconcile with the CLI's position on shims (I don't know what that is).

Sorry, I know this is a royal PITA. But just slapping a shim into the example is not the answer.

Issue #1522, opened 2016-05-25
Add e2e tests for Basics Dart examples


chalin:
- [x] architecture - #1524
- [x] displaying-data - #1524
- [x] user-input - #1524
- [x] forms - #1524
- [x] dependency-injection - #1573
- [x] template-syntax - #1465
- cheatsheet, N/A.
- [ ] style-guide, cf. #1597.
- glossary, N/A.

Contributes to #1521.

chalin:
There is currently no plan to create a style guide. This was the only remaining subitem. Closing.

Issue #1508, opened 2016-05-25
Update Basics to new design scheme using Jade extends/blocks


chalin:
- [x] overview - #1410
- [x] architecture - #1807
- [x] displaying-data - #1819
- [ ] user input
- ~forms: TS and Dart versions have diverged too much~
- [x] dependency-injection - #1573
- [x] template-syntax - #1465
- ~cheatsheet~
- ~style guide (no Dart version yet)~
- [x] glossary - #2100

Contributes to #1506.

wardbell:
Please hold on this project!

chalin:
> Please hold on this project!

Ok. FYI, I am almost done (95%) with the dependency-injection chapter. I'll commit on Tuesday. Will hold off working on the other chapters until I hear news from you. 

cc: @kwalrath

chalin:
We will no longer be implementing this new scheme.

Issue #1563, opened 2016-05-31
Alternatives to OpaqueToken for use in dependency-injection (DI) chapter


chalin:
[OpaqueToken](https://angular.io/docs/dart/latest/api/angular2.core/OpaqueToken-class.html) cannot be implemented in Dart, see https://github.com/dart-lang/angular2/issues/35. I am opening this issue to hold a thread of discussion about alternatives for both the Dart and TS sides --- given that we want minimize difference across languages so as to provide a consistent UX. I'll focus the discussion on Dart, but the proposal holds for both.

For the purpose of this discussion, let's reuse the example given in the TS [OpaqueToken](https://angular.io/docs/ts/latest/guide/dependency-injection.html#!#opaque-token) section of the **angular.io di chapter**. Consider the following snippets of a Dart version of that example:

``` dart
const APP_CONFIG = const OpaqueToken('app.config'); // Token definition
// ... providers
providers: const [const Provider(APP_CONFIG, useValue: CONFIG)]
// ... constructor
AppComponent(@Inject(APP_CONFIG) Config config) {...}
```

Given that an `OpaqueToken` is no better than a `String` in Dart, why not simply use an fresh type name introduced via an abstract class declaration:

``` dart
abstract class APP_CONFIG {} // Token definition
```

Given that [TS also has abstract classes](https://www.typescriptlang.org/docs/handbook/classes.html#abstract-classes), this approach could be used consistently across languages.

Thoughts @vicb @mhevery @wardbell @thso @kwalrath ?

chalin:
Btw, another alternative to using an abstract class type name is to use a **named function** (closure) as a token.

chalin:
This issue was moved to dart-lang/site-webdev#256

PR #1573, opened 2016-06-01
docs(dependency-injection): revise Dart and TS code and prose

chalin:
Contributes to #1508 and #1522.

- The main change on the TS side is to present best practices first, rather than presenting sample code and then saying "don't do that".
- The "Optional dependencies" section has been moved towards the end of the provider variants section and its sample code has been simplified.
- Reworked *_1.{ts,dart} sources so that main_1.{ts,dart} can actually be run, which helped find some bugs in those sources.
- Despite https://github.com/angular/angular/issues/8928, the Dart and TS OpaqueToken sections have been left in the prose. They can be revisited later.

Enhancements to the Dart code include:

- Added TestComponent.
- Added a provider sample that makes use of an app config Map (to match TS); also added the more Dart-like use of a custom app configuration class, whose instance is initialized via the cascade operator.
- Fixed Engine class hierarchy so that a single field named `cylinders` is defined (rather than two fields: one in the Engine supertype and one in the subclass), because having two fields is a bad design smell.
- Similarly, fixed Logger hierarchy so that a single field named `logs` is defined. Also made `logs` a read-only field.
- Reintroduced the SilentLogger().

### DIFF
--- a/public/docs/_examples/dependency-injection/dart/lib/app_component.dart
+++ b/public/docs/_examples/dependency-injection/dart/lib/app_component.dart
@@ -11,6 +11,7 @@ import 'user_service.dart';
 //PENDING: check whether we intend to hide injector_component.dart & providers_component.dart; if so, change docregion name?
 // #enddocregion imports
 import 'injector_component.dart';
+import 'test_component.dart';
 import 'providers_component.dart';
 
 @Component(
@@ -31,21 +32,21 @@ import 'providers_component.dart';
       CarComponent,
       HeroesComponent,
       InjectorComponent,
+      TestComponent,
       ProvidersComponent
     ],
-// #docregion providers
+    // #docregion providers
     providers: const [
-      Logger,
-      UserService,
-      const Provider(AppConfig, useValue: config1)]
-// #enddocregion providers
+      Logger, UserService,
+      const Provider(APP_CONFIG, useFactory: heroDiConfigFactory)]
+    // #enddocregion providers
 )
 class AppComponent {
   final UserService _userService;
   final String title;
 
-  //#docregion ctor
-  AppComponent(AppConfig config, this._userService)
+  // #docregion ctor
+  AppComponent(@Inject(APP_CONFIG) AppConfig config, this._userService)
       : title = config.title;
   // #enddocregion ctor
 
@@ -64,4 +65,3 @@ class AppComponent {
   String get userInfo => 'Current user, ${user.name}, is'
       '${isAuthorized ? "" : " not"} authorized. ';
 }
-// #enddocregion
--- a/public/docs/_examples/dependency-injection/dart/lib/app_component.dart
+++ b/public/docs/_examples/dependency-injection/dart/lib/app_component.dart
@@ -1,6 +1,3 @@
-// #docplaster
-// #docregion
-// #docregion imports
 import 'package:angular2/core.dart';
 
 import 'app_config.dart';
@@ -8,8 +5,6 @@ import 'car/car_component.dart';
 import 'heroes/heroes_component.dart';
 import 'logger_service.dart';
 import 'user_service.dart';
-//PENDING: check whether we intend to hide injector_component.dart & providers_component.dart; if so, change docregion name?
-// #enddocregion imports
 import 'injector_component.dart';
 import 'test_component.dart';
 import 'providers_component.dart';
@@ -62,6 +57,6 @@ class AppComponent {
     return _userService.user;
   }
 
-  String get userInfo => 'Current user, ${user.name}, is'
-      '${isAuthorized ? "" : " not"} authorized. ';
+  String get userInfo => 'Current user, ${user.name}, is' +
+    (isAuthorized ? '' : ' not') + ' authorized. ';
 }
--- a/public/docs/_examples/dependency-injection/dart/lib/logger_service.dart
+++ b/public/docs/_examples/dependency-injection/dart/lib/logger_service.dart
@@ -3,10 +3,12 @@ import 'package:angular2/core.dart';
 
 @Injectable()
 class Logger {
-  List<String> logs = [];
+  List<String> _logs = [];
+
+  List<String> get logs => _logs;
 
   void log(String message) {
-    logs.add(message);
+    _logs.add(message);
     print(message);
   }
 }
--- a/public/docs/_examples/dependency-injection/dart/lib/logger_service.dart
+++ b/public/docs/_examples/dependency-injection/dart/lib/logger_service.dart
@@ -4,7 +4,6 @@ import 'package:angular2/core.dart';
 @Injectable()
 class Logger {
   List<String> _logs = [];
-
   List<String> get logs => _logs;
 
   void log(String message) {
--- a/public/docs/dart/latest/guide/dependency-injection.jade
+++ b/public/docs/dart/latest/guide/dependency-injection.jade
@@ -95,8 +95,10 @@ block config-obj-maps
 
 block what-should-we-use-as-token
   :marked
-    But what should we use as the token?
-    We could use the **[Map][]** interface.
+    But what should we use as the token? 
+    While we _could_ use **[Map][]**, we _should not_ because (like
+    `String`) `Map` is too general. Our app might depend on several maps, each
+    for a different purpose.
 
     [Map]: https://api.dartlang.org/stable/dart-core/Map-class.html
 
@@ -107,32 +109,32 @@ block what-should-we-use-as-token
       Dart doesn't have this limitation;
       every class implicitly defines an interface,
       so interface names are just class names.
-
-  :marked
-    The problem with using `Map` as a token, is not that it is an abstract
-    class, but rather that it is very general purpose (like `String`) and so
-    our application may depend on several maps each used for a
-    different purpose.
+      `Map` is a *valid* token even though it's the name of an abstract class;
+      it's just *unsuitable* as a token because it's too general.
 
 block dart-map-alternative
   :marked
     As an alternative to using a configuration `Map`, we can define
-    a custom class type:
+    a custom configuration class:
 
   +makeExample('dependency-injection/ts/app/app.config.ts','config-alt','app/app-config.ts (alternative config)')(format='.')
 
   :marked
-    One of the key benefits of defining a configuration object class, is
-    strong static checking relative to configuration property names and their types.
-    The Dart [cascade operator][cascade-op] provides a convenient means of initializing
-    the configuration object. Use of the cascaded object initialization idiom
+    Defining a configuration class has a few benefits. One key benefit
+    is strong static checking: we'll be warned early if we misspell a property
+    name or assign it a value of the wrong type.
+    The Dart [cascade notation][cascade] (`..`) provides a convenient means of initializing
+    a configuration object. 
+
+    Use of the cascaded initialization idiom
     prevents us from defining our configuration object instance as `const`. 
-    As a consequencem we cannot declare a [value provider](#value-provider),
-    but a [factory provider](#factory-provider) does the trick &mdash; as
-    we illustrate by providing and injecting the configuration object in our
+    As a consequence, we cannot declare a [value provider](#value-provider),
+    but a [factory provider](#factory-provider) does the trick.
+
+    We illustrate this below by providing and injecting the configuration object in our
     top-level `AppComponent`:
 
-    [cascade-op]: https://www.dartlang.org/docs/dart-up-and-running/ch02.html#cascade
+    [cascade]: https://www.dartlang.org/docs/dart-up-and-running/ch02.html#cascade
 
   +makeExcerpt('lib/app_component.dart','providers')
   +makeExcerpt('lib/app_component.dart','ctor')
--- a/public/docs/dart/latest/guide/dependency-injection.jade
+++ b/public/docs/dart/latest/guide/dependency-injection.jade
@@ -126,13 +126,11 @@ block dart-map-alternative
     The Dart [cascade notation][cascade] (`..`) provides a convenient means of initializing
     a configuration object. 
 
-    Use of the cascaded initialization idiom
-    prevents us from defining our configuration object instance as `const`. 
-    As a consequence, we cannot declare a [value provider](#value-provider),
-    but a [factory provider](#factory-provider) does the trick.
-
-    We illustrate this below by providing and injecting the configuration object in our
-    top-level `AppComponent`:
+    If we use cascades, the configuration object can't be declared `const` and
+    we can't use a [value provider](#value-provider). 
+    A solution is to use a [factory provider](#factory-provider).
+    We illustrate this next. We also show how to provide and inject the
+    configuration object in our top-level `AppComponent`:
 
     [cascade]: https://www.dartlang.org/docs/dart-up-and-running/ch02.html#cascade
 
--- a/public/docs/dart/latest/guide/dependency-injection.jade
+++ b/public/docs/dart/latest/guide/dependency-injection.jade
@@ -1,279 +1,138 @@
-include ../_util-fns
+extends ../../../ts/latest/guide/dependency-injection.jade
 
-+includeShared('{ts}', 'intro')
+block includes
+  include ../_util-fns
+  - var _thisDot = '';
 
-:marked
-  The complete source code for the example app in this chapter is
-  [in GitHub](https://github.com/angular/angular.io/tree/master/public/docs/_examples/dependency-injection/dart).
+block ctor-syntax
+  .l-sub-section
+    :marked
+      We also leveraged Dart's constructor syntax for declaring parameters and
+      initializing properties simultaneously.
 
-+includeShared('{ts}', 'why-1')
-+makeExample('dependency-injection/dart/lib/car/car_no_di.dart', 'car', 'lib/car/car.dart (without DI)')
-+includeShared('{ts}', 'why-2')
-+makeTabs(
-  'dependency-injection/dart/lib/car/car.dart, dependency-injection/dart/lib/car/car_no_di.dart',
-  'car-ctor, car-ctor',
-  'lib/car/car.dart (excerpt with DI), lib/car/car.dart (excerpt without DI)')(format=".")
-+includeShared('{ts}', 'why-3-1')
-+includeShared('{ts}', 'why-3-2')
-- var stylePattern = { otl: /(new Car.*$)/gm };
-+makeExample('dependency-injection/dart/lib/car/car_creations.dart', 'car-ctor-instantiation', '', stylePattern)(format=".")
-+includeShared('{ts}', 'why-4')
-.l-sub-section
+block service-in-its-own-file
+  //- N/A
+
+block one-class-per-file-ts-tradeoffs
+  //- N/A
+
+block injectable-not-always-needed-in-ts
+  //- The [Angular 2 Dart Transformer](https://github.com/angular/angular/wiki/Angular-2-Dart-Transformer)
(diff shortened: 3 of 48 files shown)


chalin:
@Foxandxss @wardbell @thso @kwalrath : this is ready for review. As usual, it is less painful to view the [diff while ignoring changes in whitespace (such as indentation)](https://github.com/angular/angular.io/pull/1573/files?w=1).

public/docs/_examples/dependency-injection/dart/lib/app_component.dart
```diff
@@ -11,6 +11,7 @@ import 'user_service.dart';
 //PENDING: check whether we intend to hide injector_component.dart & providers_component.dart; if so, change docregion name?
```
kwalrath: Should this comment really be visible? Also, it doesn't reflect test_component.dart.

public/docs/_examples/dependency-injection/dart/lib/app_component.dart
```diff
@@ -31,21 +32,21 @@ import 'providers_component.dart';
...
     ],
-// #docregion providers
+    // #docregion providers
     providers: const [
-      Logger,
-      UserService,
-      const Provider(AppConfig, useValue: config1)]
-// #enddocregion providers
+      Logger, UserService,
+      const Provider(APP_CONFIG, useFactory: heroDiConfigFactory)]
```
kwalrath: APP_CONFIG seems like an un-Darty name. (But if it's mentioned in text, it'd be easier to leave this capitalization.)

public/docs/_examples/dependency-injection/dart/lib/app_component.dart
chalin: That was an old comment that I missed (I had only searched for todo|fixme). I have removed it since it is not relevant (the entire source file is never shown).

public/docs/_examples/dependency-injection/dart/lib/app_component.dart
chalin: The OpaqueToken section will need to be revisited once #1563 gets resolved so I decided to leave the token name like it is in the TS example (because then the token name _and_ the interface name will match, both being `AppConfig`).

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -9,34 +10,35 @@ include ../_util-fns
...
 :marked
   Our `Car` creates everything it needs inside its constructor.
   What's the problem?
-
   The problem is that our `Car` class is brittle, inflexible, and hard to test.
 
   Our `Car` needs an engine and tires. Instead of asking for them,
-  the `Car` constructor creates its own copies by "new-ing" them from
+  the `Car` constructor instantiates its own copies from
   the very specific classes, `Engine` and `Tires`.
```
kwalrath: Remove the "," after "classes" while you're in here. (It's a restrictive phrase, not a parenthetical one.) _Or_ delete the "the" before "very" and keep the comma; that way the phrase becomes parenthetical.

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -377,144 +375,110 @@ include ../_util-fns
...
+      li <b>Consistency:</b> All services follow the same rules, and we don't have to wonder why #{_a} #{_decorator} is missing.
+
+:marked
+  Injectors are also responsible for instantiating components
+  like `HeroesComponent`. Why haven't we marked `HeroesComponent` as
+  `@Injectable()`?
+
+  We *can* add it if we really want to. It isn't necessary because the
+  `HeroesComponent` is already marked with `@Component`, and this
+  !{_decorator} class (like `@Directive` and `@Pipe` which we will study later
```
kwalrath: "study" sounds so serious. Also, "which" (when introducing a phrase/clause) should have a comma before it; if a comma doesn't seem right, then you should probably be using "that" instead.

```
` which we will study later on
```
->
```
`, which we'll see later
```

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -377,144 +375,110 @@ include ../_util-fns
...
+  on) is a subtype of <a href="#{injMetaUrl}">InjectableMetadata</a>.  It is in
+  fact <a href="#{injMetaUrl}">InjectableMetadata</a> #{_decorator}s that
```
kwalrath: Remove the link from the second occurrence of InjectableMetadata.

public/docs/ts/latest/guide/dependency-injection.jade
chalin: removed ","

public/docs/_examples/dependency-injection/dart/lib/logger_service.dart
```diff
@@ -3,10 +3,12 @@ import 'package:angular2/core.dart';
 
 @Injectable()
 class Logger {
-  List<String> logs = [];
+  List<String> _logs = [];
+
```
kwalrath: The blank line above the getter seems excessive. Is it standard? (It could be; I don't happen to know, and I don't see anything definitive in Effective Dart.)

public/docs/ts/latest/guide/dependency-injection.jade
chalin: Fixed
> (like `@Directive` and `@Pipe`, which we'll learn about later)

public/docs/ts/latest/guide/dependency-injection.jade
chalin: done

Foxandxss:
I have nothing to add, @kwalrath is doing an awesome job on the review. Sounds fine for me.

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -748,105 +717,117 @@ code-example(format, language="html").
...
+  +makeExcerpt('app/app.component.ts','providers')
+
+#optional
 :marked
-// end typescript only
+  ## Optional dependencies
+
+  Our `HeroService` *requires* a `Logger`, but what if it could get by without
+  a logger?
+  As is illustrated next, we can indicate that to the dependency injection
```
kwalrath: This sounds very formal. Also, we should mention `@Optional()` in the text, not just in the code. Maybe:

We can tell Angular that the dependency is optional by annotating the constructor argument with `@Optional()`:

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -748,105 +717,117 @@ code-example(format, language="html").
...
++ifDocsFor('ts')
+  +makeExample('dependency-injection/ts/app/providers.component.ts','import-optional', '')
++makeExample('dependency-injection/ts/app/providers.component.ts','provider-10-ctor', '')(format='.')
 
-// Skip for Dart (we have another example)
 :marked
-  Or we can provide and inject the configuration object in our top-level `AppComponent`.
-+makeExample('dependency-injection/ts/app/app.component.ts','providers', 'app/app.component.ts (providers)')(format=".")
-+makeExample('dependency-injection/ts/app/app.component.ts','ctor', 'app/app.component.ts (constructor)')(format=".")
+  Be prepared for a null logger. If we don't register one somewhere up the line,
```
kwalrath: Instead of 2nd person here, let's continue with a "we" focus. Maybe:

When using `@Optional()`, our code must be prepared for a null value. If we don't register a logger somewhere up the line, the injector will set the value of `logger` to `null`.

public/docs/_examples/dependency-injection/dart/lib/logger_service.dart
chalin: dartfmt doesn't seem to care. It leaves the line there if present, and doesn't add it if it is absent. I'll all for compactness, so I removed that extra line.

public/docs/ts/latest/guide/dependency-injection.jade
chalin: > This sounds very formal ...

Yes, what you propose is much better, thanks. Fixed.

public/docs/ts/latest/guide/dependency-injection.jade
chalin: Fixed.

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -683,61 +652,61 @@ code-example(format, language="html").
...
   In our sample, we need it only in the `HeroesComponent`,
-  where it replaces the previous `HeroService` registration in the metadata `providers` array.
+  where it replaces the previous `HeroService` registration in the metadata `providers` #{_array}.
   Here we see the new and the old implementation side-by-side:
-// #enddocregion providers-factory-5
+
+- var stylePattern = { otl: /(providers.*),$/gm };
 +makeTabs(
   `dependency-injection/ts/app/heroes/heroes.component.ts,
   dependency-injection/ts/app/heroes/heroes.component.1.ts`,
```
kwalrath: I get BAD FILENAME for v2 in the TS docs. (bad docregion?)

Also, the Dart docs show this:

```
import 'hero_list_component_2.dart';
import 'hero_service_1.dart';
/*
import 'hero_list_component.dart';
import 'hero_service.dart';
*/
```

They should instead look more like the v3 imports, shouldn't they?

```
import 'hero_list_component.dart';
import 'hero_service.dart';
```

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -748,105 +717,117 @@ code-example(format, language="html").
...
 :marked
   We have similar good fortune when we write a constructor that requires an injected class-based dependency.
   We define a constructor parameter with the `HeroService` class type,
   and Angular knows to inject the
   service associated with that `HeroService` class token:
-// #enddocregion tokens-2
-+makeExample('dependency-injection/ts/app/providers.component.ts','provider-8-ctor')(format=".")
-// #docregion tokens-3
+
++makeExample('dependency-injection/ts/app/heroes/hero-list.component.ts', 'ctor-signature')
```
kwalrath: I'm so glad this doesn't say ProviderComponent8 any more!

public/docs/ts/latest/guide/dependency-injection.jade
chalin: Good catch. Fixed.

public/docs/ts/latest/guide/dependency-injection.jade
chalin: :)

public/docs/dart/latest/guide/dependency-injection.jade
```diff
@@ -1,279 +1,138 @@
...
-+includeShared('{ts}', 'tokens-opaque-2')
-+makeExample('dependency-injection/dart/lib/providers_component.dart','use-opaque-token')(format=".")
-:marked
-  Here's an example of providing configuration information
-  for an injected class. First define the class:
-+makeExample('dependency-injection/dart/lib/providers_component.dart','configurable-logger')(format=".")
-:marked
-  Then inject that class and its configuration information:
-+makeExample('dependency-injection/dart/lib/providers_component.dart','providers-7')(format=".")
+    The problem with using `Map` as a token, is not that it is an abstract
```
kwalrath: This is a really long sentence, and it continues the thought of the previous paragraph. Let's shorten this sentence and combine the two paragraphs. Maybe:

But what should we use as the token? We can't use `Map` because it's too general purpose. Our app might depend on several maps, each for a different purpose.

Dart difference: Interfaces are valid tokens
... are just class names. `Map` is a _valid_ token even though it's the name of an abstract class; it's just _unsuitable_ as a token because it's too general.

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -748,105 +717,117 @@ code-example(format, language="html").
...
-    There is no interface type information left for Angular to find at runtime.
-// end Typescript only
-
-// #docregion tokens-opaque-1
-<a id="opaque-token"></a>
-- var lang = current.path[1]
-- var opaquetoken = lang == 'dart' ? '<code>OpaqueToken</code>' : '<a href="../api/core/OpaqueToken-class.html"><code>OpaqueToken</code></a>'
-h3 OpaqueToken
-p.
   The solution is to define and use an !{opaquetoken}.
```
kwalrath: By this time we might have forgotten the problem. :) How about:

> The solution to specifying non-class dependencies is...

Of course, that raises the question (especially for Dart): what should that solution be? is it the only one? Maybe "The solution" -> "One solution"?

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -748,105 +717,117 @@ code-example(format, language="html").
...
+:marked
+  Now we can inject the configuration object into any constructor that needs it, with
+  the help of an `@Inject` #{_decorator}:
+
++makeExample('dependency-injection/ts/app/app.component.2.ts','ctor')(format=".")
+
+- var configType = _docsFor == 'dart' ? '<code>Map</code>' : '<code>AppConfig</code>'
 .l-sub-section
   :marked
     Although it plays no role in dependency injection,
```
kwalrath: Define "it" before using it:

it <-> the !{configType} interface

public/docs/dart/latest/guide/dependency-injection.jade
```diff
@@ -1,279 +1,138 @@
...
+    a custom class type:
 
-+includeShared('{ts}', 'summary')
+  +makeExample('dependency-injection/ts/app/app.config.ts','config-alt','app/app-config.ts (alternative config)')(format='.')
 
-+includeShared('{ts}', 'appendix-explicit-injector-1')
-+makeExample('dependency-injection/dart/lib/injector_component.dart', 'injector', 'lib/injector_component.dart')
-+includeShared('{ts}', 'appendix-explicit-injector-2')
+  :marked
+    One of the key benefits of defining a configuration object class, is
```
kwalrath: Move the comma to be before "relative". Maybe change "relative..." to be clearer. I'm not sure if you're comparing the map & class benefits, or something else.

(Where the comma is, it's equivalent to "This, is that", which is you wouldn't normally do. If the sentence seems too hard to read without the comma, I'd rewrite to make it clearer. Maybe: Defining a configuration object class has a few benefits. One key benefit is strong static checking: we'll be warned early if we misspell a property name or assign it a value of the wrong type.)

public/docs/dart/latest/guide/dependency-injection.jade
```diff
@@ -1,279 +1,138 @@
...
+    strong static checking relative to configuration property names and their types.
+    The Dart [cascade operator][cascade-op] provides a convenient means of initializing
```
kwalrath: Cascade isn't an operator, strictly speaking. Also, be sure to show it "(`..`)", not just name it.

I'm not sure what the most important points of this discussion are. It needs to be clearer, and probably broken into multiple paragraphs.

public/docs/dart/latest/guide/dependency-injection.jade
```diff
@@ -1,279 +1,138 @@
...
+    the configuration object. Use of the cascaded object initialization idiom
+    prevents us from defining our configuration object instance as `const`. 
+    As a consequencem we cannot declare a [value provider](#value-provider),
```
kwalrath: consequencem -> consequence,

(if it isn't rewritten out of existence)

kwalrath:
Looks great! The only real problem I saw was the "key benefits of defining a configuration object class" paragraph, which needs to be clarified and probably broken up.

kwalrath:
@Foxandxss & @wardbell: I don't know the TS side, so you still might want to take a look to make sure the new text is OK by you...

wardbell:
Will look

public/docs/dart/latest/guide/dependency-injection.jade
chalin: fixed per our discussion

public/docs/ts/latest/guide/dependency-injection.jade
chalin: Both good points. Fixed (and yes, it is only one solution).

public/docs/ts/latest/guide/dependency-injection.jade
chalin: done

public/docs/dart/latest/guide/dependency-injection.jade
chalin: I prefer your rewrite. Fixed.

public/docs/dart/latest/guide/dependency-injection.jade
chalin: Fixed. Changed 'operator' to 'syntax'.

public/docs/dart/latest/guide/dependency-injection.jade
chalin: fixed

chalin:
Updates done.

> Looks great! The only real problem I saw was the "key benefits..." paragraph

That paragraph has been reworked.

public/docs/dart/latest/guide/dependency-injection.jade
```diff
@@ -107,32 +109,32 @@ block what-should-we-use-as-token
...
-    strong static checking relative to configuration property names and their types.
-    The Dart [cascade operator][cascade-op] provides a convenient means of initializing
-    the configuration object. Use of the cascaded object initialization idiom
+    Defining a configuration class has a few benefits. One key benefit
+    is strong static checking: we'll be warned early if we misspell a property
+    name or assign it a value of the wrong type.
+    The Dart [cascade notation][cascade] (`..`) provides a convenient means of initializing
+    a configuration object. 
+
+    Use of the cascaded initialization idiom
```
kwalrath: On the right track, but this could still be simplified. Maybe:

If we use cascades, the configuration object can't be declared `const` and we can't use
a [value provider](...). The solution is to use a [factory provider](...):

public/docs/dart/latest/guide/dependency-injection.jade
```diff
@@ -107,32 +109,32 @@ block what-should-we-use-as-token
...
     prevents us from defining our configuration object instance as `const`. 
-    As a consequencem we cannot declare a [value provider](#value-provider),
-    but a [factory provider](#factory-provider) does the trick &mdash; as
-    we illustrate by providing and injecting the configuration object in our
+    As a consequence, we cannot declare a [value provider](#value-provider),
+    but a [factory provider](#factory-provider) does the trick.
+
+    We illustrate this below by providing and injecting the configuration object in our
```
kwalrath: Could we just delete this paragraph? If not, let's simplify the wording.

kwalrath:
I had a [couple more quibbles about the cascade wording](https://github.com/angular/angular.io/pull/1573/commits/dbcff2ddbbf88033ec3cb730c6b2bb66d49fda91#r65644008), but otherwise LGTM.

public/docs/dart/latest/guide/dependency-injection.jade
chalin: Yes, what you propose is better. Done.

public/docs/dart/latest/guide/dependency-injection.jade
chalin: Actually, that paragraph is there to match up with the TS prose which reads:
> ... we can provide and inject the configuration object in our top-level `AppComponent`.

How about:
> We illustrate this next. We also show how to provide and inject the
    configuration object in our top-level `AppComponent`:

chalin:
Updates done.

kwalrath:
LGTM. I'll merge this unless someone screams NO...

PR #1589, opened 2016-06-02
[WIP] chore(dart/api): enable API doc generation for Dart

chalin:
This is a first step of many that will incrementally introduce proper API docs for Dart. This first step enables rudimentary API pages for Dart (generated from the TS source but that will be using Dart API examples).

### DIFF
--- a/public/docs/dart/latest/api/index.jade
+++ b/public/docs/dart/latest/api/index.jade
@@ -2,9 +2,9 @@
   > **Known issues:** Although this generated API reference displays Dart APIs
   and includes some Dart examples, the text reflects the TypeScript APIs. The
   angular.io issue tracker contains [all known issues][api-issues]; if you
-  notice others, please [report them][ngio-issues]. Thanks!
+  notice others, please [report them][new-issue]. Thanks!
 
-  [ngio-issues]: https://github.com/angular/angular.io/issues
-  [api-issues]: https://github.com/angular/angular.io/labels/api
+  [new-issue]: https://github.com/angular/angular.io/issues/new?labels=dart,api&title=%5BDart%5D%5BAPI%5D%20
+  [api-issues]: https://github.com/angular/angular.io/issues?q=label%3Aapi+label%3Adart
 
 api-list(src="api-list.json")
--- a/public/docs/dart/latest/api/index.jade
+++ b/public/docs/dart/latest/api/index.jade
@@ -1,10 +1,10 @@
-.l-main-section
-  h2 Beta
+:marked
+  > **Known issues:** Although this generated API reference displays Dart APIs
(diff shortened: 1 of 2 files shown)


public/docs/dart/latest/api/index.jade
```diff
@@ -1,10 +1,10 @@
...
+  > **Known issues:** Although this generated API reference displays Dart APIs
+  and includes some Dart examples, the text reflects the TypeScript APIs. The
+  angular.io issue tracker contains [all known issues][api-issues]; if you
+  notice others, please [report them][ngio-issues]. Thanks!
 
-  p.
-    The proposed Angular 2 API does not yet have Dart-specific documentation.
-    However, because the Dart and JavaScript APIs are generated from the same source,
-    you might find the JavaScript API docs helpful:
+  [ngio-issues]: https://github.com/angular/angular.io/issues
```
kwalrath: How about ngio-issues -> new-issue and use this URL:

https://github.com/angular/angular.io/issues/new

kwalrath:
Before pushing this, we need to do a few things:

* [ ] Make sure the issue search uses the right labels (probably 'api' & 'dart': https://github.com/angular/angular.io/issues?q=is%3Aissue+label%3Aapi+label%3Adart)
* [ ] Add known issues to the database.
* [ ] Add text to each page linking to known issues.
* [ ] Either add Dart examples everywhere or change the text to NOT imply that we have them everywhere.
* [ ] Make sure the output is useful (which maybe it will be once Dart examples are everywhere?). I didn't realize that all the generated entries in the docs are directly from TS, not from generated Dart. 
* [ ] Be more specific in the caveat about how TypeScripty the docs are. Perhaps create a page telling people how to figure out what the API actually is? E.g., how do you find out all the args to bootstrap()? Array -> List, string -> String, boolean -> bool etc. Consider pointing people to both the angular.io and dartdocs.org docs? E.g. https://www.dartdocs.org/documentation/angular2/latest/angular2.common/Control-class.html seems to have correct type info except in methods, which lack types (it also shows lots of inherited properties); http://localhost:3000/docs/dart/latest/api/common/index/Control-class.html is much shorter and/but has little useful information except the text.

public/docs/dart/latest/api/index.jade
chalin: How about this:
> https://github.com/angular/angular.io/issues/new?labels=dart,api&title=[Dart][API]

The title prefix is because not everyone has the permission to set the labels, so a title prefix is better than nothing.

kwalrath:
I like the new links.

chalin:
A complete solution is on its way.

chalin:
A more complete solution is on its way.

PR #1594, opened 2016-06-03
chore: move to provide object literal

Foxandxss:
I used the "provided object literal" to define this new syntax. We were using it already and I agreed with it.

I removed two examples from `dependency injection` and touched Dart prose and one of the files so please @chalin or @kwalrath confirm that I didn't break it.

Removing two examples the numbers jump from example 1 to example 3a, but that is just internal. I can fix the numbers if needed.


Foxandxss:
Seems like I left some behind, working on that.

kwalrath:
I took a quick look; @chalin will take a closer one.

chalin:
@Foxandxss : let me know once this is ready for review.

Foxandxss:
I definitely broke the Dart part, but @chalin will fix it when this hit master.

Foxandxss:
Also changed a blue box where it explains differences from Dart to Typescript in regard of provide. I changed the prose there, but feel free to change it too.

wardbell:
Proceeding with the agreement that @chalin will fix what we broke in Dart at his convenience.

PR #1604, opened 2016-06-05
docs(upgrade): restrict component directive to elements

gkalpak:
The `heroDetail` directive is shown as an example of a component directive, such as those built using the `.component()` helper in v1.5+. In order for the directive to be functionaly equivalent with the component directive created with `.component`, it needs to be restricted to elements only. (If the `restrict` is omitted, it defaults to `'AE'`, which would also match attributes.)

/cc @teropa

### DIFF
--- a/public/docs/_examples/upgrade-adapter/ts/app/hero-detail.directive.ts
+++ b/public/docs/_examples/upgrade-adapter/ts/app/hero-detail.directive.ts
@@ -1,6 +1,7 @@
 // #docregion
 export function heroDetailDirective() {
   return {
+    restrict: 'E',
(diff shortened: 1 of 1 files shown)


Foxandxss:
From [here](https://github.com/angular/angular.js/blob/master/src/ng/compile.js#L1135) I can see they are restricted to `E` by default.

gkalpak:
Components are restricted to `'E'`. This PR is about changing a directive (created via `.directive()`) - by default restricted to `'EA'`.

wardbell:
Merged. No harm in being clear. Thanks.

PR #1606, opened 2016-06-05
docs(cli-quickstart): add cli 5 min quickstart

Foxandxss:
I took #1241 and got the ideas to create a QuickStart like the one we have, but using the CLI.

So I merged the ideas from @filipesilva and the ideas we already had in the QuickStart and created a new QuickStart with same ideas, same explanations (translated to CLI).

Now we have two QuickStart with the same content and goal but with different tools in the back.

Also, the project can be e2e-tested too.

Let me copy what Filipe said in his PR

>This PR includes both the Angular-CLI quickstart chapter and e2e test support for CLI applications.

>It does not add the CLI-Quickstart to the sidebar menu though, nor live example via plunker.

>To run e2e tests for CLI applications, you must have Angular-CLI installed globally and using the same version as specified in public/docs/_examples/package.json.

>When adding a new CLI-based chapter, create a project using ng new chapter-name and copy the contents of the chapter-name dir onto public/docs/_examples/chapter-name/ts/.

>Afterwards, add an example-config.js in the example project root, rename package.json to package.1.json, remove declare var module: { id: string }; from src/typings.d.ts (angular.io typings already include node).

>Finally, add public/docs/_examples/chapter-name/e2e-spec.js as usual.

Next step is extract the other part of his tutorial into a deploy cookbook.

I skipped the testing part, it is nice but I don't see it on a QuickStart of this sorts.


wardbell:
I tweaked the text and merged. Thanks @Foxandxss and @filipesilva.

Looking forward to the companion CLI cookbook

chalin:
I think that `tsconfig.json` is missing.
Does this supersede #1241? If so maybe #1241 should be closed.

Foxandxss:
Yes, seems that it is missing some files. We ignore a lot and I manually checked all of them but I surely missed that one at least.

This PR doesn't entirely supersede that one, just half of it. I left it open as a reminder that I need to do the other half of the job.

PR #1623, opened 2016-06-08
chore(node): recommend v5 via .nvmrc

chalin:
@wardbell @Foxandxss @kwalrath : enough of us have had enough problems with node that I think it is worth while setting the `.nvmrc` to v5.


Foxandxss:
Well, for ourselves at this project yes. But we are a few that use NVM, the rest doesn't.

chalin:
Sure, but this addition won't hurt those who don't use nvm.

Foxandxss:
That is right.

I don't like to constrain people to use certain versions and only recommend.

But I don't really mind adding this.

chalin:
As you know it doesn't force anyone to use v5 (one can always `nvm use 6`);
its just that v5 is the best default for now.

PR #1625, opened 2016-06-08
docs(cb-a11y): add accessibility (a11y) cookbook

AlmeroSteyn:
**READY FOR REVIEW**

Current and up to date PR for the **a11y cookbook**.

Also contains copy-edits from @naomiblack from https://github.com/angular/angular.io/pull/1181 .

Both https://github.com/angular/angular.io/pull/1049 and https://github.com/angular/angular.io/pull/1181 are now outdated in terms of content and should be closed, but conserved for comments therein.


public/docs/ts/latest/cookbook/a11y.jade
```diff
@@ -0,0 +1,1087 @@
...
+  ### Skiplinks
+  
+  For a sighted person who navigates the web using a mouse, it is very easy to skip sections of web pages.
+  This could be because they are returning to a website they know very well, or because they can immediately see and interact
+  with a specific section of interest.
+
+  Similarly, we need to provide quick in-page links to help users who rely on assistive
+  technology to navigate our page.
+
+  These are called `skiplinks` and remain completely hidden until focused through `keyboard navigation`.
```
mattdistefano: I'm noting this here but it's more of a general question - are the code backticks appropriate for calling out technical _terms_ or should some of these really be ems?

public/docs/ts/latest/cookbook/a11y.jade
```diff
@@ -0,0 +1,1087 @@
...
+  </a11y-custom-button>
+
+:marked
+  **Important**:  Note the `role` and `tabindex`. 
+  
+  The `ARIA role` of this element is `button`. It tells any assistive technologies that this element is
+  a button, regardless of the original design of the HTML element. More information on this later.
+  
+  Setting the `tabindex` to `0` inserts the element in the default flow of`keyboard navigation` focus. This means that our
+  element also becomes accessible via the keyboard!
```
mattdistefano: IMO this is a little misleading since it's really the combination of `tabindex` and the `keydown` bindings that make it accessible via keyboard. So I'm wondering if it'd be more effective to just walk through the whole implementation of the component vs. calling out these two items specifically.

public/docs/_examples/cb-a11y/ts/app/dev-tools/a11y-pass/a11y-pass.component.html
```diff
@@ -0,0 +1,52 @@
...
+              aria-label="Close error">&times;</button>
+    </div>
+
+    <button #submitButton
+            type="submit"
+            class="btn btn-primary">Submit</button>
+
+    <div class="clearfix">
+      <a href="https://www.w3.org/TR/WCAG20/"
+         title="WCAG specification"
```
mattdistefano: Any reason for different text in the `title` vs. `aria-label`?

public/docs/ts/latest/cookbook/a11y.jade
```diff
@@ -0,0 +1,1087 @@
...
+  
+  We'll illustrate how to do this by creating a `button` that shows an `alert`, then sets focus on the `alert`
+  and then allows the user to close the error message with a `close button`.
+
++makeTabs('cb-a11y/ts/app/managing-focus/a11y-error-demo.component.html,cb-a11y/ts/app/managing-focus/a11y-error-demo.component.ts',
+null, 'a11y-error-demo.component.html,a11y-error-demo.component.ts')
+
+:marked
+  We are setting focus on an `alert` that starts out hidden. As `hidden` and `disabled` elements cannot accept focus,
+  we first need this element to become visible again. To give the browser time to apply this change we set our focus using
```
mattdistefano: Is there a better way to handle this than the timeout? I've typically done the same thing in my angular1 apps but it's always felt like a hack.

public/docs/ts/latest/cookbook/a11y.jade
```diff
@@ -0,0 +1,1087 @@
...
+:marked
+  ### How to apply an ARIA Role in Angular&nbsp;2
+  
+  We give an `ARIA Role` to an element by setting the value of the `role` attribute.
+
+  When we write this directly in our HTML
+  it is as simple as:
+
+code-example(language="html" escape="html" format="linenums").
+  <h2 role="alert">I am an alert.</h2>
```
mattdistefano: Does this run afoul of the [second rule of ARIA use](https://www.w3.org/TR/aria-in-html/#second-rule-of-aria-use)?

public/docs/ts/latest/cookbook/a11y.jade
```diff
@@ -0,0 +1,1087 @@
...
+  - menu
+  - menubar
+  - radiogroup
+  - tablist
+  - tree
+  - treegrid
+
+.l-sub-section
+  :marked
+    The names of these widget roles are self-explanatory. Visit the `W3C` to read more about
```
mattdistefano: Would it be worth linking to the [design patterns](https://www.w3.org/TR/wai-aria-practices/#aria_ex) too? IMO they have a lot of value but aren't linked very well from the widget roles page.

public/docs/ts/latest/cookbook/a11y.jade
```diff
@@ -0,0 +1,1087 @@
...
+  a screen reader you can ignore. It does offer a `40 minute per session` trial which helps the beginner `a11y` tester
+  also check `a11y` in this screen reader.
+  
+  `JAWS` is best used in `Internet Explorer 10+`.
+  
+  For an introduction on using `JAWS`, head over to `WebAIM` and read their article on 
+  [Using JAWS to Evaluate Web Accessibility](http://webaim.org/articles/jaws/).
+  
+:marked
+  #### VoiceOver (Apple OS X integrated screen reader) in Safari
```
mattdistefano: Worth calling out any mobile screen readers?

public/docs/ts/latest/cookbook/a11y.jade
```diff
@@ -0,0 +1,1087 @@
...
+  :marked
+    You will use the skills you gained in the `keyboard navigation` section when using a screen reader. First get 
+    comfortable with keyboard-only navigation before adding a screen reader to it. This will help reduce the 
+    confusion that first time users of screen readers usually get confronted by.
+
+.l-sub-section
+  :marked
+    You will not get the same experience in every `screen reader / browser combination`, so do not spend too much time
+    trying to fix *"errors"* in combinations not mentioned here. Often you will be trying to fix a technology 
+    integration issue and not a problem on your page. Focus on the three combinations
```
mattdistefano: Possible to cite a source for this?

public/docs/ts/latest/cookbook/a11y.jade
```diff
@@ -0,0 +1,1087 @@
...
+  
+  [NonVisual Desktop Access](http://www.nvaccess.org/), better known as `NVDA` is an open source screen reader for
+  `Microsoft Windows`. For best results, use this screen reader with `Mozilla Firefox`.
+  
+  Both `WebAIM` ([Using NVDA to Evaluate Web Accessibility](http://webaim.org/articles/nvda/)) as well as 
+  `The A11y Project` ([Getting Started with NVDA](http://a11yproject.com/posts/getting-started-with-nvda)) have 
+  excellent introductory guides for getting started with `NVDA`. We recommend that you read either one or both of them. 
+  
+:marked
+  #### JAWS (Most used screen reader) in IE 10+
```
mattdistefano: Does 'IE 10+' leave some ambiguity around whether you mean just IE 10 and 11 or IE 10, 11, and Edge?

public/docs/ts/latest/cookbook/a11y.jade
AlmeroSteyn: These do provide a particular emphasis highlight that is used a lot inside the docs.

public/docs/ts/latest/cookbook/a11y.jade
AlmeroSteyn: I don't see it doing that here. It is not changing the semantic meaning of the element itself buit indicating that it is important information. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role

public/docs/ts/latest/cookbook/a11y.jade
mattdistefano: Right, but typically the `code` backticks are for actual snippets of code or commands - variable names, hex values, CLI commands, that sort of thing. Whereas technical terms get the _em treatment_.

public/docs/ts/latest/cookbook/a11y.jade
AlmeroSteyn: I was actually thinking about it. But then decided that this was not the place for it. Wanted to home in on a few dev tools that can easily be used during development time, and especially to introduce developers to testing accessibility matters. The mobile screen readers, I feel, is one more step that naturally happens when someone gets into accessibility. Maybe it will show up in a doc in future.

public/docs/ts/latest/cookbook/a11y.jade
AlmeroSteyn: http://webaim.org/projects/screenreadersurvey6/#used

public/docs/ts/latest/cookbook/a11y.jade
AlmeroSteyn: Agreed. However that is more an A2 doc wide discussion so will align this one to the styles of the other docs for now.

public/docs/ts/latest/cookbook/a11y.jade
mattdistefano: That's what I figured - can you include that in the cookbook though? I think having that linked here would help justify the recommendations. Also - unrelated, but does the doc mention WCAG anywhere?

public/docs/ts/latest/cookbook/a11y.jade
mattdistefano: I think it'd be worth mentioning somewhere. 

Just acknowledging the fact that screen readers do exist for mobile devices would be beneficial, as I think there's a lay perception that screen reader users are always on a "desktop" device/browser/OS. Likewise, if you're building an angular mobile app without a web counterpart, advice about testing on Firefox and IE and whatnot probably doesn't do much for you.

public/docs/ts/latest/cookbook/a11y.jade
AlmeroSteyn: Valid point. Will add mention of this in the next commit. However a whole walkthrough of the implementation is perhaps not the best thing for the cookbook. In fact the cookbook is constantly trying to convince developers to rather use the native elements.

public/docs/_examples/cb-a11y/ts/app/dev-tools/a11y-pass/a11y-pass.component.html
AlmeroSteyn: Will be aligned in the next commit.

public/docs/ts/latest/cookbook/a11y.jade
AlmeroSteyn: Sure! A mention is worth it.

public/docs/ts/latest/cookbook/a11y.jade
AlmeroSteyn: I don't want to add this link to the docs. It is a survey that can be superseded at any stage and therefore adds more maintenance. Like many things in the cookbook we are simply suggesting it without references all over the place. 

As for **WCAG** I honestly dunno if it should be added. The cookbook is about how to do accessibility in Angular and not supposed to be a full handbook on `a11y`. I find the doc extremely overbearing for a new developer for `a11y`. I think a dev learning about `a11y` for the first time from this cookbook will inevitably find his way to the **WCAG**, but I am scared of throwing the doc at everyone seeing it for the first time. I think **webAIM** and **a11yproject** has much more digestible material for the newbie and I'd like them to end up there first.

But I am still open to being convinced. Let me know what you think.

public/docs/ts/latest/cookbook/a11y.jade
AlmeroSteyn: Changed this to IE 10/11 in the next commit. Even though there are good things in the pipeline for Edge, there are too many warnings out there to not use Edge. Including the **Freedom Scientific** site itself. So let's limit to these two for now. The docs will evolve with the field as time passes.

public/docs/ts/latest/cookbook/a11y.jade
AlmeroSteyn: Good idea. Added in the next commit.

public/docs/ts/latest/cookbook/a11y.jade
mattdistefano: Sorry about the late response. I think the wording around the browser support recommendations is probably OK as-is. I just worry about offering that sort of guidance without having a source to back it up. 

Re. WCAG... I think you're right that the material is pretty imposing and could be off-putting. So I don't disagree about not linking to it immediately as another source of information. That said, I almost would like to see WCAG woven a bit more into the cookbook, as I think a lot of people interested in this subject will be looking for info specifically on how to meet WCAG in angular (because of organizational policy, for instance). But that probably has the same risk of alienating newcomers, so maybe it's just something to consider for the future if other folks ask for it.

chalin:
@AlmeroSteyn - e2e tests are failing because the ts compiler is reporting errors. Can you investigate and fix the issues?

cc @naomiblack @wardbell

chalin:
All: @AlmeroSteyn isn't currently available. As agreed discussed with @wardbell via slack: @Foxandxss or @filipesilva could you look into upgrading the code from use of ng beta/rc to 2.1.x?

niceredfrog:
I authored the most recent commit. I'm replying to confirm.

marcysutton:
What do we need to do to get this up to date and merged?

Issue #1628, opened 2016-06-08
Proof read toh-1 to -5


chalin:
Proof reading tutorial: prose, code
 
 - [ ] toh-1 - #1614, #1615.
 - [ ] toh-2 - #1621, #1622.
 - [ ] toh-3 - #1629, #1630.
 - [ ] toh-4 - #1637, #1638.
 - [ ] toh-5 - #1643, #1644.

wardbell:
closed by PRs

chalin:
Extending this issue to also cover toh-6.

PR #1638, opened 2016-06-09
docs(toh-4/ts): copyedits

chalin:
Contributing to #1628.

### DIFF
--- a/public/docs/_examples/toh-4/ts/app/hero.service.ts
+++ b/public/docs/_examples/toh-4/ts/app/hero.service.ts
@@ -12,8 +12,8 @@ export class HeroService {
   getHeroes() {
     return Promise.resolve(HEROES);
   }
-  // #enddocregion get-heroes
-  // #enddocregion just-get-heroes
+  // #enddocregion get-heroes, just-get-heroes
+  // #enddocregion
   // See the "Take it slow" appendix
   // #docregion get-heroes-slowly
   getHeroesSlowly() {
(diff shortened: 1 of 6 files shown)


public/docs/_examples/toh-4/ts/app/hero.service.ts
```diff
@@ -12,8 +12,8 @@ export class HeroService {
   getHeroes() {
     return Promise.resolve(HEROES);
   }
-  // #enddocregion get-heroes
-  // #enddocregion just-get-heroes
+  // #enddocregion get-heroes, just-get-heroes
+  // #enddocregion
```
Foxandxss: Shouldn't this one disappear?

wardbell:
LGTM.  I don't know about @Foxandxss question but if we messed up, we can fix it.

public/docs/_examples/toh-4/ts/app/hero.service.ts
chalin: I've always assumed that we could not specify the unnamed docregion when other docregions are being named. But I could be wrong (I'll test that out some time).

chalin:
No, it would not mess anything up. I think that Jesus was commenting about a potentially unnecessary extra line (i.e., have two `#enddocregion` tags instead of one).

public/docs/_examples/toh-4/ts/app/hero.service.ts
chalin: Had a conversation with @Foxandxss in a separate thread, and he is ok with the change made here. FYI.

PR #1644, opened 2016-06-10
docs(toh-5/ts): minor edits

chalin:
- Adjust text which used to say "create a new 'styles.css'" when we've been carrying it around since QuickStart.
- The HTTP tutorial chapter is no longer "forthcoming".
- Added missing path prefix `app` to code excerpt titles.
- Copyedits.

Contributing to #1628.


Foxandxss:
This one LGTM but I will add a reminder that I need to change your "add style.css to root" because that will no longer true, soon.

PR #1646, opened 2016-06-10
chore(ignored/boilerplate files): minor cleanup

chalin:
(1) We are ignoring _all_ style files including the master styles.css:
```bash
> cd .../_examples
> git ls-files --ignored --exclude-standard | grep 'styles*.css'
styles.css
```
We don't want to do that. Now we only ignore `styles.css` in subdirectories. Cf. #894.

(2) The gulp target `copy-example-boilerplate` is also copying ts boilerplate into dart folders. This is fixed.

(3) Boilerplate files are now chmod -w so as to avoid that they be edited by mistake. Both of the following have been tested after these changes and continue to work as expected:
```bash
gulp add-example-boilerplate
gulp remove-example-boilerplate
```
(4) Entries in `public/docs/_examples/.gitignore` corresponding to boilerplate files have been moved to the top.

(5) Duplicate entries in `public/docs/_examples/.gitignore` relative to the top-level `.gitignore` have been removed.

(6) Entries in `public/docs/_examples/.gitignore` that are global in nature have been moved to the top-level `.gitignore`.


wardbell:
LGTM. I am merging it

I'm prepared to see how life is with 
>Boilerplate files are now chmod -w 

Sometimes I alter them for local purposes. But I see that could be a source of error.  We can always roll this particular change back if it proves too onerous.

chalin:
Btw, `chmod -w` doesn't prevent us from editing the files (I sometimes make local edits to); most modern editors/IDEs like Emacs and VSCode will warn the developer "This file is read-only, do you still want to save your changes?" and save changes to a read-only file if that is what the developer wants.

PR #1656, opened 2016-06-13
chore: add linting to travis

Foxandxss:
I put a lint error to try this.

The idea is travis exiting early if an error is produced.


Foxandxss:
Tested, it works good.

PR #1658, opened 2016-06-13
chore: add section for a11y

Foxandxss:
We plan to put a11y sections through the guides so we make a11y a first class citizen for Angular.

This sections is just a bare copy of the sub-section for now.


Foxandxss:
Ping @naomiblack and @ericjim. We need to make this section different, to say somehow that those sections are a11y things. Not sure if a header on them, different BG color, I don't know.

PR #1663, opened 2016-06-14
api-builder: fix jade formatting of indented constructor code

### LINKED ISSUE #1662: API doc generation broken by angular commit f39c9c9 [style(lint): re-format modules/@angular]
Hey @alexeagle and @ericjim and @petebacondarwin -- this lint change introduced jade errors in the API docs and is preventing us from deploying the site any later than https://github.com/angular/angular/commit/f39c9c9. 

The first error encountered is in `@angular/core/src/metadata/directives.ts` which produces http://localhost:3000/docs/ts/latest/api/core/index/ComponentMetadata-class.html but it's possible this is just the first file we hit with the issue. 

It looks like a bunch of `div` tags that should be output to jade as `.div` are being incorrectly formatted by the template.

screen shot 2016-06-13 at 9 55 41 pm

What's odd is that other API doc pages, e.g. http://localhost:3000/docs/ts/latest/api/core/index/ComponentRef-class.html are not affected. 

I need some help figuring out 1) which dgeni template to fix and 2) how to actually fix it.

Urgent, because we cannot publish RC2 API docs without a fix.

petebacondarwin:
Closes #1662

### DIFF
--- a/tools/api-builder/angular.io-package/templates/class.template.html
+++ b/tools/api-builder/angular.io-package/templates/class.template.html
@@ -33,7 +33,8 @@ div(layout="row" layout-xs="column" class="ng-cloak")
       pre(class="prettyprint no-bg-with-indent")
         a(class="code-anchor" href="#constructor")
           code(class="code-background api-doc-code") {$ doc.constructorDoc.name $}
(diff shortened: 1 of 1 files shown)


naomiblack:
merged and fixed! thank you so much @petebacondarwin

PR #1664, opened 2016-06-14
chore(gulp): avoid watching tmp files

chalin:


chalin:
@Foxandxss : very minor change is ready for review and merge.

Foxandxss:
What is that globbing going to filter?

chalin:
Temporary files generated by Emacs during editing.

chalin:
Actually, I don't think that we should be watching _any_ dot files, but in particular not autosave files. Maybe that line can be changed simply to `**/.?*`. Let me know what you prefer.

chalin:
FYI, I changed it to `**/.?*`.

chalin:
Actually, this still isn't solving my original problem (of getting browser page refreshes when Emacs does an auto-file backup). So I'll close this for now.

PR #1666, opened 2016-06-14
docs(util-fns): generalize getExampleName()

chalin:
For some chapters the Jade source file name (e.g., `toh-pt6.jade`) does not match with the example directory name (`toh-6`). For such cases, we can now set `_example` to, say `toh-6`; `getExampleName()` will now use `_example` if set, otherwise determine the name from `current` as it did before.

### DIFF
--- a/public/_includes/_util-fns.jade
+++ b/public/_includes/_util-fns.jade
@@ -100,10 +100,11 @@ mixin makeExcerpt(_filePath, _region, _title, stylePatterns)
   - if (title) title = title + ' (' + excerpt + ')';
   +makeExample(filePath, region, title, stylePatterns)(format='.')
 
-//- Extract the doc example name from `current`.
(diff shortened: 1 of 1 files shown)


chalin:
@Foxandxss : ready for review and merge.
cc @kwalrath

Foxandxss:
So with this, we can define a _example variable in the .jade so we can do better on the example paths through the jade?

chalin:
Yes. E.g., `toh-pt6.jade` will soon have:
```jade
- var _example = 'toh-6';
```
at the top. Then the functions and mixins in the `_utls-fns.jade` files can uniformly call `getExapleName()` without concern of the means by which the name is obtained.

chalin:
Subsumed by #1686 (since this was a necessary change for toh-6).

PR #1669, opened 2016-06-15
docs(quickstart): add a section to installing typings manually

Foxandxss:
We can add this warning until we can move to TS2 and @types/

/cc @naomiblack


Foxandxss:
On master

PR #1671, opened 2016-06-15
chore: update webpack guide to RC2

Foxandxss:


ericmartinezr:
@Foxandxss just a thought, wouldn't be nice to add `@angular/forms` for those who want to try the new api?

Foxandxss:
We are not using that for the other examples just yet, as soon as we update, I will.

Foxandxss:
On master

PR #1675, opened 2016-06-16
chore: add async/await support to e2e tests

filipesilva:
This PR adds support for async/await usage in our e2e tests. 

To use it you will need to:
- run `gulp add-example-boilerplate` to correctly propagate the `tsconfig.json` changes to all the existing `e2e-spec.ts` folders
- delete `public/docs/_examples/_protractor/typings/` and run `npm run postinstall` on `public/docs/_examples/_protractor`

/cc @chalin @Foxandxss @wardbell

### DIFF
--- a/public/docs/_examples/_protractor/tsconfig.json
+++ b/public/docs/_examples/_protractor/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es6",
     "module": "commonjs",
     "moduleResolution": "node",
     "sourceMap": true,
--- a/public/docs/_examples/_protractor/typings.json
+++ b/public/docs/_examples/_protractor/typings.json
@@ -1,7 +1,6 @@
 {
   "globalDependencies": {
     "angular-protractor": "registry:dt/angular-protractor#1.5.0+20160425143459",
-    "core-js": "registry:dt/core-js#0.0.0+20160317120654",
     "jasmine": "registry:dt/jasmine#2.2.0+20160505161446",
     "node": "registry:dt/node#4.0.0+20160509154515",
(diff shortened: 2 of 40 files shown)


chalin:
Really, that's it? (I wonder why that didn't work for me.) Anyhow, fantastic! I can't wait to start using async/await. The test code will be much more readable when using it!

chalin:
On a separate but related note: will it be possible to use async/await in the sample TS code? Maybe with the arrival of TS 2.0?

wardbell:
I think as soon as TS 2 arrives

filipesilva:
That was mostly it, but the setup was important. Without you would get both code intelligence and compilation errors.

I also forgot to add in something: all specs need to have `'use strict';` to prevent this node error:

```
[01:10:42] [01:10:42] E/launcher - Error: SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
```

Basically, since we are transpiling to ES6, node (not TS) needs the `'use strict';` to allow some ES6 features.

chalin:
Ah, yes, those were the errors I was getting. Glad you caught that too.

filipesilva:
Also had to remove argument destructuring from `public/docs/_examples/animations/e2e-spec.ts`, since node doesn't support it natively.

Foxandxss:
So far so good 👍

naomiblack:
@Foxandxss is this ready to merge?

Foxandxss:
Yes, definitely.

wardbell:
Put that in CLI blueprints?

filipesilva:
Yes we could definitely update the CLI and quickstart with this.

naomiblack:
merged, live.

PR #1680, opened 2016-06-17
fix(e2e): update broken node typings

filipesilva:
Current typings were causing the following error:

```
[13:06:52] [13:06:52] E/launcher - Error: TSError: ⨯ Unable to compile TypeScript
typings\globals\node\index.d.ts (415,11): Interface 'NodeBuffer' incorrectly extends interface 'Uint8Array'.
  Types of property 'fill' are incompatible.
```

/cc @Foxandxss


Foxandxss:
While you are at it, could you please update the typings.1.json that could be there? (webpack)

filipesilva:
Done

PR #1681, opened 2016-06-17
chore: add forms package

filipesilva:
This PR adds the new `@angular/forms` package to angular.io examples.

/cc @Foxandxss


Foxandxss:
And after we propagate this to all examples, no one cries?

filipesilva:
Tried ToH-6 and it was fine.

Foxandxss:
ToH, the ultimate _example where we can try everything.

It is good for me then.

filipesilva:
I'll wait for the router update to do that as well.

filipesilva:
Added in the router, but we need to check tests. I expect some of them to fail.

filipesilva:
Unsurprisingly, `public/docs/_examples/router/ts` failed - but it was the only one, which is good.

@Foxandxss do we have an update in the works?

Foxandxss:
What kind of update?

filipesilva:
Blocked by https://github.com/angular/angular.io/pull/1691

naomiblack:
merged and live!

PR #1688, opened 2016-06-17
docs(toh-5/dart): make dashboard more robust - match TS slice()

chalin:
Originally the dashboard TS expression ``heroes.slice(1, 5))` had been
written as:
> _heroService.getHeroes().getRange(1, 5)

which is brittle; it fails if there are not enough heroes. Slice doesn't fail; an equivalent expression in Dart is
> _heroService.getHeroes().skip(1).take(4)

This is now used.

Other changes:
- Fix in css (missed TS-side update).
- Ran `dartfmt` on `heroes_component.dart`.


chalin:
@kwalrath @thso : ready for review and merge.

kwalrath:
LGTM, but @thso should take a look.

PR #1693, opened 2016-06-18
docs(dynamic-cookbook): upgrade to use new forms api

thelgevold:
I have added a version of the Dynamic Form Cookbook using the new forms API.

### DIFF
--- a/public/docs/_examples/systemjs.config.plunker.js
+++ b/public/docs/_examples/systemjs.config.plunker.js
@@ -7,6 +7,7 @@
 
   var ngVer = '@2.0.0-rc.2'; // lock in the angular package version; do not let it float to current!
   var routerVer = '@3.0.0-alpha.7'; // lock router version
+  var formsVer = '@0.1.0'; // lock forms version
 
   //map tells the System loader where to look for things
   var  map = {
@@ -14,6 +15,7 @@
 
     '@angular':                   'https://npmcdn.com/@angular', // sufficient if we didn't pin the version
     '@angular/router':            'https://npmcdn.com/@angular/router' + routerVer,
+    '@angular/forms':             'https://npmcdn.com/@angular/forms' + formsVer,
     'angular2-in-memory-web-api': 'https://npmcdn.com/angular2-in-memory-web-api', // get latest
     'rxjs':                       'https://npmcdn.com/rxjs@5.0.0-beta.6',
     'ts':                         'https://npmcdn.com/plugin-typescript@4.0.10/lib/plugin.js',
@@ -56,6 +58,9 @@
 
   // No umd for router yet
   packages['@angular/router'] = { main: 'index.js', defaultExtension: 'js' };
+
+  // Forms not on rc yet
+  packages['@angular/forms'] = { main: 'index.js', defaultExtension: 'js' };
 
   var config = {
     // DEMO ONLY! REAL CODE SHOULD NOT TRANSPILE IN THE BROWSER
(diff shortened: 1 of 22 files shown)


public/docs/_examples/systemjs.config.plunker.js
```diff
@@ -56,6 +56,9 @@
 
   // No umd for router yet
   packages['@angular/router'] = { main: 'index.js', defaultExtension: 'js' };
+  
+  //Forms not on rc yet
```
Foxandxss: Nit: space after //

thelgevold: Updated

public/docs/ts/latest/cookbook/dynamic-form.jade
```diff
@@ -32,11 +40,27 @@ include ../_util-fns
...
+<a id="bootstrap"></a>
+:marked
+   ## Bootstrap
+
+   During bootstrap we have to register the new forms module by calling `provideForms()` and pass the result to the provider array. 
+
++makeExample('cb-dynamic-form/ts/app/main.ts','','app/main.ts')   
+
+:marked
+   The old forms API is going through a deprecation phase. During this transition we are supporting both form modules.
```
Foxandxss: This "we" is confusing in here because I don't understand if you mean "Angular" or this concrete example.

We need to differentiate that "we" because two lines after, we mention that we are going to disable the old forms and the "we" there means the example and not angular.

Foxandxss:
I like it.

public/docs/ts/latest/cookbook/dynamic-form.jade
```diff
@@ -32,11 +40,27 @@ include ../_util-fns
...
+   During bootstrap we have to register the new forms module by calling `provideForms()` and pass the result to the provider array. 
+
++makeExample('cb-dynamic-form/ts/app/main.ts','','app/main.ts')   
+
+:marked
+   The old forms API is going through a deprecation phase. During this transition Angular is supporting both form modules. 
+
+   To remind us that the old API is deprecated, Angular will print a warning message to the console. 
+   
+   Since we are converting to the new API, and no longer need the old API, we can call `disableDeprecatedForms()` to disable the old form functionality and the warning message.
```
Foxandxss: I read better "we call" than "we can call" but that could be just me.

Foxandxss:
This is a bit of a nitpick, but could you please put your imports in alphabetic order and leave an space between angular imports and your own imports? You are doing that for almost all files but a few :)

Thanks!

thelgevold:
Yep. I just pushed better ordering of imports

public/docs/_examples/systemjs.config.plunker.js
```diff
@@ -56,6 +56,9 @@
 
   // No umd for router yet
   packages['@angular/router'] = { main: 'index.js', defaultExtension: 'js' };
+  
+  // Forms not on rc yet
+  packages['@angular/forms'] = { main: 'index.js', defaultExtension: 'js' };
```
filipesilva: I think it will also need changes above in the mapping, the same treatment as the router (locking the version and mapping with it).

public/docs/ts/latest/cookbook/dynamic-form.jade
thelgevold: Updated

public/docs/_examples/systemjs.config.plunker.js
thelgevold: Updated

PR #1691, opened 2016-06-18
docs(router): Updated docs to reflect alpha 7 release

brandonroberts:
cc @filipesilva

PR #1695, opened 2016-06-19
Update route docs to version 7 & fix linting issues (closes #1691)

iamarcel:
I took the liberty to fix the linting issues form #1691, @brandonroberts' PR.

### DIFF
--- a/public/docs/_examples/package.json
+++ b/public/docs/_examples/package.json
@@ -31,7 +31,7 @@
     "@angular/http": "2.0.0-rc.2",
     "@angular/platform-browser": "2.0.0-rc.2",
     "@angular/platform-browser-dynamic": "2.0.0-rc.2",
-    "@angular/router": "2.0.0-rc.2",
+    "@angular/router": "3.0.0-alpha.7",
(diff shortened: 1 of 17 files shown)


Foxandxss:
Yes, but this duplicates his PR again instead of fixing the concrete error.

If you don't mind, I am going to close this and let him fix it. I need the authors to be able to lint their solutions too.

Foxandxss:
Oh and thanks tho, I like the initiative like this, but would be problematic to merge if both PRs contain the same changes but a bit of difference.

iamarcel:
Sure, no problem! I thought it looked like a bit of a mess already ;)
I'll send my PR to Brandons' fork (so that my hard 3 minutes' work isn't wasted, right? :stuck_out_tongue_winking_eye:).

Foxandxss:
That wouldn't do any good, so instead of 3 mins, it would be 5 :P

Let him fix the lint errors. I want each author here to get used to lint their solutions.

iamarcel:
Dammit! But you've definitely got a point. I'll wait until the day I can fix my own linting issues then ;)

PR #1699, opened 2016-06-20
fix(quickstart): add styles.css to js quickstart

### LINKED ISSUE #1684: Quickstart should have a step to add styles.css
Although this is really tangential to the quickstart, we should have a step where we add it so the user doesn't get an error.

migrated issue from https://github.com/angular/angular/issues/9307 @fadiwissa

---

Current behavior
The quick start guide is missing the Styles.css file. This file is referenced but no where in the "walkthrough" is it described. I got it from the Plunker version of the demo.

Expected/desired behavior
The Styles.css file should be created as one of the steps of the walkthrough.

Reproduction of the problem
Just walk through the walkthrough steps and the browser will throw an error due to its inability to find the referenced Styles.css file (referenced in Index.html)

What is the expected behavior?
The file should simply exist and no error should be thrown.

filipesilva:
Fix #1684

While looking at this, I saw that plunkers were getting some unintended unit testing files and added them to the default plunker builder ignore.

### DIFF
--- a/public/docs/_examples/quickstart/js/styles.1.css
+++ b/public/docs/_examples/quickstart/js/styles.1.css
@@ -1,0 +1,14 @@
+/* #docregion */
+h1 {
+  color: #369;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 250%;
+}
+body {
+  margin: 2em;
+}
+
+ /*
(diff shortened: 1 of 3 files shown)


naomiblack:
merged, live.

PR #1709, opened 2016-06-21
chore(style-guide): remove routing styles

filipesilva:
Routing styles were breaking CI and are out of date, since we're using the 3.0.0 router.

We should remove them until we have recommendations.

/cc @Foxandxss @johnpapa @mgechev


Foxandxss:
Yes, I recommended Brandon to create a several new styles and also check if this one was valid. When I went to bed, I realized that this style wouldn't even allow us to go live, and since it is not relevant right now, I vote to delete it.

mgechev:
As far as I saw, the snippets are not used anywhere else so it seems safe to remove them.

naomiblack:
merged and live.

PR #1710, opened 2016-06-21
chore: update ngSwitchWhen to ngSwitchCase

filipesilva:
See https://github.com/angular/angular/commit/e1fcab7

/cc @Foxandxss


thelgevold:
@filipesilva I forgot to update the switchWhen in the deprecated version of the dynamic forms https://github.com/angular/angular.io/blob/master/public/docs/_examples/cb-dynamic-form-deprecated/ts/app/dynamic-form-question.component.html

filipesilva:
@thelgevold updated the PR with it.

PR #1718, opened 2016-06-21
fix(doc): fix 404 problem of some api url

Priestch:


Priestch:
I signed it!

naomiblack:
thank you! merged, live!

Issue #1723, opened 2016-06-22
Refer to 'lite-server' instead of 'live-server' in the testing documentation


fourcube:
The quickstart uses `lite-server` for running a development server. The testing documentation refers to reading up on `live-server`. 

I'd be willing to make the appropriate changes to the docs.

Foxandxss:
Please do.

PR #1737, opened 2016-06-23
docs(testing): refer to 'lite-server' instead of 'live-server'.

fourcube:
Since the QuickStart uses 'lite-server', the 'lite-server' module should also be used in the testing guide.


fourcube:
I have just signed the CLA.

(implements #1723)

Foxandxss:
Appreciated, but I don't advice spending time on the testing guide because it is going to be revamped soon.

wardbell:
Accepting just so I have latest state before rewrite.

PR #1754, opened 2016-06-26
chore(install): put all install commands into a single script

chalin:
This PR makes it easier to get a local repo back into a sane state after a change in typings (something that a few of us have had some challenges with recently :). Specifically, this PR addresses two things.

(1) On June 18th, @filipesilva made the following comment on the docs-authoring slack channel:
> in regards to setup, if you're ever uncertain just check `travis.yml`. It's the definite resource

I agree with this approach of relying on the travis config file. Furthermore, I think that it should be easy to manually create a clean repo (when cloning afresh isn't an option). This PR puts the 5 travis install commands into a single script: `./script/install.sh`. Note that the `angular` team has a similarly named script that they also use for travis installs.

With this PR, one can get a refreshed repo by issuing the following commands from the repo dir:
```bash
git clean -xdf
./script/install.sh
```

(2) I changed two of the install commands as follows:
```diff
  npm install --no-optional
- npm install --prefix public/docs/_examples
- npm install --prefix public/docs/_examples/_protractor
+ (cd public/docs/_examples && npm install)
+ (cd public/docs/_examples/_protractor && npm install)
  npm run webdriver:update --prefix public/docs/_examples/_protractor
  gulp add-example-boilerplate
```

While travis builds seemed to be working, I'm a bit puzzled as to why the two `npm install --prefix foo` commands gave the desired effect. It is my understanding that such invocations of `npm install` would use the top-level `package.json` file (which I don't think is what we wanted). Anyhow, these two `npm install --prefix foo` commands **did not** work when run on a local repo. The adjusted commands work both locally and for travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,6 @@ before_install:
 before_script:
   - sh -e /etc/init.d/xvfb start
 install:
-  - npm install --no-optional
-  - npm install --prefix public/docs/_examples
-  - npm install --prefix public/docs/_examples/_protractor
-  - npm run webdriver:update --prefix public/docs/_examples/_protractor
-  - gulp add-example-boilerplate
+  - ./script/install.sh
 script:
   - gulp $SCRIPT
--- a/script/install.sh
+++ b/script/install.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -ex -o pipefail
+
+npm install --no-optional
+(cd public/docs/_examples && npm install)
+(cd public/docs/_examples/_protractor && npm install)
+npm run webdriver:update --prefix public/docs/_examples/_protractor
+gulp add-example-boilerplate


chalin:
@filipesilva @Foxandxss : as you may recall I had a broken local environment before I went away last week. Now things are back in order, after following the setup in the travis config file (thanks for the suggestion @filipesilva). This PR just makes it easier to do so (since I don't expect that we have seen the last of a change in typings :).

Btw, if you agree with the changes proposed in this PR, then maybe we should update the [README.md](../blob/master/README.md) to reflect this new way of setting up / refreshing a local repo. We could then also add that no PR should be submitted unless `lint` is happy and all `e2e` tests pass.

Foxandxss:
Yes, something like that should be done. We will talk about that tomorrow I guess.

chalin:
Travis is all green.

Strangely, when I run the tests locally, I get two failures:
```
Suites failed:
  public/docs/_examples/pipes/ts
  public/docs/_examples/user-input/ts
```
But when I perform the failed test cases manually, I observe the desired behavior. Maybe the tests are failing because protractor isn't waiting for the actions to complete?  (Btw, the same tests pass in Dart.)  Any ideas why the tests are failing locally?

Foxandxss:
Those two failing tests are a known issue, completely unrelated to your changes. Forget about them for now, they are an issue on chrome and/or protractor.

filipesilva:
As far as I can tell, the `--prefix` command doesn't work on windows or something. I tried it a couple of times and really don't get how it works, but according to the npm docs it should.

But, on the other hand, neither do shell files work on windows (without git bash). So I'm not sure this PR helps much in that sense.

If it is to go in, then the `webdriver:update` command should also be changed to not use `--prefix`.

chalin:
@Foxandxss : thanks for confirming that those are known failures (FMI, is there an issue open to track this)?

@filipesilva : thanks for clarifying the situation under Windows. Note that this PR does help macOS and linux folks :) ... without making things any worse under Windows, as far as I understand.

chalin:
@filipesilva : FYI, `--prefix` is needed for the `webdriver:update` command.

filipesilva:
@chalin so the prefix thing doesn't work on osx and linux either? That I didn't expect.

For the webdriver command, why can't it be done in the same scheme as the others? Does it not work either? `(cd public/docs/_examples/_protractor && npm run webdriver:update)  --prefix `

chalin:
The `webdriver:update` _can_ be set in the format of the others, but it doesn't _need_ to.

When used with `install`, the "`--prefix `_path_" causes `node_modules` to be output under _path_, but I think it still picks up the `package.json` from the local directory (at least that is the behavior I saw under macOS ... and the reason why I changed the `install` commands.

filipesilva:
Got it. Didn't fully understand that before.

kwalrath:
fwiw, I just used this script, and now my repo is the happiest it's been in weeks (I'm able to deploy the docs again).

Foxandxss:
I agree. We do crazy changes to the repo and be able to restart it to brand new condition without losing the branches (the problem of a fresh copy) is 200% needed.

filipesilva:
Let's go for it then :D

PR #1758, opened 2016-06-27
Restyle resources page with new schema

ericjim:
### Changes
* Make a new resources page

### Preview
* https://new-schema.firebaseapp.com/resources/

@StephenFluin

### DIFF
--- a/public/_data.json
+++ b/public/_data.json
@@ -31,21 +31,5 @@
 
   "presskit": {
     "title": "Press Kit"
-  },
-
-  "books": {
-    "title": "Books"
-  },
-
-  "training": {
-    "title": "Training"
-  },
-
-  "communities": {
-    "title": "Communities"
-  },
-
-  "tooling": {
-    "title": "Tooling and Libraries"
   }
 }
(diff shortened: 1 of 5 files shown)


public/_data.json
```diff
@@ -31,21 +31,5 @@
 
   "presskit": {
     "title": "Press Kit"
-  },
```
ericjim: Originally we believed these would be separate pages, eventually we consolidated everything. So these pages are no longer needed.

naomiblack:
I'll need a bunch of data populated before I'm happy merging to master. Right now most of the entries show up as _No Description_

StephenFluin:
@naomiblack I can have that in tonight. Plan on merging it Thursday?

ericjim:
@StephenFluin thanks for pointing that out, I'll fix it.

ericjim:
@StephenFluin I've made an update, you can see it here now https://new-schema.firebaseapp.com/resources/ , make sure to do a hard refresh. I don't know what to do with the nav in smaller screens. So I am hiding it when the screen real estate gets tight. We can revisit the design later, or we can gather suggestions.

StephenFluin:
@ericjim This fixes the usability and think it's good enough. We can revisit further. @naomiblack content is in, so this might be ready to merge

PR #1775, opened 2016-06-29
chore(examples): revert back to stable TS

filipesilva:
The new router once required TS1.9 but it doesn't anymore. It's best to revert our examples to TS 1.8 since that is what we tell users to install, and we want to be testing on the current setup.

@brandonroberts @Foxandxss @wardbell


Foxandxss:
Yes please, I was getting some weird behaviors with this.

naomiblack:
merged and live!

PR #1791, opened 2016-06-30
docs(router): explicit explaination of terminal property

rexebin:


Foxandxss:
I think I like this change.

PR #1797, opened 2016-06-30
docs: Typo (I think?), third > second

patrickmcd:
Unless I'm confused, isn't the id the second item in the array?


wardbell:
Actually, the situation is much worse than you describe. 

We kept this section mostly intact when we translated from the deprecated router to the v.3 router. Unfortunately, the new version only has a one element parameter array and therefore **doesn't illustrate the possibilities of a multi-valued parameters array at all.**

We need to rewrite/rethink this entire section.  See issue #1868

PR #1799, opened 2016-06-30
docs: replace terminal with pathMatch

Foxandxss:
This went deprecated so here is the update.


wardbell:
Fixed text and also replace `<base href=".">` with `<base href="/">` because (and I should have discovered this earlier.

PR #1801, opened 2016-07-01
docs(toh-5): use `Array.prototype.find()` instead of `filter`

kutyel:
We are in *TypeScript*, so we can avoid worrying about ECMAScript methods being implemented or not (as `find()` is already implemented in TypeScript) and, personally, my eyes hurt when I see a `filter()` returning its first element `[0]` without checking if it returned **something** before.

### DIFF
--- a/public/docs/_examples/toh-5/ts/app/hero.service.ts
+++ b/public/docs/_examples/toh-5/ts/app/hero.service.ts
@@ -20,7 +20,7 @@ export class HeroService {
   // #docregion get-hero
   getHero(id: number) {
     return this.getHeroes()
-               .then(heroes => heroes.filter(hero => hero.id === id)[0]);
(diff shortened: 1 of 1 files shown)


wardbell:
I agree. And many thanks!

But why stop with this file? Fix 'em all. That's what I did with #1816 which supersedes your PR here.

kutyel:
Thanks @wardbell! I agree with you as well. 

Nevertheless, I really would have liked to make the change myself, to feel that I was *doing my bit...* :wink:

wardbell:
Oh. Sorry. But you really did the most important thing by calling it out. The rest took me 15 minutes. Thanks again.

kutyel:
Ok, don't worry! Your welcome! :smile:

PR #1817, opened 2016-07-03
Correct sample reference for milestone #2

rexebin:
We have received an issue at angular-cn: [issue #13](https://github.com/angular/angular-cn/issues/13) thanks to [gogobook](https://github.com/gogobook).

1. in the final code samples of milestone 2, `providers:[HeroService]` is missed in `app.component`
2. `crisis-center` folder does not exist yet at milestone 2, therefore reference to `app.routes.ts` is not correct.


Foxandxss:
This got fixed in RC5 changes.

PR #1821, opened 2016-07-04
chore: update node and npm recommendations

Foxandxss:
So node works fine, what we need is to ensure that people uses npm 3.x or more.


wardbell:
Should we warn against node v.6? Or does that only fail for us?

Merging anyway. We can warn against 6 in a separate PR if necessary.

PR #1822, opened 2016-07-04
chore: add support for multiline annotations in the API

Foxandxss:
All the kudos to @chalin for finding this problem.


Foxandxss:
Oh, we need this because we have the first multiline annotation with BrowserModule.

chalin:
LGTM!

PR #1825, opened 2016-07-04
docs(webpack): add angular2-template-loader

Foxandxss:
This loader allows us to use Webpack a bit closer of what System.js does and what the CLI is going to do.

### DIFF
--- a/public/docs/ts/latest/guide/webpack.jade
+++ b/public/docs/ts/latest/guide/webpack.jade
@@ -233,9 +233,10 @@ a(id="common-configuration")
 
 :marked
   * ts - a loader to transpile our Typescript code to ES5, guided by the `tsconfig.json` file
+  * angular2-template-loader - loads angular components' template and styles
   * html - for component templates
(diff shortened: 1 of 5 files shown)


public/docs/ts/latest/guide/webpack.jade
```diff
@@ -441,9 +446,9 @@ p.
   The `HtmlWebpackPlugin` inserts them dynamically at runtime.
     
   * The `AppComponent` in `app.component.ts` imports the application-wide css with a simple `import` statement.
-  
-  * The `AppComponent` itself has its own html template and css files which we load with the `require()` method
-  supplied by Webpack. Webpack stashes those component-scoped files in the `app.js` bundle too.
+
+  * The `AppComponent` itself has its own html template and css file. They are being load using `require()` behind
```
bampakoa: typo: They are being load -> They are loaded ?

wardbell: I'll fix that as I merge. Thanks for catching it.

PR #1827, opened 2016-07-05
fix ngOnInit method from tutorial

EmmN:
screen shot 2016-07-05 at 13 00 20

In tutorial i was instructed to call `this.getHeroes();`, but `getHeroes` method is defined in service not on component.


Foxandxss:
Please, don't create a PR without description. I would like to know why do we want this change.

EmmN:
I signed it!

Foxandxss:
It is defined in the component too.

EmmN:
It seems you are right and I missed this part:

screen shot 2016-07-05 at 16 59 45

I will close this PR.

PR #1833, opened 2016-07-05
chore: dgeni tag definitions for ts2dart tags

chalin:
New `dgeni` tag definitions for `ts2dart` tags which are currently causing spurious warnings when generating API docs.

### DIFF
--- a/tools/api-builder/angular.io-package/tag-defs/Annotation.js
+++ b/tools/api-builder/angular.io-package/tag-defs/Annotation.js
@@ -1,0 +1,4 @@
+// A ts2dart compiler annotation that can be ignored in API docs.
+module.exports = function() {
+  return { name: 'Annotation', ignore: true };
+};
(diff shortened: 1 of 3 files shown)


chalin:
@Foxandxss @wardbell @kwalrath : ready for review.

tools/api-builder/angular.io-package/tag-defs/Annotation.js
```diff
@@ -0,0 +1,4 @@
+// A ts2dart compiler annotation that can be ignored in API docs.
+module.exports = function() {
+  return { name: 'Annotation' };
```
petebacondarwin: I am pretty sure that, in addition you can also add an ignore flag:

```
{ name: 'Annotation', ignore: true };
```

petebacondarwin: See https://github.com/angular/dgeni-packages/blob/master/jsdoc/processors/parse-tags.js#L92

petebacondarwin:
Apart from additional `ignore` flag, this LGTM

wardbell:
Merging on basis of @chalin and @petebacondarwin good credit ;-)

PR #1836, opened 2016-07-06
chore(glossary/ts): redirect old glossary to correct location

wardbell:
Redirect old glossary to correct location within the guide folder. This addresses the problem with bad fragments for people who are still navigating here from an old link.

Difficult to test because no firebase redirects work with CheckDeploy. Need to at least have a test server.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -20,7 +20,13 @@
     },
     {
       "source": "/docs/ts/latest/guide/setup.html",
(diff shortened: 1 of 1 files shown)


wardbell:
I think Naomi reverted this because mistaken propert `"type": 301`.

Should re-do the intent properly in a future PR

PR #1839, opened 2016-07-06
chore: update testing example to RC4

Foxandxss:
This updates the example to RC4


Foxandxss:
This is LGTM, only needing the setInterval stuff.

PR #1842, opened 2016-07-06
chore(linkChecker): fix to checker config

chalin:
This PR eliminates false positives by adjusting the checker config options:
- Some sites don’t support `HEAD` so use the `GET` method.
- Exclude known problematic URLs from being checked.


chalin:
cc @kwalrath @Foxandxss @wardbell : with this change, and the latest PRs that fix the broken links, only two broken links remain (as reported elsewhere). Ready for review and merge.

wardbell:
LGTM

PR #1844, opened 2016-07-07
chore: make animation test wait a bit longer

Foxandxss:
That particular test is a bit slower than the rest (animation takes more time) so I bump its sleep a bit more so it doesn't fluke.


Foxandxss:
After a lot of runs, this seems to be working perfect.

PR #1848, opened 2016-07-07
docs(style-guide): spell const variables in lower camel case

wardbell:
Changes previous guidance on const which insisted on UPPER_SNAKE_CASE


wardbell:
Merged by 0664a271

PR #1849, opened 2016-07-08
chore(api-builder): support API doc language agnostic URIs

chalin:
This PR is in parts (and best explained and reviewed that way, in part because of the rename):

**chore(api-builder): rename linkDevGuide to linkDocs**

Rename the specialized API docs `@linkDevGuide` tag to `@linkDocs` since it is meant to be used to refer to any angular.io chapter, not just the developer guide.

**chore(api-builder): let linkDocs links resolve to proper language**

The old linkDevGuide required URIs to be given in the form 'ts/latest/guide/architecture' which forced all language versions of the API pages to refer to the ts versions of the doc pages they linked to.

This PR fixes that problem. URIs are now given in the form 'guide/architecture' and then rendered properly for each language. Of course, when needed, it is still possible to give a language specific URI.


wardbell:
I will squash and merge (no one cares about the steps)

PR #1853, opened 2016-07-08
fix text to match code for HeroListComponent

jamesmanning:
current text refers to importing the ActivatedRoute service, but the actual code imports Router instead.  This text was added as part of the v.3 router doc update in https://github.com/angular/angular.io/commit/c61d8195f3b63c3e03bf2a3c12ef2596796c741d


Foxandxss:
Thanks but I believe that the text is correct now.

jamesmanning:
Agreed, the current version is correct.

PR #1854, opened 2016-07-08
style-guide(add-a11y): Add accessibility quick wins

AlmeroSteyn:
Adding accessibility recommendations to style guide.

These are all quick accessibility wins making use of HTML and CSS alone and requiring no use of ARIA.


public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -1870,6 +1871,280 @@ a(href="#toc") Back to top
...
+
+.s-why.s-why-last
+  :marked
+    **Why?** This enables styling and font selection based on the page language.    
+
+:marked
++makeExample('style-guide/ts/10-01/app/index.avoid.html', 'page-lang', 'index.html')(avoid=1)
+
+:marked
+  Provide the ISO language code as well the country code (where appropriate).
```
Foxandxss: We are not putting messages between avoid and do (correct me if I am wrong). We could put that as a **Consider**

Foxandxss: Seems like we do. Let's hold on this and see if we can do better.

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -1870,6 +1871,280 @@ a(href="#toc") Back to top
...
+:marked
++makeExample('style-guide/ts/10-03/app/app.component.html', '', 'app.component.html')
+
+.l-main-section
+:marked
+  ### <a id="10-04"></a>Always support and test keyboard navigation.
+  #### <a href="#10-04">Style 10-04</a>
+.s-rule.do
+  :marked
+    **Do** unplug your mouse and test if al functionality on your page can be reached and activated with `TAB`, `SHIFT+TAB`, `ENTER`, `SPACE` and the arrow keys alone.
```
Foxandxss: all, typo.

AlmeroSteyn: Fixed.

public/docs/ts/latest/guide/style-guide.jade
```diff
@@ -1869,7 +1870,334 @@ a(href="#toc") Back to top
...
+  :marked
+    **Do** keep the `outline` css property intact.
+
+.s-rule.consider
+  :marked
+    **Consider** only removing it if absolutely required and an alternate `:focus` style is provided.
+
+.s-rule.avoid
+  :marked
+    **Avoid** totaly removing the focus outline on your page.
```
Foxandxss: That is totally a typo.

Foxandxss:
I like it, but it needs a bit of @wardbell magic on the English 👍 

/cc @johnpapa.

public/docs/ts/latest/guide/style-guide.jade
AlmeroSteyn: Now it's completely fixed.

AlmeroSteyn:
@Foxandxss: There is one flow issue now, because of number `10` being reserved for the router styles. **Accessibility** now shows as `10` on the menu above and then you end up in section `11`. Will fix itself when the router styles are added.

PR #1856, opened 2016-07-08
docs(style-guide): remove routing from TOC

Foxandxss:
It is a bit misleading having an entry in the TOC with no styles for it.


wardbell:
We WILL have style guidelines for it soon. Do you want to add it back when we do?

Foxandxss:
For the time being I feel like they confuse the end user. They click the link, nothing happen, they search for the content, it is not there.

We can add it back when we create the first style.

PR #1857, opened 2016-07-08
docs(style-guide): create a route redirection

Foxandxss:
Without one, you get some nice errors.


bampakoa:
@Foxandxss I have also notice it in a demo application that I am working on. I do not know if we mean the same error but I suppose so.When I add a router-outlet in the main component, I get an error stating that I have not declare a default route.If I add a redirect in my routes, the error is gone.Is that reasonable?

Foxandxss:
Different thing I guess.

bampakoa:
Ok, sorry about the irrelevance of the topic.

PR #1858, opened 2016-07-08
change var keyword to const for HEROES constant

thomasjshafer:
HEROES is a `const` and is defined as a `const` originally: https://github.com/angular/angular.io/blob/master/public/docs/_examples/toh-2/ts/app/app.component.ts#L10


thomasjshafer:
I signed it!

wardbell:
LGTM

PR #1864, opened 2016-07-11
fix(package.webpack.json): add angular2-template-loader

SekibOmazic:
Add missing angular2-template-loader dependency


Foxandxss:
Ouch, thanks.

PR #1867, opened 2016-07-11
[WIP] chore(ci): add check-deploy to CI

filipesilva:
This PR adds CI entries to the existing matrix that run `gulp check-deploy` for the latest release and for the master branch of angular.io.

### DIFF
--- a/scripts/examples-install.sh
+++ b/scripts/examples-install.sh
@@ -2,7 +2,6 @@
 
 set -ex -o pipefail
 
-npm install --no-optional
 (cd public/docs/_examples && npm install --no-optional)
 (cd public/docs/_examples/_protractor && npm install --no-optional)
 npm run webdriver:update --prefix public/docs/_examples/_protractor
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,20 +10,24 @@ env:
     - DBUS_SESSION_BUS_ADDRESS=/dev/null
     - DISPLAY=:99.0
     - CHROME_BIN=chromium-browser
+    - LATEST_RELEASE=2.0.0-rc.4
   matrix:
-    - SCRIPT=lint
-    - SCRIPT="run-e2e-tests --fast"
-    - SCRIPT="run-e2e-tests --fast" PREVIEW=true
+    - TASK=lint
+    - TASK="run-e2e-tests --fast" SCRIPT=examples-install.sh
+    - TASK="run-e2e-tests --fast" SCRIPT=examples-install-preview.sh
+    - TASK=harp-compile SCRIPT=deploy-install.sh
+    - TASK=harp-compile SCRIPT=deploy-install-preview.sh
 matrix:
   fast_finish: true
   allow_failures:
-    - env: "SCRIPT=\"run-e2e-tests --fast\" PREVIEW=true"
+    - env: "TASK=\"run-e2e-tests --fast\" SCRIPT=examples-install-preview.sh"
+    - env: "TASK=harp-compile SCRIPT=deploy-install-preview.sh"
 before_install:
   - npm install -g gulp --no-optional
+install:
+  - npm install --no-optional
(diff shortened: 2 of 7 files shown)


scripts/examples-install.sh
```diff
@@ -7,3 +7,4 @@ npm install --no-optional
 (cd public/docs/_examples/_protractor && npm install --no-optional)
 npm run webdriver:update --prefix public/docs/_examples/_protractor
 gulp add-example-boilerplate
+sh -e /etc/init.d/xvfb start
```
Foxandxss: Problem with this is that I couldn't use this script anymore on my machine to install the repo, right? Couldn't we leave this on the travis file?

filipesilva: Sure, didn't know that was an issue but it makes sense.

filipesilva:
This PR will require that the latest release be updated on the `LATEST_RELEASE` travis env global as part of the release process for new angular.io updates.

filipesilva:
@chalin can you review and tell me if this is what you intended?

I'm talking with @naomiblack for a way around the `LATEST_RELEASE` variable, but for now it works.

chalin:
Great to see this move forward!  Please see my comments below.

-----
The `check-deploy --dry-run` essentially just does a harp compile. Defining and using a simpler target might make sense (in fact, I've been using such a target in my Dart API work):
```js
gulp.task('check', ['build-docs'], function() {
  return harpCompile();
});
```

Note that **`scripts/install.sh`** is now part of our instructions to developers on how to install/reset their workspace. Could you ensure that you keep a copy of this file?

Actually, I think that we can simplify the situation by having a single `install.sh` script with conditional logic inside (defaulting to the install steps we had before). To accomplish this we might need a `matrix` something like this:
```yml
env:
  ...
  matrix:
    - TASK=lint
    - TASK='e2e --fast' PREVIEW=false
    - TASK='e2e --fast' PREVIEW=true
    - TASK='check' PREVIEW=false
    - TASK='check' PREVIEW=true
matrix:
  ...
  allow_failures:
    - env: "TASK='e2e --fast' PREVIEW=true"
    - env: "TASK='check' PREVIEW=true"
...
install:
  - ./scripts/install.sh
```
Then `scripts/install.sh` could have conditional logic based on the value of `TASK`. This is just a suggestion, but having a single `install.sh` file also makes it easier to assess the installation variants.

Btw, eventually we may also want to add checks on the output produced by the harp compile command to help determine if the `check` was truly successful or not.

Looking forward to having this in master.

filipesilva:
@chalin thanks for the feedback! I think the simpler task for harp seems like a good idea, I'll put that in.

I don't think having a simple script with conditionals inside is a good idea though. It might default to what `install.sh` did, but that's not a one size fit all. In fact, that `install.sh` was only part of the install procedure - the bit that prepared the environment for authors (and only TS/JS ones I suspect).

That is why I renamed it to `examples-install.sh`. The only difference from before is that it doesn't do a base `npm install --no-optional`, which I agree we should add back for simplicity's sake.

I see the CI being expanded to also running dart if possible, but even in that case it's still useful to have the standalone scripts that setup your environment for what you need instead of having to setting flags locally to guide an install script.

filipesilva:
Ah, I remember why I left `npm install --no-optional` out of `examples-install.sh`... because linting the project doesn't need the rest of the install procedure, but needs `npm install --no-optional`.

I'll update the readme to reflect that.

filipesilva:
Updated the PR with a dedicated gulp task and updated readme.

chalin:
LGTM

filipesilva:
@naomiblack this PR still missing the `latest` release tag in `angular/angular` but it good otherwise. After it's in, we should look at triggering `angular/angular.io` CI on master when there is a `angular/angular` build.

cc @IgorMinar

wardbell:
Merging because @chalin and @filipesilva say so. Presumably they'll fix a busted CI.

PR #1872, opened 2016-07-12
chore: update to @types

filipesilva:
This PR updates all examples to use `@types/*`. It also updates all chapters that mention `typings` to talk about `@types/*` instead.

The `cli-quickstart` example was removed from tests though, since the version of TS it uses is stuck in 1.8.x right now and would fail due to lack of typings.

/cc @wardbell @Foxandxss


filipesilva:
`@types/angular@1.5.13` fixes the issue in `DefinitelyTyped/DefinitelyTyped#10082`, so this PR can be updated and should pass CI.

Foxandxss:
This needs update on other places were we pull typescript 1.x (like systemjs for plunkers and perhaps more places)

filipesilva:
@Foxandxss updated plunker generator and a few other places in the files provided in https://github.com/angular/angular.io/pull/2227.

Also searched for `1.8.` and `1.9.0-dev` throughout the project, don't seem to have found any more references to pre 2.0 TS.

wardbell:
Ugh. I agreed that the prose needed work. I updated and force pushed and now you can't see what I wrote. Here it is:
>#### Why declare _angular_ as _any_?
>A strongly typed angular reference to Angular 1 would be great. But we can't import its UMD typings library, `@types/angular`, without also importing Angular 1 itself via `import * as angular from 'angular'`.
>
>Angular 1 is currently loaded by a script tag in `index.html` and switching to an ES6 import at this time is not worth the considerable effort. Instead we declare `angular` as an untyped `any` to avoid typing errors.

Foxandxss:
Yes please. I need it to work on some other stuff tomorrow 👍

filipesilva:
I agree with the edits, it looks better and addresses the problems that Jesus highlighted. Much better.

Issue #1593, opened 2016-06-03
Enable API doc generation for Dart - quickfix


chalin:
This issue tracks progress in generating Dart API docs using the current angular.io tool chain. This approach has the advantage of being quick to get up and running, but its main disadvantage (at the moment) is that it is generated from the TS sources.

@kwalrath's task list from #1589 
- [x] Make sure the issue search uses the right labels ('api' & 'dart')
- [ ] Add known issues to the database.
- [ ] Add text to each page linking to known issues.
- Either
  - add Dart examples everywhere or 
  - [ ] change the text to NOT imply that we have them everywhere.
- [ ] Make sure the output is useful (which maybe it will be once Dart examples are everywhere?). I didn't realize that all the generated entries in the docs are directly from TS, not from generated Dart.
- [ ] Be more specific in the caveat about how TypeScripty the docs are. Perhaps create a page telling people how to figure out what the API actually is? E.g., how do you find out all the args to bootstrap()? Array -> List, string -> String, boolean -> bool etc.
- [ ] Consider pointing people to both the angular.io and dartdocs.org docs? E.g. https://www.dartdocs.org/documentation/angular2/latest/angular2.common/Control-class.html seems to have correct type info except in methods, which lack types (it also shows lots of inherited properties); https://angular.io/docs/dart/latest/api/common/index/Control-class.html is much shorter and/but has little useful information except the text.

chalin:
Once #1589, is in place, here is what we will be able to do:
- Generate the same API entries that exist on the TS side
- Each API page corresponds to a generated `.jade` file.

The Jade source files are setup much like the prose, and so we have access to language-specific "macros"; e.g. making the following change in [angular/common/src/pipes/async_pipe.ts][ap.ts]
```diff
- * The `async` pipe subscribes to an `Observable` or `Promise` and returns the latest value it has
+ * The `async` pipe subscribes to !{_an} `!{_Observable}` or `!{_Promise}` and returns the latest value it has
```
renders as before on the TS side and on the Dart side we get
> The `async` pipe subscribes to a `Stream` or `Future` and

With **no other change to the API source**, we also get `@example` entries like this:
> {@example core/pipes/ts/async_pipe/async_pipe_example.ts region='AsyncPipePromise'}

correctly picking up Dart samples. (This is because such directives become `makeExample` entries which have been adapted to reinterpret paths based on the page language.)

[...]

PR #1888, opened 2016-07-14
docs(api/dart): add support for generation and display

### LINKED ISSUE #1880: Dart API docs: add support for generation and display
Supersedes #1593.
### Task breakdown

**Package dependencies:**
- [ ] [**BLOCKING**] Update to new [harp](https://github.com/sintaxi/harp) package which depends on latest [terraform](https://github.com/sintaxi/terraform) release. This is needed so that dots (`.`) can be supported in site URIs; something that is essential for Dart given that library names contain dots. See https://github.com/sintaxi/harp/pull/526 (/cc @sintaxi).<br>Blocked awaiting the new release.

**Gulp tasks / build scripts:**
- [x] New tasks to run `pub` and `dartdoc` over (existing) `angular-dart` repo, peer to ng.io repo.
- [x] `build-api-docs` should also do Dart API docs.
- [x] Shredder should run over Dart sources and generate example fragments.
- [x] Create script to patch local `terraform` package, while we await new harp release, as a means of **unblocking**.

**Generator package:**
- [x] New `tools/dart-api-builder`; stand alone tool re-using part of the basis of a Dart API dgeni package.

**Harp/Jade site infrastructure:**
- [x] New `public/docs/_layout-dart-api.jade`.
- [x] Updates to other Harp/Jade template files.

**API List directive:**
- [ ] Create new or update existing `<api-list>` directive:
  - [x] Display only filter categories that are relevant to Dart.
  - [ ] Make library header a link to the library "index" page.

**Site data and Jade file generation:**
- [x] Main API page
  - [x] Generate `api-list.json`.
  - [x] Add warning that API is preliminary and that code might still appear as TS.
- [x] Generate data (`_data.json`) and Jade files for
  - [x] For each library (i.e., top-level folder).
    - [x] `_data.json`
    - [x] Library "index" Jade page.
    - [x] Each member Jade page.
  - [x] For library subfolders (e.g., containing class member details).
    - [x] `_data.json`
    - [x] Folder "index" page (it actually has the same name as the container -- e.g., class name).
    - [x] Each member page.
- [x] API Jade file page contents:
  - [x] Attributes like name and type
  - [x] Breadcrumbs
  - [ ] Page header section tabs.
  - [x] Core page content excerpt from dartdoc-generated page
    - [x] Exclude the "left" nav menu (full ng2 package content list) from the core page content.
  - [x] `{@example ...}` tags are replaced by their designated example code fragments.
- [x] Filter generated API pages (configurable via regExp in build file): exclude entries from
  - [x] Non-public, developer/internal libraries.
  - [x] Top-level catch all `angular2` library -- otherwise every entry appears twice.
- [x] Support developer & production modes for data (JSON) file generation
  - [x] Developer mode (pretty-printed).
  - [x] [NOT implemented] ~~Production mode (compact).~~

cc: @naomiblack @kwalrath @ferhatb @keertip

chalin:
Fixes #1880, which has a detailed task breakdown.
Supersedes #1593.

Note about the API subsite design: I had to resort to using Jade extends/includes; hence disabling Harp _layout/partials.

cc: @naomiblack @kwalrath @ferhatb @keertip

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -30,6 +30,7 @@ var tslint = require('gulp-tslint');
 //  2. Think about using spawn instead of exec in case of long error messages.
 
 var TOOLS_PATH = './tools';
+var ANGULAR_IO_PROJECT_PATH = path.resolve('.');
 var ANGULAR_PROJECT_PATH = '../angular';
 var PUBLIC_PATH = './public';
 var TEMP_PATH = './_temp';
@@ -63,7 +64,16 @@ var _devguideShredJadeOptions =  {
 };
 
 var _apiShredOptions =  {
+  lang: 'ts',
   examplesDir: path.join(ANGULAR_PROJECT_PATH, 'modules/@angular/examples'),
+  fragmentsDir: path.join(DOCS_PATH, '_fragments/_api'),
+  zipDir: path.join(RESOURCES_PATH, 'zips/api'),
+  logLevel: _dgeniLogLevel
+};
+
+var _apiShredOptionsForDart =  {
+  lang: 'dart',
+  examplesDir: path.resolve(ngPathFor('dart'), 'examples'),
   fragmentsDir: path.join(DOCS_PATH, '_fragments/_api'),
   zipDir: path.join(RESOURCES_PATH, 'zips/api'),
   logLevel: _dgeniLogLevel
@@ -95,6 +105,14 @@ var _exampleProtractorBoilerplateFiles = [
 ];
 
 var _exampleConfigFilename = 'example-config.json';
+
+var lang, langs;
+function configLangs(langOption) {
+  lang = (langOption || 'all').toLowerCase();
+  if (lang === 'all') { lang = '(ts|js|dart)'; }
+  langs = lang.match(/\w+/g); // the languages in `lang` as an array
+}
+configLangs(argv.lang);
 
 function isDartPath(path) {
   // Testing via indexOf() for now. If we need to match only paths with folders
@@ -131,6 +149,7 @@ gulp.task('run-e2e-tests', runE2e);
  *     all means (ts|js|dart)
  */
 function runE2e() {
+  if (!argv.lang) configLangs('ts|js'); // Exclude dart by default
   var promise;
   if (argv.fast) {
     // fast; skip all setup
@@ -183,8 +202,6 @@ function runE2e() {
 // each app/spec collection sequentially.
 function findAndRunE2eTests(filter, outputFile) {
   // create an output file with header.
-  var lang = (argv.lang || '(ts|js)').toLowerCase();
-  if (lang === 'all') { lang = '(ts|js|dart)'; }
   var startTime = new Date().getTime();
   var header = `Doc Sample Protractor Results for ${lang} on ${new Date().toLocaleString()}\n`;
   header += argv.fast ?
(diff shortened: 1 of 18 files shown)


chalin:
@wardbell @Foxandxss : ready for review.

chalin:
I forgot to mention that to build the docs do: an install, patch terraforma, and then check-deploy:
```bash
./scripts/install.sh
./scripts/patch.sh
gulp check-deploy --dgeni-log=warn  # drop the flag if you want `info` level messages as usual
```

kwalrath:
Taking a look... First impression: nice!

The second thing I noticed is the lack of a language dropdown at /docs/dart/latest/api/ (or any page under it). Would it be possible to add it? (And add some reasonable behavior in the likely case that there's no equivalent page at the dart -> ts/js version of the URL.)

If a dropdown isn't practical, we should put "Dart" in that space, so it's easier to tell that you're in the Dart version of the API docs.

chalin:
I chose to leave out the dropdown because IMHO maintaining a complete dart <-> ts/js mapping might be tricky and not really worth the effort (we can open a separate issue as a "feature request" so that we can track discussion about it). Leaving the dropdown there with only Dart is a good idea. I'll look into it.

chalin:
@filipesilva @Foxandxss @wardbell : I just rebased. Note the following change:
- Renamed `'harp-compile'` to `'check'` (as in `check-deploy` without the `deploy` :).
- Created new `'harp-compile'` that _only_ does a harp compile; i.e., doesn't depend on `'build-docs'` (where as `'check'` does).

This makes it easier to selectively rebuild only parts of the docs and then quickly get harp to start compiling.

filipesilva:
@chalin that change lgtm

chalin:
@kwalrath : I've created #1895 for the API page language-cross-referencing issue. I'd like to suggest that nothing be done in the scope of this PR. We can explore the range of alternative solutions via #1895 and a follow up PR. Thoughts?

kwalrath:
LGTM.

kwalrath:
@filipesilva @Foxandxss @wardbell @naomiblack: Unless you object, I plan to merge this after confirming that I can, indeed, generate the docs. 2 pm Pacific time is my goal.

wardbell:
Please **no gulp task called _check_**  Check _what_? I don't approve.  

I do understand the impulse. It comes down to naming. I strongly prefer one of these alternatives (in my preferred order)

1. an option on `harp-compile` such as `harp-compile --no-build`
1. new  `build-compile` task that is `check-deploy` without the deploy.

gulpfile.js
```diff
@@ -596,10 +643,35 @@ gulp.task('git-changed-examples', ['_shred-devguide-examples'], function(){
...
+  gutil.log('Note: issuing this command directly from the command line will show harp comiple warnings.');
+  return execPromise(cmd);
+});
+
+gulp.task('serve-www', [], function() {
+  // Serve generated site.
+  return execPromise('npm run live-server ./www');
+});
+
+gulp.task('check', ['build-docs'], function() {
```
wardbell: This could be `build-compile`

wardbell:
Please also update the READM.md to explain these gulp tasks-and-options.

Otherwise, LGTM

chalin:
@wardbell : named changed to `build-compile`.

More install / travis / gulpfile changes are on their way (to support Dart API docs building). I'll update the README then.

wardbell:
LGTM. @kwalrath is free to merge :+1:

chalin:
Design note (cc @petebacondarwin): while originally I started writing a dgeni package for Dart API processing, I ran into problems -- due to my lack of mastery of the dgeni framework --- and so, in the hopes of making some progress, I switched to building a stand alone transformer from the dgeni package parts. Maybe this can be revised later, or, I might just write a revised API page generator in Dart :).

PR #1900, opened 2016-07-16
chore(install): reintroduce single install script

chalin:
Our installation process can change rapidly. I think it helps if we have a single script to run that we ensure is up-to-date for all.

This PR reintroduces a single install script, which, in particular:
- **Checks for the proper version of node** (some of us develop with node
v4 on other sites, and it is too easy to forget to switch to v5 for
angular.io).
- Applies the **patch** (that is necessary to support harp compilation of
Dart API pages).

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail
+
+cd `dirname $0`/..
+
+if [[ "$(node --version)" < "v5" ]]; then
+    echo "ERROR: bad version of node detected. If you have nvm installed, type:"
+    echo "  nvm use"
+    echo "Aborting installation."
+    exit 1;
+else
+    echo "Node version: $(node --version)"
+fi
+
+echo "Installing main packages ..."
+npm install --no-optional
+
+echo "Patching ..."
+source ./scripts/patch.sh
+
+if [ "$TRAVIS" != "true" ]; then
+    echo "Rebuilding node-sass, just in case ..."
+    npm rebuild node-sass;
+fi
+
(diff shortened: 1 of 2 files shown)


chalin:
@filipesilva @Foxandxss @wardbell @kwalrath : ready for review. Note that I have _not_ touched the travis config.

scripts/install.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail
+
+cd `dirname $0`/..
+
+if [[ "$(node --version)" < "v5" ]]; then
```
Foxandxss: Node 4 is a valid node. Well, there is a new change that made v4 unusable but we could fix that.

chalin: Well, node v4 doesn't work for @kwalrath, and more often than not, forgetting to switch to v5 for Kathy has cause problems and results in wasted time.

Are you aware of anyone who _must_ stick with node 4 and who is required to build ng.io docs? (Kathy has to do this every time she tests out a PR for merging :) If not, IMHO, we should just just drop v4.

Foxandxss: No, no one, but we were always striving to have node 4.x support, but I guess it is fine to not have it on this project.

filipesilva:
lgtm.

@chalin you mentioned something about the dart compilation, is it possible to run dart specific tasks via the CI at the moment? I tried to cover the ts/js ones I know of, but I don't know anything about the dart workflow for docs.

chalin:
> is it possible to run dart specific tasks via the CI at the moment?

Not at the moment. That is part of what I will be working on next.

PR #1902, opened 2016-07-16
docs(cheatsheet/dart): temporarily use TS cheatsheet data

chalin:
Also: deleted `copyApiDocsToJsFolder()` because it is not used.


wardbell:
Ok if you say so. Merging.

PR #1903, opened 2016-07-16
docs(toh-6): use consistent header syntax

thelgevold:
@wardbell @chalin 
Per https://github.com/angular/angular.io/issues/1800 I have updated the header in the delete call to be consistent with `post` and `put`


chalin:
LGTM

PR #1912, opened 2016-07-18
docs(cheatsheet/dart): generate from dart sources

### LINKED ISSUE #1906: [Dart] Cheatsheet: generate from Dart sources

chalin:
Fixes #1906.

### DIFF
--- a/tools/dart-api-builder/index.js
+++ b/tools/dart-api-builder/index.js
@@ -37,8 +37,7 @@ module.exports = new Package('dart-api-and-cheatsheet-builder', [basePackage, ta
       parseTagsProcessor.tagDefinitions.concat(getInjectables(tagDefs));
   })
 
-  .config(function (readTypeScriptModules, writeFilesProcessor, readFilesProcessor) {
-
+  .config(function (readFilesProcessor) {
     // confirm that the angular repo is actually there.
     if (!fs.existsSync(ANGULAR_REPO_PATH)) {
       throw new Error('dart-api-and-cheatsheet-builder task requires the angular2 repo to be at ' + ANGULAR_REPO_PATH);
--- a/tools/dart-api-builder/index.js
+++ b/tools/dart-api-builder/index.js
@@ -1,0 +1,93 @@
+var fs = require('fs');
+var path = require('canonical-path');
+var Package = require('dgeni').Package;
+var basePackage = require('../api-builder/docs-package');
+var targetPackage = require('../api-builder/target-package');
+var cheatsheetPackage = require('../api-builder/cheatsheet-package');
+
+var PROJECT_PATH = path.resolve(__dirname, "../..");
+var PUBLIC_PATH = path.resolve(PROJECT_PATH, 'public');
+var DOCS_PATH = path.resolve(PUBLIC_PATH, 'docs');
+var ANGULAR_REPO_PATH = path.resolve(__dirname, '../../../angular-dart');
+var ANGULAR2_DOCS_PATH = path.resolve(ANGULAR_REPO_PATH, 'docs');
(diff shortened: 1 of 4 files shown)


chalin:
@kwalrath : ready for review. I'm unsure who else.
@petebacondarwin : could you take a look at this new dgeni package? It is largely inspired by the TS cheatsheet package.
cc @wardbell

chalin:
This PR depends on #1910.

tldr; Travis doc build is failing because I seem to have not fully turned off the Dart API and cheatsheet generation on Travis; but the cheatsheet _is_ the subject of this PR. So, I think it will be simpler to just say that this PR depends on #1910. The build works locally, but once #1910 is merged, it should build successfully on Travis too.

kwalrath:
LGTM

tools/dart-api-builder/index.js
```diff
@@ -0,0 +1,93 @@
...
+    };
+  })
+
+  .config(function (parseTagsProcessor, getInjectables) {
+    const tagDefs = requireNgIoPkg('./tag-defs');
+    parseTagsProcessor.tagDefinitions =
+      parseTagsProcessor.tagDefinitions.concat(getInjectables(tagDefs));
+  })
+
+  .config(function (readTypeScriptModules, writeFilesProcessor, readFilesProcessor) {
```
petebacondarwin: `readTypeScriptModules` not required?

tools/dart-api-builder/services/packageInfo.js
```diff
@@ -0,0 +1,38 @@
...
+
+var fs = require('fs');
+var path = require('canonical-path');
+
+/**
+ * Load information about this project from the pubspec.yaml
+ * @return {Object} The package information
+ */
+module.exports = function packageInfo() {
+  const ngPath = '../angular-dart';
```
petebacondarwin: This is hard coded here but is a constant in the index.js file. Perhaps it should be passed as some kind of property or dependency?

petebacondarwin:
My feeling with this PR is that it is duplicating a lot of what is done to generate the cheatsheets for TS and JS. Perhaps there is some way of making that processing more generic to accommodate Dart as well?

tools/dart-api-builder/index.js
chalin: No, we only need to read read `'cheatsheet/*.md'` files as input; see line 49 below.

petebacondarwin: Right, so you can remove this parameter

tools/dart-api-builder/services/packageInfo.js
chalin: I agree, that bothered me too, but it's hard coded in the same way on the TS side :)
See `tools/api-builder/angular.io-package/services/packageInfo.js` and that's good enough for me (for now).

chalin:
> My feeling with this PR is that it is duplicating a lot of what is done to generate the cheatsheets for TS and JS.

Actually, I felt that this was a clean design that reused only the necessary pieces of the TS/JS doc processing chain.

 > Perhaps there is some way of making that processing more generic to accommodate Dart as well?

I thought of that too, but some of our other Dart work has been intrusive enough already. Have a separate package that reused TS-side components, seemed like the most TS friendly way to go.  One of the main problems is that cheatsheet generation on the TS side is interwoven with the API generation, when it doesn't need to be. If someone were to do a clean refactoring on the TS side, then indeed we could reuse that for Dart. Any volunteers :)

Aside from that, is this good to go?

tools/dart-api-builder/index.js
chalin: Oh, ... right (and `writeFilesProcessor` can go too). Done.

petebacondarwin:
I think if it works then let's good with it

kwalrath:
I merged because I heard (elsewhere) that Ward's good with this.

PR #1915, opened 2016-07-18
chore(gulp): except for check-deploy, gulp targets are for (ts|js) only

chalin:
As before, by default, the various gulp tasks will be for all languages. To skip Dart API doc generation and running E2E tests use the `--no-dart` flag.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -106,11 +106,28 @@ var _exampleProtractorBoilerplateFiles = [
 
 var _exampleConfigFilename = 'example-config.json';
 
-var lang, langs;
+// Gulp flags:
+//
+//   --lang=[all | ts | js | dart | (ts|js) | (ts|js|dart) | ...]
+//
+//    This affects which language API docs and E2E tests are run. Can be 'all',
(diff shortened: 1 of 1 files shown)


chalin:
@Foxandxss @wardbell : ready for review and merge.
@almerosteyn: this should fix it.

cc @kwalrath

Foxandxss:
I like but I don't feel confortable having to put the  `--no-dart` in all my commands. Shouldn't it be otherwise? `--dart` or something like that. There is only 2 people in here that does dart and we have a l10n project to localize this project and all of them needs to run this command too and none of them really care about dart.

chalin:
I agree and debated that myself. I'm just worried that it is much easier to forget to put `--dart` when generating docs (and get no warning) than to forget `--no-dart` in other case (which _will_ result in a friendly reminder being generated).

@naomiblack, @kwalrath : any opinion about this?

wardbell:
I thought the default was no-dart except for check-deploy

chalin:
As per @wardbell's suggestion (made elsewhere) I'll make `all` the default for full-site build tasks (like  `check-deploy`) and `(ts|js)` all other gulp targets.

PR #1920, opened 2016-07-18
Added capital lettings to titles

sgmills365:
Hello, I am from ENGL 210J in the Google projects


kwalrath:
Actually, our new style is going to have _Title Case_ only for page names, and _Sentence case_ for section names.

I'm going to close this PR, but I look forward to seeing more from you!

PR #1929, opened 2016-07-19
docs(toh-6/dart): quickfix to eliminate "BAD FILENAME" errors

chalin:
Exclude the new Observables section entirely for now.


chalin:
@kwalrath ready for review. If [viewed by ignoring changes in whitespace](https://github.com/angular/angular.io/pull/1929/files?w=1), you'll see that only two lines were changed in the TS .jade file.

@wardbell : FYI ... I think it is safe for @kwalrath to merge given that there is no visible change on the TS side and it eliminates BAD FILENAME on the Dart side.

kwalrath:
LGTM. I plan to push the site after checking my build.

PR #1934, opened 2016-07-20
tests(toh-1/e2e): cleanup and updates incl. use of async/await

chalin:
Supersedes #1785.


chalin:
@wardbell @Foxandxss : ready for review. Jesus confirms that the toh-6 failure is a flake.

PR #1945, opened 2016-07-21
docs(api/dart): avoid duplicate entries and fix breadcrumbs

chalin:
Currently two entries are generated for top-level library members like classes, and the main API page for these members has no-so-useful breadcrumbs. E.g., the main API page for `PipeMetadata` has breadcrumb: `angular2 > angular2 > PipeMetadata`, vs. the proper breadcrumb `angular2 > angular2.core > PipeMetadata` that appears on the second copy of the API page entry.

TLDR; within the `angular2` package there is a top-level catch-all library of the same name the exports most of the contained libraries. This results in duplicate API page entries (one for `angular2.foo` and one for `angular2.some_sub_lib_id.foo`). Furthermore, all internal page links to `foo` go to `angualr2.foo` so we loose breadcrumb information. This commit fixes the issue by hiding the top-level `angular2` library definition only for the purpose of running dartdoc.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -108,20 +108,21 @@ var _exampleConfigFilename = 'example-config.json';
 
 // Gulp flags:
 //
-//   --lang=[all | ts | js | dart | (ts|js) | (ts|js|dart) | ...]
+//   --lang=[all | ts | js | dart | 'ts|js' | 'ts|js|dart' | ...]
 //
 //    This affects which language API docs and E2E tests are run. Can be 'all',
 //    or a regex pattern to match any one of 'ts', 'js', or 'dart'.
-//    Default: '(ts|js)' except for check-deploy for which it is 'all'.
+//    Default: 'ts|js' except for the "full site build" tasks (see below),
+//    for which it is 'all'.
 //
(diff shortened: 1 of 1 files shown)


chalin:
@kwalrath @wardbell @Foxandxss : ready for review.

I also did a minor cleanup, removing superfluous parentheses in the language regex.

chalin:
Travis doc build is green. The failure is due to the e2e toh-6 flake.

kwalrath:
The doc builds, but I don't see a difference in the output, compared to what's on angular.io. What should I be look at/for?

chalin:
Visit, say, [Pipes](https://angular.io/docs/dart/latest/api/angular2.core/Pipe-class.html), and then follow the `PipeMetadata` link in the Inheritance section.

kwalrath:
Ah, OK. So those files are still being generated, but they aren't being linked to. Goodl

chalin:
Yes.

Anything else to do before this is merged?

kwalrath:
@wardbell or @Foxandxss does the code look good to you?

PR #1947, opened 2016-07-21
docs(Npm Packages): grammar and capitalization consistency

kylerob:


public/docs/ts/latest/guide/npm-packages.jade
```diff
@@ -150,16 +150,16 @@ a(id="dev-dependencies")
   They do not have to be deployed with the production application although there is rarely harm in doing so.
 
   ***[concurrently](https://www.npmjs.com/package/concurrently)*** - 
-  a utility to run multiple *npm* commands concurrently on OS/X, Windows, and Linux operating systems.
+  A utility to run multiple *npm* commands concurrently on OS/X, Windows, and Linux operating systems.
```
Foxandxss: An

kylerob: Thanks for taking a look at the PR so fast. I think "A" is correct in this case because of the phonetics involved. Other similar examples would be "A university", "A UI", "A user guide", etc. It looks like Oxford uses "A" as well: http://www.oxforddictionaries.com/us/definition/american_english/utility

Foxandxss: I have been told that if the first letter of the next word sounds like a vowel, it is `an`. But I am not english at the end. So let's roll with what the oxford says.

kylerob: Sounds good, thanks again for taking a look.

wardbell: Interesting. The weight of the writer world is in _your favor_.
I read several posts like [this one](https://owl.english.purdue.edu/owl/resource/591/01/) which say

>The choice of article is actually based upon the phonetic (sound) quality of the first letter in a word, not on the orthographic (written) representation of the letter. If the first letter makes a vowel-type sound, you use "an"; if the first letter would make a consonant-type sound, you use "a." 

News to me and English is my first language. 

Thanks.

PR #1905, opened 2016-07-17
docs(router): Added content updates to developer guide

brandonroberts:
Added section for RouterLinkActive
Added wildcard route to example configuration
Updated static links to use one-time binding
Removed query params from Heroes list
Updated section on route query params
Added section for global query params and fragments
Added section for RouterState
Updated code samples
Renamed .guard files to .service
Renamed interfaces.ts to can-deactivate-guard.service.ts
Removed unused files

Closes #1882

PR #1952, opened 2016-07-22
docs(router): Minor content update for Heroes tutorial

massimocode:
Delete routerLinkActive from one of the earlier code examples as it is not relevant to what the user is trying to achieve at that point in the tutorial

When I was running through the tutorial, it said "Finally, add a dashboard navigation link to the template, just above the Heroes link.", but the code sample showed the routerLinkActive attribute added as well to the code. I was wondering "what's that for? why are both links being set as active?". I only realised what it was for later on in the tutorial in the section that describes the directive. It's not relevant to adding a new navigation link and can detract from the flow of the tutorial so it's better it only gets added in the bit that actually discusses it.

### DIFF
--- a/public/docs/_examples/toh-5/ts/app/app.component.3.ts
+++ b/public/docs/_examples/toh-5/ts/app/app.component.3.ts
@@ -11,10 +11,8 @@ import { HeroService } from './hero.service';
   template: `
     <h1>{{title}}</h1>
     <nav>
-      // #docregion router-link-active
-      <a [routerLink]="['/dashboard']" routerLinkActive="active">Dashboard</a>
-      <a [routerLink]="['/heroes']" routerLinkActive="active">Heroes</a>
-      // #enddocregion router-link-active
+      <a [routerLink]="['/dashboard']">Dashboard</a>
+      <a [routerLink]="['/heroes']">Heroes</a>
     </nav>
     <router-outlet></router-outlet>
   `,
--- a/public/docs/_examples/toh-5/ts/app/app.component.ts
+++ b/public/docs/_examples/toh-5/ts/app/app.component.ts
@@ -11,8 +11,10 @@ import { HeroService } from './hero.service';
   template: `
     <h1>{{title}}</h1>
     <nav>
+      // #docregion router-link-active
       <a [routerLink]="['/dashboard']" routerLinkActive="active">Dashboard</a>
       <a [routerLink]="['/heroes']" routerLinkActive="active">Heroes</a>
+      // #enddocregion router-link-active
     </nav>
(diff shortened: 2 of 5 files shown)


Foxandxss:
Not sure if this is valid right now. We changed the guide a bit and I think that this PR is not valid anymore.

/cc @brandonroberts

massimocode:
I have seen the recent PR (https://github.com/angular/angular.io/pull/1905) merged by @brandonroberts and feel that routerLinkActive was prematurely added to this first example in that PR. This PR is simply removing it from the first example, but leaving it intact on the later examples where it is more relevant.

Unless there has been a major refactoring which has not yet been committed to master, I would say this PR is relevant.

Foxandxss:
Before that snippet there is an explanation of what `routerLinkActive` is.

massimocode:
@Foxandxss You're quite right, I had updated the wrong ts file!

I still haven't managed to get the docs to run up locally due to npm install issues (node-sass on windows!! still trying to solve...), however I've updated the PR to apply the change to the correct ts file.

Foxandxss:
I see. You are right on this one, but the solution is not the appropriate one. If you modify that file, the `routerLinkActive` will disappear altogether.

massimocode:
I can't run it up locally yet but I just want to say there is a separate example file showing them added immediately after routerLinkActive is discussed. The path to the example file is `toh-5/ts/app/app.component.3.ts`. If you diff the 2 example files, the only difference is adding the routerLinkActive directive.

Foxandxss:
When we create a guide, we have two things:

* The complete example, that can be run from the terminal or via plunker.
* Extra files to show previous or alternative steps  (those *.1.*, *.2.*, etc)

That means, that the `app.component.ts` is the one that ends in the final example, so we can't use it to show a previous step.

massimocode:
Ok that makes sense. Thanks for explaining. I think I know what I need to do now. I will do my best to update the PR and run it all locally to ensure it works. Will let you know when it's done.

brandonroberts:
@massimocode I think its fine to be removed from there, but as @Foxandxss stated there needs to be a separate file provided for that example and the `app.component.ts` remain unchanged.

massimocode:
@brandonroberts @Foxandxss This has been updated now as per the feedback. Let me know if there's anything else to consider. I have managed to get everything running locally and it all looks good! Also fixed 2 x small spelling mistake in the logging of one of the build tools.

Foxandxss:
I am sorry sir, we changed that chapter quite a bit and apart from the typos, I am not sure if the other part is relevant. Could you please rebase this PR or open a new one? Thank you sir.

PR #1960, opened 2016-07-24
docs(dependency-injection): Clarify how TypeScript generates decorator metadata

Igonato:
I think it's important to mention the compiler option. After reading the note I wanted to see how this magic works and what it looks like in a compiled javascript and tried to make a simple test in TypeScript playground. I got nothing and it took some time to figure out that a compiler option is involved in this.


Igonato:
I signed it!

Foxandxss:
I like it.

wardbell:
Thank you. I agree. I took the liberty of further clarifying this point.

PR #1961, opened 2016-07-25
Double slash for bundles path

ziglee:
This slash before the bundles path is generating a double slash for the scripts path, e.g. http://localhost:3000/node_modules/@angular/common//bundles/common.umd.js. Some server are not happy with this double slash and redirect to 404 pages like Firebase Hosting.

### DIFF
--- a/public/docs/_examples/quickstart/ts/systemjs.config.1.js
+++ b/public/docs/_examples/quickstart/ts/systemjs.config.1.js
@@ -41,7 +41,7 @@
 
   // Bundled (~40 requests):
   function packUmd(pkgName) {
-    packages['@angular/'+pkgName] = { main: '/bundles/' + pkgName + '.umd.js', defaultExtension: 'js' };
+    packages['@angular/'+pkgName] = { main: 'bundles/' + pkgName + '.umd.js', defaultExtension: 'js' };
   }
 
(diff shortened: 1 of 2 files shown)


ziglee:
I signed it!

Foxandxss:
Weird, I think I tried that way in the past and it didn't work. But it certainly does.

Can you change the `systemjs.config.plunker.js` as well?

ziglee:
I suppose it can be changed too. I'm having some errors trying to run "npm start" inside public/docs/_examples folder to test this.

Foxandxss:
You can't do that there.

Just run `gulp build-plunkers` and then go to any example, inside their ts folder, there is a `plunker.no-link.html`. Open that file in your browser and you will go to plunker.

ziglee:
I had a problem with that too. 
`gulpfile.js:515
      spawnInfo = spawnExt('npm', ['install', ...sources], {cwd: EXAMPLES_PATH});
                                              ^^^

SyntaxError: Unexpected token ...`

Foxandxss:
You need node 5.

ziglee:
Plunkr works according, regardless of how the bundles path is setted at `systemjs.config.plunker.js`

Foxandxss:
That is weird. I will check.

wardbell:
Sounds good. Merging

PR #1967, opened 2016-07-25
Uwaterloo 210J T3ai edits

t3ai:
This is edit from Tian Yi Ai working in Ashley Kelly class.

### DIFF
--- a/public/docs/ts/latest/guide/template-syntax.jade
+++ b/public/docs/ts/latest/guide/template-syntax.jade
@@ -1364,12 +1364,12 @@ figure.image-display
 
 h3#aliasing-io Aliasing input/output properties
 :marked
-  Sometimes we want the public name of an input/output property to be different from the internal name.
+  Sometimes the user may require the public name of an input/output property to be different from the internal name.
 
   This is frequently the case with [attribute directives](attribute-directives.html).
-  Directive consumers expect to bind to the name of the directive.
+  Directive consumers expects to be binded to the name of the directive.
   For example, when we apply a directive with a `myClick` selector to a `<div>` tag,
-  we expect to bind to an event property that is also called `myClick`.
+  we expects to be binded to an event property that is also called `myClick`.
 +makeExample('template-syntax/ts/app/app.component.html', 'my-click')(format=".")
 :marked
   However, the directive name is often a poor choice for the name of a property within the directive class.
--- a/public/docs/ts/latest/guide/template-syntax.jade
+++ b/public/docs/ts/latest/guide/template-syntax.jade
@@ -459,17 +459,18 @@ table
   We write a template **property binding** when we want to set a property of a view element to the value of
   a [template expression](#template-expressions).
 
-  The most common property binding sets an element property to a component property value. An example is
-  binding the `src` property of an image element to a component’s `heroImageUrl` property:
+  The most common property binding sets an element property to a component property value. Please see the
+  following examples:
+  Example 1. Binding the `src` property of an image element to a component’s `heroImageUrl` property:
 +makeExample('template-syntax/ts/app/app.component.html', 'property-binding-1')(format=".")
 :marked
-  Another example is disabling a button when the component says that it `isUnchanged`:
+  Example 2. Disabling a button when the component says that it `isUnchanged`:
 +makeExample('template-syntax/ts/app/app.component.html', 'property-binding-2')(format=".")
 :marked
-  Another is setting a property of a directive:
+  Example 3. Setting a property of a directive:
 +makeExample('template-syntax/ts/app/app.component.html', 'property-binding-3')(format=".")
 :marked
-  Yet another is setting the model property of a custom component (a great way
+  Example 4. Setting the model property of a custom component (a great way
   for parent and child components to communicate):
 +makeExample('template-syntax/ts/app/app.component.html', 'property-binding-4')(format=".")
 :marked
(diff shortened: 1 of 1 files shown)


public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -496,14 +497,14 @@ table
 
 :marked
   ### Binding target
-  An element property between enclosing square brackets identifies the target property. The target property in the following code is the image element’s `src` property.
+  An element's target property is placed between enclosing square brackets. The target property in the following code is the image element’s `src` property.
 
 +makeExample('template-syntax/ts/app/app.component.html', 'property-binding-1')(format=".")
 :marked
-  Some people prefer the `bind-` prefix alternative, known as the *canonical form*:
+  Note, some people prefers the `bind-` prefix alternative, known as the *canonical form*:
```
bampakoa: typo: prefers -> prefer

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -496,14 +497,14 @@ table
 
 :marked
   ### Binding target
-  An element property between enclosing square brackets identifies the target property. The target property in the following code is the image element’s `src` property.
+  An element's target property is placed between enclosing square brackets. The target property in the following code is the image element’s `src` property.
```
kwalrath: I'm not sure whether this change is correct. (The author should comment and, if possible, provide clearer wording.) Also, we avoid passive voice such as "is placed", because it hides the doer.

public/docs/ts/latest/guide/template-syntax.jade
kwalrath: We usually avoid "note" unless we really want to make it a note (which we'd usually denote with a callout or alert).

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -496,14 +497,14 @@ table
...
 
 +makeExample('template-syntax/ts/app/app.component.html', 'property-binding-1')(format=".")
 :marked
-  Some people prefer the `bind-` prefix alternative, known as the *canonical form*:
+  Note, some people prefers the `bind-` prefix alternative, known as the *canonical form*:
 +makeExample('template-syntax/ts/app/app.component.html', 'property-binding-5')(format=".")
 :marked
-  The target name is always the name of a property, even when it appears to be the name of something else. We see `src` and may think it’s the name of an attribute. No. It’s the name of an image element property.
+  Keep in mind, the target name is always the name of a property, even when it appears to be the name of something else. Some may see `src` as the name of an attribute. But it's not. It’s the name of an image element property.
```
kwalrath: I like these changes. You got rid of "We", and you clarified the "No".

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -524,10 +525,10 @@ table
   Of course, our expression might invoke a property or method that has side effects. Angular has no way of knowing that or stopping us.
 
   The expression could call something like `getFoo()`. Only we know what `getFoo()` does.
-  If `getFoo()` changes something and we happen to be binding to that something, we risk an unpleasant experience. Angular may or may not display the changed value. Angular may detect the change and throw a warning error. Our general advice: stick to data properties and to methods that return values and do no more.
+  If `getFoo()` changes something and we happen to be binding to that something, we risk an unpleasant experience. Angular may or may not display the changed value. Angular may detect the change and throw a warning error. Our general advice is to stick to data properties, as well as methods that return values and do no more.
```
kwalrath: This change isn't bad, but I'm not sure it's any clearer than the original.

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -524,10 +525,10 @@ table
...
 
   ### Return the proper type
-  The template expression should evaluate to the type of value expected by the target property.
+  The template expression should evaluate the type of value expected by the target property.
```
kwalrath: The "to" is important. Maybe this instead:

The template expression should evaluate to a value of the type expected by the target property.

(You could probably come up with better words than mine.)

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -553,21 +554,21 @@ block dart-type-exception-example
 a(id="one-time-initialization")
 :marked
   ### One-time string initialization
-  We *should* omit the brackets when all of the following are true:
+  The brackets *should* be omited when the following conditions are true:
```
kwalrath: Avoid passive voice! (I think it'd be "omitted" if we let it in, but we won't. :) Also, it's important to say *all*, for clarity. (Otherwise, someone might think it's *any*.) Instead, maybe:

Omit the brackets when all of the following conditions are true:

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -553,21 +554,21 @@ block dart-type-exception-example
...
   * The target property accepts a string value.
   * The string is a fixed value that we can bake into the template.
   * This initial value never changes.
 
-  We routinely initialize attributes this way in standard HTML, and it works
+  This initialization is routinely used in standard HTML, and it works
```
kwalrath: Again, avoid passive voice. Maybe:

Initializing attributes this way is common in standard HTML, and...

public/docs/ts/latest/guide/template-syntax.jade
```diff
@@ -553,21 +554,21 @@ block dart-type-exception-example
...
   The following example initializes the `prefix` property of the `HeroDetailComponent` to a fixed string,
-  not a template expression. Angular sets it and forgets about it.
+  not a template expression. In this case, Angular sets it and forgets about it.
 +makeExample('template-syntax/ts/app/app.component.html', 'property-binding-7')(format=".")
 :marked
   The `[hero]` binding, on the other hand, remains a live binding to the component's `currentHero` property.
 
   ### Property binding or interpolation?
-  We often have a choice between interpolation and property binding. 
+  There often exists a choice between interpolation and property binding.
```
kwalrath: You're right to avoid "we", but I also usually avoid "there" in technical writing... it's a content-free word that's a clue that the writing could be clearer.

Once we rewrite this whole doc to use "you" instead of "we", rewriting this sentence will be easier...

Our task here is more difficult because the text isn't clear. I don't think the point is that a choice exists, but that either choice or both might be valid. In this case, I'd follow up with the author to make sure we're communicating the right thing. Maybe this text and example aren't even needed.

kwalrath:
I'm going to close this PR since we have no CLA.

PR #1972, opened 2016-07-26
docs(architecture): rename missing library

### LINKED ISSUE #1970: Outdated Architecture section
[Architecture](https://angular.io/docs/ts/latest/guide/architecture.htmll)
In the **libraries** section it says

> There are other important Angular libraries too, such as @angular/common, @angular/router, and @angular/animate. We import what we need from an Angular library module.

But apparently, _@angular/animate_ doesn't exist (as far as I can see). Am I wrong?

Foxandxss:
@angular/animate doesn't exist, so let's use http so we don't change the sentence.

Fixes #1970

### DIFF
--- a/public/docs/ts/latest/guide/architecture.jade
+++ b/public/docs/ts/latest/guide/architecture.jade
@@ -120,7 +120,7 @@ block angular-library-modules
   `!{_at_angular}/core` is the primary Angular library from which we get most of what we need.
   <br clear="all">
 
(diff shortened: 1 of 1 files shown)


chalin:
Hi, my bad for having chosen `animate`. When I made that edit a little while ago I was looking for three libraries that existed in _both_ TS and Dart. Unfortunately, `http` doesn't exist on the Dart side (it _does_ exists as a stand alone library, not a submodule of ng2).

How about using the following three: **`core`, `common`** and **`router`**, in that order?

cc @kwalrath

wardbell:
Closed after merging

PR #1982, opened 2016-07-27
chore: bump node requirements in the README

Foxandxss:
Node v.4 won't work anymore so we need to tell people.


wardbell:
Closed after merging

Issue #1981, opened 2016-07-27
API References is empty on Safari 9


DaSchTour:
The API Reference page has not content on Safari 9. There seams to be some problem in parsing the content. I get this error in the console `[Error] SyntaxError: Unexpected token '>'
	(anonyme Funktion) (api-list.js:42)`

ericjim:
I believe Safari doesn't like arrow functions yet, https://github.com/angular/angular.io/blob/master/public/resources/js/directives/api-list.js#L42

PR #1985, opened 2016-07-27
change arrow style function to regular function

ericjim:
#### Changes:
* Change arrow function to regular function. Safari does not support arrow functions yet, and as a result, it was giving a syntax error on the API reference page.

#### Closes:
* https://github.com/angular/angular.io/issues/1981

@naomiblack who uses Safari? 😄

### DIFF
--- a/public/resources/js/directives/api-list.js
+++ b/public/resources/js/directives/api-list.js
@@ -39,8 +39,9 @@ angularIO.directive('apiList', function () {
         { cssClass: 'const', title: 'Const', matches: ['var', 'let', 'const'] }
       ];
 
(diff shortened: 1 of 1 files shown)


wardbell:
LGTM. To be clear, we're talking about use of fat arrow in JS that appears in _reader's browsers_.

We will continue to use ES6 construct such as fat arrow within documentation development JavaScript tooling such as `gulpfile.js`

chalin:
Oups, my bad (I wasn't aware of that limitation in Safari). Thanks for the fix @ericjim.

ericjim:
@chalin the two people that use Safari are probably also not aware of this limitation 😄

PR #1990, opened 2016-07-28
output an audit file to keep track of documentation progress on API

ericjim:
#### Changes
* The command `gulp build-api-doc` will output a file called `ts/.../api-list-audit.json` and `js/.../api-list-audit.json`. Why? @naomiblack wants the API reference outputted in the format of `api-list-audit.template.html` so that she can keep track of progress documenting the Angular 2 APIs.
* Creating the file produces no visible side-effects on angular.io, besides the new files created after the build is made.

### DIFF
--- a/tools/api-builder/angular.io-package/templates/api-list-audit.template.html
+++ b/tools/api-builder/angular.io-package/templates/api-list-audit.template.html
@@ -1,0 +1,15 @@
+[{% for module, items in doc.data %}
+  {% for item in items %}
+    {
+      "title": "{$ item.title $}",
+      "path": "{$ item.exportDoc.path $}",
+      "docType": "{$ item.docType $}",
+      "stability": "{$ item.stability $}",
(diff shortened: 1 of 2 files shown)


tools/api-builder/angular.io-package/templates/api-list-audit.template.html
```diff
@@ -0,0 +1,15 @@
+[{% for module, items in doc.data %}
+  {% for item in items %}
+    {
+      "title": "{$ item.title $}",
+      "path": "{$ item.exportDoc.path $}",
+      "docType": "{$ item.docType $}",
+      "stability": "{$ item.stability $}",
+      "secure": "{$ item.security $}",
+      "howToUse": "{$ item.howToUse $}",
```
chalin: @ericjim @naomiblack : `howToUse` is prose, right? Do you actually want to dump it all here or just have an "Exists" / "Not Done" status like for `whatItDoes`?

PR #1991, opened 2016-07-28
docs(vstudio): Update docs to note having latest VS Update

MarkPieszak:
Update docs to note that the user should have the latest VS2015 Update (3 at this time, but just a good idea if they upgrade it either way)
Also that right clicking on Dependencies let's you `restore packages` as well, if they're using .NET Core.


Foxandxss:
Ping @DeborahK

MarkPieszak:
Apologies. Added note about the Dependencies folder (if they're using .NET Core)

PR #2000, opened 2016-07-29
chore: fix linting

Foxandxss:
For some reason there is a breaking change in TsLint that broke our build. Even with the rule disable, it would still fail.

The reality is that we don't need that `For testing only` line, is outside regions and e2e pass with normality.

I will debug the tslint issue on my own and report it.


wardbell:
The tslint pragma was misspelled and it looks like the new tslint can't survive that. But i don't mind removing it altogether.

PR #2005, opened 2016-07-31
chore(examples): Updated live example links to use helper directive

brandonroberts:
Makes all existing "live example" links open in a new window

cc: @chalin


chalin:
LGTM, including the clean tweak to the directive and helper class to also support js examples!

PR #2007, opened 2016-07-31
docs(glossary): improve DI word

cironunes:


wardbell:
meh. The original English is fine. But .. what the heck. I'll take yours.

Issue #1924, opened 2016-07-19
[Dart] ToH-6 updates following addition of Observables section on TS side


chalin:
Dart follow-up to #1528.

kwalrath:
@chalin, please handle this. (I can't assign to you.)

PR #2008, opened 2016-07-31
docs(toh-6): search query fixes

chalin:
Changes:
- Drop `asObservable()` since it is deprecated in RxJS 5 (see the [migration guide](https://github.com/ReactiveX/RxJS/blob/master/MIGRATION.md#operators-renamed-or-removed)).
- Drop `+` from hero search query URL: `app/heroes/?name=${term}+`. At best it is interpreted as a regex op that serves no purpose, at worst, it gets interpreted as a space (cf. [HTML 4.01 section 17.13.4](https://www.w3.org/TR/REC-html40/interact/forms.html#h-17.13.4)).
- Replace the generic term "term" by "pattern" since the serach string is actually a search pattern (regex).
- Rename `searchSubject` to `searchPatterns` (or maybe it should be `searchPattern$`) since subjects are just observables that can have data injected into them.
- Other minor tweak to prose.

This work is in preparation for #1924.

Note: toh-6 tests pass.

### DIFF
--- a/public/docs/_examples/toh-6/ts/app/dashboard.component.html
+++ b/public/docs/_examples/toh-6/ts/app/dashboard.component.html
@@ -1,13 +1,10 @@
 <!-- #docregion -->
 <h3>Top Heroes</h3>
 <div class="grid grid-pad">
-<!-- #docregion click -->
   <div *ngFor="let hero of heroes" (click)="gotoDetail(hero)" class="col-1-4">
-<!-- #enddocregion click -->
     <div class="module hero">
       <h4>{{hero.name}}</h4>
     </div>
   </div>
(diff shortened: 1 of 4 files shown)


Foxandxss:
Honestly, I prefer the "term" term.

wardbell:
I like these changes ... except the word "pattern".  Although you are technically correct that it is a pattern, _semantically_ it's about search terms.  I can roll with this if Patrice doesn't mind sticking with that.

I'm happy to fix as i merge or he can update. I'm standing by.

chalin:
I don't mind sticking with "term". I'll update `searchSubject` to `searchTerms` as agreed offline.

PR #2010, opened 2016-08-01
fix(docs): Renamed files for JavaScript/Dart dynamic form cookbook

brandonroberts:
Fixes issue with white page being displayed when switch to JavaScript/Dart on the dynamic form cookbook page. Also fixes the menu item to match.

cc: @wardbell @naomiblack


Foxandxss:
Is not the first time that this happens and won't be the last one. Perhaps we should look for a different solution altogether.

wardbell:
I'm confused. I thought the topic was "Dynamic Forms" (plural). "Dynamic Form" (singular) sounds weird. What was the problem?

brandonroberts:
Yes, the topic is "Dynamic Forms" (plural). If you switch to viewing the JavaScript version of "Dynamic Forms" the cookbook menu item displays "Dynamic Form" (singular)

wardbell:
um ... OK.  I think I understand. Let's move on.

PR #2011, opened 2016-08-01
Change command line for node version

mexx:
`node -v` retrieves the version of node cli, not of nodejs engine.


mexx:
I signed it!

Foxandxss:
Hello!

What we need in here is what you call the `node CLI version` and not the V8 version.

Both are a valid way of measure what version you have, but the CLI version is what the people usually use for this cases.

Thank you.

mexx:
Ok, thanks.

I was confused by some major discrepancies in the versions of `node cli` and `v8 engine` on my machine, but with your explanation it now makes sense.

Sorry for disturbance and many thanks for the great project!

PR #2015, opened 2016-08-01
docs(architecture): give example of libraries in both TS and Dart

chalin:
Follow-up to https://github.com/angular/angular.io/pull/1972#issuecomment-235738082. Original issue is #1970.

cc @kwalrath @wardbell @Foxandxss

### DIFF
--- a/public/docs/ts/latest/guide/architecture.jade
+++ b/public/docs/ts/latest/guide/architecture.jade
@@ -120,7 +120,8 @@ block angular-library-modules
   `!{_at_angular}/core` is the primary Angular library from which we get most of what we need.
   <br clear="all">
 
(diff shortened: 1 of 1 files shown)


chalin:
@wardbell : oups, sorry, I forgot to push to IdeaBlade (I forgot to _not_ just hit publish in the GitHub app).

Btw, Travis is green on [my fork for the same commit](https://travis-ci.org/chalin/angular.io/builds/148949824).

Foxandxss:
Problem is that we put `core` as an example and then we put `core` again.

chalin:
Ah, right! Fixed: I've kept the 3 libraries for TS, but only include 2 for Dart (though I could have simply reduced it to two examples for all).

PR #2020, opened 2016-08-02
Formatting - Back Quote to Wrap Identifier

mattezell:
"...builds a tree of `ActivatedRoute`s,..." should be "...builds a tree of `ActivatedRoutes`,..."

### DIFF
--- a/public/docs/ts/latest/guide/router.jade
+++ b/public/docs/ts/latest/guide/router.jade
@@ -166,7 +166,7 @@ code-example(format="", language="html").
    the `RouterLink` if we so desired.
 :marked
   ### Router State
-  After the end of each successful navigation lifecycle, the router builds a tree of `ActivatedRoute`s,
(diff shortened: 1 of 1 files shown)


mattezell:
I signed it!

kwalrath:
We don't want to imply that the type name is `ActivatedRoutes`, so this change won't work, as is.

But you're right that "`ActivatedRoute`s" is not great.

How about "`ActivatedRoute` objects"?

wardbell:
@kwalrath Exactly!  I'll fix it as I merge it

mattezell:
Wow. Thanks, @wardbell...(?) Maybe you'll catch it 4 days earlier next time...

mattezell:
Ha! Thanks, @kwalrath - you're right... I completely overlooked that before hitting submit :) 'ActiveRoute' not 'ActiveRoutes'.. Rookie mistake.  Glad it got sorted.

PR #2028, opened 2016-08-04
webpack-guide: fix ordering of resolve-extensions.

HennerM:
the .ts extension has to be before the .js extension for the angular2-template-loader to work.

### DIFF
--- a/public/docs/_examples/webpack/ts/config/webpack.common.js
+++ b/public/docs/_examples/webpack/ts/config/webpack.common.js
@@ -15,7 +15,7 @@ module.exports = {
 
   // #docregion resolve
   resolve: {
-    extensions: ['', '.js', '.ts']
+    extensions: ['', '.ts', '.js']
   },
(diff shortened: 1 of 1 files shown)


HennerM:
I signed it!

Foxandxss:
Uhm? It is working for me. 

/cc @thelarkinn

HennerM:
I experienced that the angular2-template-loader only gets called on the entrypoint files, not on the ones who are imported.

Foxandxss:
I am trying right now (RC5, ngModules). I created a `foo.component.ts` that uses a `foo.component.html` and works nice.

TheLarkInn:
Yeah id need to see an example of this to try and reproduce your issue.

Lander:
This issue as far as I know could be reproduced on Windows (Win10) environment only.

Foxandxss:
Fixed on another PR.

PR #2042, opened 2016-08-06
chore: swap reflect on webpack to core-js

Foxandxss:
New reflect-metadata 0.1.4 broke webpack.

On the one hand, there is a dozen of warnings with this line:

`const nodeCrypto = isNode && function () { try { return (void 0, require)("crypto"); } catch (e) { } }();`

Also, reflect is stuck on an old Typescript and that fires hundred of errors (which broke our build).

So I am recommending that we swap `reflect-metadata` for the one inside `core-js` which works perfect.

On the other hand I am investigating the warnings to see if they can be fix on webpack or reflect-metadata.

### DIFF
--- a/public/docs/_examples/webpack/ts/src/polyfills.ts
+++ b/public/docs/_examples/webpack/ts/src/polyfills.ts
@@ -1,6 +1,6 @@
 // #docregion
 import 'core-js/es6';
-import 'reflect-metadata';
+import 'core-js/es7/reflect';
 require('zone.js/dist/zone');
 
 if (process.env.ENV === 'production') {


wardbell:
HOLY SMOKE!  What does this mean for every sample (including the QS repo) that is loading `reflect-metadata` now?

How do we tell people about it?

Foxandxss:
No. For some reason only webpack complains about it and only on build.

Perhaps if we do system.js builds on the other examples, it will complain the same, but since we don't, we are green on that.

So in short, when we build a webpack application with 0.1.4 it gives **warnings** (a reflect or webpack bug, yet to confirm) and errors, and that is because reflect uses a really old typescript to create the final build.

If we were to build our examples (we don't) we would probably get the same errors. And since webpack is the only one with the errors, is the only one complaining.

Ideally we would fix webpack if the warnings are a bug on their side and reflect to update the typescript version.

Short them solution is to move to core-js and personally I see this like a long term solution because we can get all our polyfills in one place.

The problem is that the build is broken and we cannot leave it as is.

naomiblack:
merged, live.

PR #2046, opened 2016-08-07
chore: move a2docs.css to replace styles.css

Foxandxss:
This make effective the new `a2docs.css` as a replacement of `styles.css`.


wardbell:
Moot now that we are moving to new `aio` doc viewer

PR #2047, opened 2016-08-08
docs(InMemoryDB) In-memory API should implements the InMemoryDbService interface

Gillespie59:
Add the InMemoryDbServie interface in the angular2-in-memory-web-api sample


wardbell:
And now it does (as do other samples). Thanks for bring to our attention. Closing only because we covered it elsewhere.

PR #2052, opened 2016-08-08
chore(dart-doc-sync): data files for component-styles and lifecycle-hooks

chalin:
Note that component-styles and lifecycle-hooks examples have been sync’d for a while, though they were missing `.docsync.json` files.

cc @kwalrath


kwalrath:
LGTM. Merging now...

PR #2064, opened 2016-08-09
docs(ngmodule): remove module.id

Foxandxss:
The reasoning behind this is because `module.id` is not a well known property and since the chapter is called "Module" as well, people are thinking that `NgModule` and `module.id` are related and not really seeing a connection nor explanation.


wardbell:
Merging. Thanks.

Issue #1761, opened 2016-06-27
[TS][Example] pipes e2e fails when run locally


chalin:
This is a known problem, but I haven't found an issue to track it, so here it is. When performing the tests manually, expected results are observed. So the problem might be with protractor's timing?

```
[12:33:39] AppDir: public/docs/_examples/pipes/ts
[12:33:42]   Suite: Pipes
[12:33:46]     passed - should open correctly
[12:33:46]     passed - should show 4 heroes
[12:33:46]     passed - should show a familiar hero in json
[12:33:46]     passed - should show alternate birthday formats
[12:33:46]     passed - should be able to toggle birthday formats
[12:33:46]     passed - should be able to chain and compose pipes
[12:33:46]     passed - should be able to use ExponentialStrengthPipe pipe
[12:33:47]     passed - should be able to use the exponential calculator
[12:33:48]     failed - should support flying heroes (pure) 
    Expected 2 to equal 4, 'not mutating; should see both adds'.
[12:33:48]     Failed: Index out of bound. Trying to access element at index: 2, but there are only 2 elements that match locator By(css selector, flying-heroes #flyers div)
[12:33:48]     failed - should support flying heroes (impure) 
[12:33:48]     Expected 2 to equal 3, 'new flying hero should show in mutating array'.
[12:33:48]     passed - should show an async hero message
[12:33:48]   Suite failed: Pipes
```

chalin:
@Foxandxss : FYI, still an issue as of 07/28.

chalin:
Disabling problematic tests for now via #2068.

filipesilva:
This seems to be the cause: https://github.com/angular/angular/issues/9419

PR #2068, opened 2016-08-10
chore(pipes): disable two broken e2e tests

chalin:
Let's get Travis green again. I haven't been able to get the failing pipes test to pass, so this disables the two problematic tests until we can fix them.

cc @wardbell @Foxandxss @filipesilva


filipesilva:
lgtm

chalin:
Temporary measure in addressing  #1761.

PR #2078, opened 2016-08-10
docs(quickstart): post-RC5 Dart resync

chalin:
- Resync Dart QS prose with TS, post RC5.
- Added to-be-shared Jade `var` definitions to `_util-fns.jade`.
- Delete cached QS jade file.
- Eliminate the stub file `styles.1.css` in favor of a `docregion` in the main `styles.css`.

This commit requires a `gulp add-example-boilerplate` after pulling it in.
Contributes to #2077.

Issue #2077, opened 2016-08-10
[Dart] resync Dart .jade files, post-TS-RC5


chalin:
Follow up to #2043.
- [x] guide/architecture.jade
- [ ] guide/attribute-directives.jade
- [ ] guide/component-styles.jade
- [x] guide/dependency-injection.jade
- [ ] guide/displaying-data.jade
- [x] guide/hierarchical-dependency-injection.jade - #2080
- [x] guide/lifecycle-hooks.jade - #2141
- [x] guide/pipes.jade
- [x] guide/security.jade
- [ ] guide/server-communication.jade
- [x] guide/structural-directives.jade
- [x] guide/template-syntax.jade - #2121
- [x] quickstart.jade - #2078, #2089
- [x] tutorial/toh-pt6.jade - #2095

Foxandxss:
Closing this, if that is a mistake, we reopen.

chalin:
Yes, closing it now is fine. I've migrated the issue to https://github.com/dart-lang/site-webdev/issues/368.

PR #2080, opened 2016-08-10
docs(hierarchical-di): post-RC5 Dart resync

chalin:
Contributes to #2077.
Depends on #2078.


chalin:
@kwalrath @Foxandxss : ready to be reviewed. The changes are simple enough that it might be sufficient for Kathy to review.

kwalrath:
LGTM (both TS & Dart)

chalin:
@kwalrath @Foxandxss @wardbell : I've decided to keep the `_cache` fresh rather than minimal (i.e., rather than delete cache files as they are no longer needed). This way, we can now see the diff between the cache and the current file version.

So, FYI, the update I just pushed has a refreshed cache file for this chapter (but no other changes).

PR #2083, opened 2016-08-10
chore(build): fix so --fast can be used with check-deploy etc

chalin:
Using `--fast` with most targets would just stop after doing `_copy-example-boilerplate`. This fixes the issue.

cc @kwalrath @Foxandxss

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -441,7 +441,7 @@ gulp.task('add-example-boilerplate', function(done) {
 // copies boilerplate files to locations
 // where an example app is found
 gulp.task('_copy-example-boilerplate', function (done) {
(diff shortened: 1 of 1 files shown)


kwalrath:
Works for me!

Foxandxss:
I need to double check, because that concrete task had a problem with ordering and I think I had to remove the `return` to make it work so I will check later to see that it still does the job in the right order.

chalin:
Removing the return beaks the task flow. I think that the return needs to be there in some form, we just need to find the right form (and this _might_ be it).

Foxandxss:
Doesn't seem to be affecting the ordering, so LGTM.

PR #2092, opened 2016-08-11
chore(toh-5): improve e2e test coverage and readability (use async/await)

chalin:
Improved test coverage. Also updated tests to make use of async/await (so tests are more readable).

This is part of work to get a full suite of e2e tutorial tests to run for Dart (#1619).

### DIFF
--- a/public/docs/_examples/toh-5/e2e-spec.ts
+++ b/public/docs/_examples/toh-5/e2e-spec.ts
@@ -1,113 +1,190 @@
 /// <reference path='../_protractor/e2e.d.ts' />
 'use strict';
-describe('Tutorial', function () {
 
-  beforeAll(function () {
-    browser.get('');
-  });
+const expectedH1 = 'Tour of Heroes';
+const expectedTitle = `Angular 2 ${expectedH1}`;
+const targetHero = { id: 15, name: 'Magneta' };
+const targetHeroDashboardIndex = 3;
+const nameSuffix = 'X';
+const newHeroName = targetHero.name + nameSuffix;
 
-  function getPageStruct() {
-    let hrefEles = element.all(by.css('my-app a'));
+type WPromise<T> = webdriver.promise.Promise<T>;
+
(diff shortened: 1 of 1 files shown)


chalin:
@wardbell @Foxandxss @filipesilva : ready for review.

cc @kwalrath

public/docs/_examples/toh-5/e2e-spec.ts
```diff
@@ -1,113 +1,190 @@
...
+      element(by.buttonText('View Details')).click();
+
+      let page = getPageElts();
+      expect(page.heroDetail.isPresent()).toBeTruthy('shows hero detail');
+      let hero = await Hero.fromDetail(page.heroDetail);
+      expect(hero.id).toEqual(targetHero.id);
+      expect(hero.name).toEqual(targetHero.name);
+    });
+
+    it(`updates hero name (${newHeroName}) in details view`, updateHeroNameInDetailView);
```
Foxandxss: Isn't this the same test as line 98?

Foxandxss:
Leaving the comment aside, I don't dislike it, the new Hero class makes it read better.

public/docs/_examples/toh-5/e2e-spec.ts
chalin: Yes, we test the updating of the hero name twice: once when accessed via the Dashboard, and once when accessed via the Heroes list. This is how it was in the previous e2e tests.

Foxandxss: But I can't see how they can do anything different. Tests runs from a clean state and they seem to be run from the exact same page.

Shouldn't it be dropped? Or there is something I cannot see?

chalin: The test groups for the Dashboard and the Heroes list are **independent**. I think that it is reasonable to include this single test (editing the hero name), in both test groups.

PR #2095, opened 2016-08-11
docs(toh-6): post-RC5 Dart resync and TS fixes

chalin:
Contributes to #2077.

TS-side changes include:
- Merged three versions of `app/app.module{,1,2}.ts` into a single file and used docregions instead.
- Misnamed files:
  - `rxjs-operators.ts` -> `rxjs-extensions.ts`
  - `hero-search.service.html` -> `hero-search.component.html`
- Fixed BAD FILENAME error.

Lint reports no errors and toh-6 e2e tests pass.


chalin:
@wardbell @Foxandxss: I'm not to sure who should review this, but given that @brandonroberts was the last to edit it, I'm including his name.

@kwalrath : to see the diff between the old cached copy and the latest, look at the second commit. Note that in the final commit the cached and latest copies are the same (as they should be).

All: as usual, it is best to [view diffs by ignoring whitespace](https://github.com/angular/angular.io/pull/2095?w=1).

kwalrath:
LGTM. I compared the old & new Dart _and_ versions side-by-side, and the few diffs I noticed were improvements.

brandonroberts:
LGTM too

Issue #2035, opened 2016-08-05
[Dart] add router guide


chalin:
Follow-up work to #2026.
- [ ] Create chapter JADE.
- [ ] Create chapter example source.
- [ ] setup example under `angular-examples`.

cc @kwalrath

chalin:
Migrated to https://github.com/dart-lang/site-webdev/issues/198

PR #2101, opened 2016-08-12
docs(router): fix markdown, and example mixin cleanup

chalin:
- Fixed two instances of markdown text not under `:marked` region (and so that text was not showing up in the generated html including the heading "Router Lifecycle Hooks").
- Fixed: code excerpt title `constructor` -> `isSelected`.
- Cleanup of all makeExample mixin uses.

Preliminary step to #2035: this cleanup will help make it easier to record differences with the deprecated router chapter (which must have app-relative makeExample paths).


chalin:
@brandonroberts : ready for review.

cc @wardbell @Foxandxss @kwalrath

brandonroberts:
The first code example under `Query Params and Fragments` is missing a header. Everything else LGTM

chalin:
Good catch. Fixed.

PR #2106, opened 2016-08-13
docs(ngmodule): Fix typo of italic in backquote

trotyl:
The underscore would turn to real underscore character within backquote and resulting in `_entryComponents_`.


wardbell:
Thanks

PR #2110, opened 2016-08-14
docs(lifecycle-hooks): Remove statements of router-deprecated hooks

trotyl:
The lifecycle-hooks provided by router-deprecated has already been removed in the router, the router now uses guards rather than lifecycle-hooks to achieve the same work.


wardbell:
Thanks.

PR #2112, opened 2016-08-14
docs(cookbook) Custom exception handler service cookbook.

bennadel:
This cookbook demonstrates how to override the core `ExceptionHandler` class
with a custom service class that both logs to the console as well as sending
errors to a robust error-logginer service. This error-logging service can,
in turn, send errors to the server; or, to Software-as-a-Service (SaaS)
products like New Relic, TrackJS, and Raygun.


Foxandxss:
What was the problem with the e2e here?

bennadel:
@Foxandxss the fact that the UI stops updating after an error is thrown. As such, I couldn't capture the error and then update the UI and then test the UI in the e2e tests. It seems that the only way to test might be to use non-standard testing. So, for now, I'm just testing to make sure nothing blows up.

zoechi:
@bennadel from what I have seen this is because change detection doesn't run after an exception. Can't you invoke change detection manually? (I don't know NG2/TS testing well).

PR #2125, opened 2016-08-16
docs(glossary): add ECMAScript2015 section back

cironunes:
it was removed, but it's the explanation of the shorthand of ES2015

### DIFF
--- a/public/docs/ts/latest/glossary.jade
+++ b/public/docs/ts/latest/glossary.jade
@@ -358,7 +358,13 @@ a#aot
     to ES5 JavaScript.
 
     Angular 2 developers may choose to write in ES5 directly.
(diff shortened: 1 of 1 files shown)


public/docs/ts/latest/glossary.jade
```diff
@@ -358,7 +358,13 @@ a#aot
     to ES5 JavaScript.
 
     Angular 2 developers may choose to write in ES5 directly.
-
+:marked
+  ## ECMAScript 2015
+.l-sub-section
+  :marked
+    The latest released version of JavaScript,
```
wardbell: It's actually a "recent" release. The latest is ES2016/ES7.

PR #2131, opened 2016-08-17
chore: update karma for latest zone.js

Foxandxss:
Simplified the shim a bit and brought support for latest zone.js

### DIFF
--- a/public/docs/_examples/karma-test-shim.js
+++ b/public/docs/_examples/karma-test-shim.js
@@ -29,26 +29,20 @@ System.config({
 
 System.import('systemjs.config.js')
   .then(function () {
-    return Promise.all([
-      System.import('@angular/core/testing'),
-      System.import('@angular/platform-browser-dynamic/testing')
-    ])
-  })
-  .then(function (providers) {
-    var testing = providers[0];
-    var testingBrowser = providers[1];
-
-    testing.setBaseTestProviders(
-      testingBrowser.TEST_BROWSER_DYNAMIC_PLATFORM_PROVIDERS,
-      testingBrowser.TEST_BROWSER_DYNAMIC_APPLICATION_PROVIDERS);
-
-  })
-  .then(function() {
-    // Finally, load all spec files.
-    // This will run the tests directly.
-    return Promise.all(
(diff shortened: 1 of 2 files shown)


public/docs/_examples/karma-test-shim.js
```diff
@@ -29,26 +29,20 @@ System.config({
...
-  .then(function() {
-    // Finally, load all spec files.
-    // This will run the tests directly.
-    return Promise.all(
-      allSpecFiles.map(function (moduleName) {
-        return System.import(moduleName);
-      }));
-  })
-  .then(__karma__.start, __karma__.error);
+    System.import('@angular/core/testing')
```
wardbell: I don't understand what this change fixes.

Foxandxss: Sorry, this is not a fix. This PR Fixes zone.js update and updates karma-test-shim to use TestBed

wardbell:
Please note that these karma files will be deleted from `_examples`. The canonical versions of these and all unit testing support files will be in the Testing chapter, `_examples/testing/ts`.

Making this change because (a) the Testing chapter is the authority on unit testing and (b) there is no reason for other samples to carry the overhead of unit testing files when they don't have unit tests.

Keeping this issue alive until I have ported the npm updates to the Testing chapter samples.

wardbell:
Do we really need long-stack-trace? Why?

Foxandxss:
You are right, we don't need it. I reverted that.

wardbell:
But the other three ... including `jasmine-patch` ... all needed for sure?

Foxandxss:
Apparently not. I was having some weird issues but I guess I was doing something wrong. So sorry.

Removed it.

wardbell:
I'm closing after porting the addition of the two new files to the Testing sample. Keeping these (in their various incarnations):
```
  <script src="node_modules/zone.js/dist/sync-test.js"></script>
  <script src="node_modules/zone.js/dist/proxy-zone.js"></script>
```
I had already refactored `karma-test-shim` and its cousins so the other change is no longer pertinent.

Thanks though ;-)

PR #2138, opened 2016-08-18
docs(router): Added section on animating route components

brandonroberts:


brandonroberts:
@wardbell ready for review

public/docs/_examples/router/ts/app/crisis-center/crisis-detail.component.1.ts
```diff
@@ -26,10 +28,44 @@ import { Subscription }           from 'rxjs/Subscription';
...
+          opacity: 0,
+          transform: 'translateY(100%)'
+        }))
+      ])
+    ])
+  ]
 })
 // #docregion cancel-save
 export class CrisisDetailComponent implements OnInit, OnDestroy {
+  @HostBinding('@routeAnimation') get routeAnimation() {
```
johnpapa: I recommend clarifying that the @HostBinding decorator statements go inside the Component class. SImple way to do it would be to show the class in a code snippet below this.

public/docs/ts/latest/guide/router.jade
```diff
@@ -868,27 +869,72 @@ a#nav-to-list
...
+  setup. We'll use the **wildcard state** that matches any animation state our route component is in, along with
+  two *transitions*. One transition will animate our component entering the application view (`void => *`), while the other
+  controls our component leaving the application view (`* => void`).
+
+  We could add different transitions to different route components depending on our needs, but we'll just animate our
+  *Hero Detail* component for this milestone.
+
+.l-sub-section
+  :marked
+    We'll cover transitions between **route paths** in a future update.
```
johnpapa: Leaving us all hanging!

johnpapa:
@brandonroberts I like this but I think it would be great to add something to address the natural questions a reader may have about how this show an animation on a single component and is this good or bad. I see the note about more is coming ... but perhaps we could add something to say that we know it is not desirable to add this to every component, so it would be great to add them at a more generic level to animate components that route to each other. We'll get there in a future update.

something like that?

public/docs/ts/latest/guide/router.jade
brandonroberts: The API isn't ready yet! Maybe I should just take this out.

brandonroberts:
@johnpapa Yea, some clarification is needed if we're telling them to do it this way now until the other API is ready.

public/docs/ts/latest/guide/router.jade
johnpapa: We could do that. But this entire section leaves the reader with some natural questions about how to do multi route animations/transitions without copying and pasting this code to every component. And when to use this technique vs using CSS animations for this.

public/docs/_examples/router/ts/app/crisis-center/crisis-detail.component.1.ts
brandonroberts: Fixed

brandonroberts:
@johnpapa how this sound for clarification on adding animations to individual route components?

`Using route animations on each individual component is a viable option that should only be used sparingly. A more general approach to adding animations between **route paths** without adding animations to individual components will be covered in a future update to this chapter.`

brandonroberts:
Closed in favor of #2214

PR #2152, opened 2016-08-19
example(toh-4,5): getHeroesSlowly() to return getHeroes()

chalin:
Have `getHeroesSlowly()` delay and then return the value of `getHeroes()`. This makes it easier for user’s performing the tutorial to keep this slower method as they evolve toh-5 into toh-6.

@kwalrath @Foxandxss @filipesilva : note that I've partitioned the files. There is a TS commit and a Dart commit, each ready to be reviewed.

### DIFF
--- a/public/docs/_examples/toh-4/dart/lib/hero_service.dart
+++ b/public/docs/_examples/toh-4/dart/lib/hero_service.dart
@@ -17,7 +17,7 @@ class HeroService {
   // See the "Take it slow" appendix
   // #docregion get-heroes-slowly
   Future<List<Hero>> getHeroesSlowly() {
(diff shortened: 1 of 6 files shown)


Foxandxss:
I am good with it, but I am unsure if the new version is a bit harder to understand (the setTimeout changes).

chalin:
I think that it separates concerns so that the `setTimeout(...)` does just that (and nothing more); otherwise it gets a bit convoluted to integrate the call to getHeroes _inside_ the call to setTimeout (at least I couldn't find a clean way that worked).

chalin:
@Foxandxss: can you tag as LGTM so that @wardbell can have a look at this too?

PR #2161, opened 2016-08-21
docs(cookbook - aot compiler)

thelgevold:
Ready to go!


thelgevold:
Alex has fixed the issue in tsc-wrapped that caused problems with symlinked node_modules

thelgevold:
This cookbook is now ready for review.
The basic idea behind this cookbook is to show a simple application where we do AOT and rollup based tree shaking.
 
@wardbell @robwormald

thelgevold:
Run the sample using `npm run build:aot && npm run lite`

public/docs/ts/latest/cookbook/aot-compiler.jade
```diff
@@ -0,0 +1,177 @@
...
+    
+    As an example, if our application is not using any `Forms` functionality, there is no reason to make the end user download any Forms related Angular code.
+
+    Three Shaking and AOT compilation are separate steps, but Tree Shaking can only target JavaScript code. By doing AOT compilation we convert more of our application to JavaScript, which in turn makes more of our application "Tree Shakable".
+
+    ### Rollup
+
+    In our example we are using a framework called Rollup to do the Tree Shaking.
+
+    Rollup works best with `ES2015` source code, so we transpile our application to `ES2015` by setting the `target` and `module` properties in `tsconfig.json` to `es2015`. This takes care of onverting our TypeScript code to `ES2015` JavaScript.
```
bennadel: `onverting` -> `converting`

thelgevold:
Spoke with @robwormald . We should definitely incorporate the closure compiler into this instead of using babel. The potential improvements from switching could be significant.

public/docs/ts/latest/cookbook/aot-compiler.jade
```diff
@@ -0,0 +1,177 @@
...
+
+    ### Why do AOT compilation?
+
+    The performance improvement from doing AOT compilation can be significant for two reasons:
+
+    *Faster rendering*
+
+    When doing AOT compilation the browser downloads a pre-compiled version of the application. This means the browser can start rendering the application immediately, without delays from doing JIT compilation first.
+
+    The compiler will also inline external files like html templates and css files, which means we no longer need separate ajax requests to load external files at run time.
```
johnpapa: sounds like 3 reasons.  downloads a pre-compiled version, inlines html thus reducing requests for HTML, and small angular framework download

public/docs/ts/latest/cookbook/aot-compiler.jade
```diff
@@ -0,0 +1,177 @@
...
+
+    *Smaller Angular framework download size*
+
+    Compilation at build time means we no longer have to download the compiler with our application. The size of the compiler is significant, so omitting it will automatically reduce the download size of Angular.
```
johnpapa: maybe replace "no longer" with "don't"

public/docs/ts/latest/cookbook/aot-compiler.jade
```diff
@@ -0,0 +1,177 @@
...
+
+    ### Compiling the application
+
+    We trigger AOT compilation via the command line using the previously installed `ngc` compiler by executing:
+
+    `node_modules/.bin/ngc -p [path to project]`
+
+    `ngc` expects the `-p` switch to point to a tsconfig.json file or a folder containing a tsconfig.json file.
+
+    The first thing we notice after running `ngc` is a series of generated files with an ngfactory suffix.
```
johnpapa: Before we tell them to run `ngc` it may be good to have them set the gen folder so the files do not go side by side.

public/docs/ts/latest/cookbook/aot-compiler.jade
```diff
@@ -0,0 +1,177 @@
...
+
+    `NgFactory` files do not replace the original component files, but we can think of them as a JavaScript representation of the component's template and css. The original component class is still used internally by the generated `NgFactory`.
+
+    To see what we mean by a "JavaScript representation" of the orignal Angular template syntax, just open up one of the generated `NgFactories` and look for familiar symbols like `NgIf` and `NgFor`. We will no longer see these directives in markup, but instead we will see them represented as JavaScript code. 
+
+    Since `NgFactories` are generated, we should not make any edits to them since any edits will be lost the next time we AOT compile the application.
```
johnpapa: should --> must

public/docs/ts/latest/cookbook/aot-compiler.jade
```diff
@@ -0,0 +1,177 @@
...
+
+    In our example we are using a framework called Rollup to do the Tree Shaking.
+
+    Rollup works best with `ES2015` source code, so we transpile our application to `ES2015` by setting the `target` and `module` properties in `tsconfig.json` to `es2015`. This takes care of onverting our TypeScript code to `ES2015` JavaScript. 
+    
+    We also specify an output folder for `NgFactories` in `tsconfig.json` by defining `angularCompilerOptions` with a `genDir`. 
+
+    Next we have to create a configuration file to tell Rollup how to process our application. 
+    
+    In our example we have named the configuration file rollup.js.
```
johnpapa: rollup.js --> `rollup.js`

public/docs/ts/latest/cookbook/aot-compiler.jade
```diff
@@ -0,0 +1,177 @@
...
+
+    In rollup.js we tell Rollup that the entry point to our application is app/main.js by setting the `entry` attribute.
```
johnpapa: put file names and paths inside ticks

public/docs/ts/latest/cookbook/aot-compiler.jade
```diff
@@ -0,0 +1,177 @@
...
+  `rollup.js,
+   tsconfig.json`
+) 
+
+:marked
+    ### Rollup Plugins
+
+    Our application code is now generated in `ES2015` format and the Angular framework is by default deployed as `ES2015`. However, we have one slight challenge with RxJs. 
+    
+    Angular has a direct dependency on RxJs, but RxJs code bundles are not published as `ES2015` JavaScript.
```
johnpapa: This is something I know @wardbell has commented on in the past. The phrasing of the RxJS dependency may be something he wants to articulate.

johnpapa:
I think it is important to show JIT and AOT for the same sample. Start with JIT working, then show it working for AOT. `index.html`, `main.ts` and `tsconfig.json` are different for JIT vs AOT.

Also, we should show `module.id` in the source files for these samples because JIT needs it, and with AOT we can fake it.

public/docs/ts/latest/cookbook/aot-compiler.jade
```diff
@@ -0,0 +1,177 @@
...
+
+    The first thing we notice after running `ngc` is a series of generated files with an ngfactory suffix.
+
+    `NgFactory` files do not replace the original component files, but we can think of them as a JavaScript representation of the component's template and css. The original component class is still used internally by the generated `NgFactory`.
+
+    To see what we mean by a "JavaScript representation" of the orignal Angular template syntax, just open up one of the generated `NgFactories` and look for familiar symbols like `NgIf` and `NgFor`. We will no longer see these directives in markup, but instead we will see them represented as JavaScript code. 
+
+    Since `NgFactories` are generated, we should not make any edits to them since any edits will be lost the next time we AOT compile the application. 
+
+    With JIT compilation the factory files are out of sight, out of mind. The same `NgFactory` files are generated during JIT compilation though. We just never see the physical files because it happens at runtime - on the fly.
```
Foxandxss: The "out of sight, out of mind" sounds weird to me.

Foxandxss:
Here is my feedback after reading it:

* We should make this guide to work with the Quickstart. I tried and it works, it just needs some tweaks:
  * The number of dependency to install coming from a quickstart project.
  * Where are the aot stuff generated. (I get it at `/aot` and not `/app/aot`)
  * Most problematic is having to modify the index.html to use the bundle.js

The prose itself is good, but I feel like it is too rushed in some parts, like wanting to make it as small as possible but not explaining some stuff:

* It is missing all the rollup dependencies (rollup, the plugins, the babel es2015...).
* In the `Rollup` section you talk about `tsconfig` and I feel like that should be in its own section.
  * Also I would put an snippet showing the tsconfig more immediately than forcing the user to scroll down to see the final result.
* I don't like how the `Rollup` section is structured. You show the final rollup.js and then you start explaining the plugins like if it were "something extra to add" and well, it isn't.

On this latter point, I would personally explain what is rollup, why we want it, create a basic `rollup.js` (where you explain the default export) and then you explain the plugins and then you show the entire file.

One **important** step that is completely missing is: Ok, I have the `rollup.js` now what? How is that connected to my app because you mention a `dist/build.js` that I don't have. So you need to explain how to run rollup. 

I learnt a lot in here, but I feel like the different sections are loosely connected and there is some gaps in the prose. Also, te example is not really an example, is just theory.

I would say to get the quickstart, perhaps add some basic heroes' code (to show something, you know) and from there create a story, where we have our app, we explain the JiT vs AoT (that you did nicely) and from there we explain step by step (with snippets, instructions) how can we get our application from the quickstart to use AoT. Installing all the dependencies we don't have already (we assume that we will have TS 2 and @types), explain why we need them, show them how this `ngc` works, the tree shaking... Once we are there, we install all the rollup dependencies, we create the file with the proper explanations, we explain how to run it and lastly, we use that build to run our app.

On the other hand, I guess that we would need more content / cookbooks to explain AoT with lazy loading and other advanced topics.

Foxandxss:
@johnpapa 

>Before we tell them to run ngc it may be good to have them set the gen folder so the files do not go side by side.

I totally agree. We can show this after the tsconfig.json configuration.

> I think it is important to show JIT and AOT for the same sample. Start with JIT working, then show it working for AOT. index.html, main.ts and tsconfig.json are different for JIT vs AOT.

If we go the quickstart path, the user should be familiar enough with JiT to having to show it yet again.

thelgevold:
@robwormald I converted to use the Closure compiler and removed the dependency on Babel. Closure integration with Rollup works for this simple sample, but I have seen a few issues in larger projects. Basically I've seen the process either run out of memory or hang. Would appreciate any feedback on whether or not the Closure code is correct.

naomiblack:
small nitpick -- per dev guide editing recommendations, could we use "just-in-time (JiT) compiler" and "ahead-of-time (AoT) compiler" please? I'll make the change if it's a pita... let me know.

@kwalrath

thelgevold:
@naomiblack No problem. I will sweep the doc for that change

wardbell:
We have not had time to incorporate all of the excellent suggestions. We need to factor them into a revision.

More importantly, I have not been able to get the sample to work AND there are some small disconnects between the instructions (e.g., what packages to install) and the text. I'm hoping those can be ironed out in the morning.

Finally, I am concerned that the presence of `example-config.json` will cause our tooling to accidentally wipe out one or more files (e.g., `package.json`). This seems a bit fragile.

All of that said, we intend to merge this tomorrow morning even if in its present state. Hoping someone can fix a few things before we do. We will certainly improve it soon.

thelgevold:
@wardbell Where did the sample break?

I did a clean npm and it worked for me: 
After npm install I do the following:

From the `cb-aot/ts` folder I run 1) `npm run build:aot` 2) `npm run lite`

@Foxandxss Do you have time to give it a try to see if it works for you?

Foxandxss:
I made some changes of my own, so please double check it.

Important bits are:

* I removed the compiler from the dependencies to install (it exist already) and instead of doing --save-dev I do --save only.
* Changed the rollup stuff to --save only too.

Our package.json is a bit of a mess right now. In a good world, we would have all the dependencies needed to do a *build* in `dependencies` and the ones not needed for a build at `devDependencies`. Since all the rollup & AoT stuff is needed for build, it is better to save it there.

At the moment we have `typescript` in the `devDependencies` and it should be on `dependencies` too, but that is a sweep for another day.

Foxandxss:
Ah, I tested it. Works like a charm and the `es5` target is a brilliant idea @wardbell

PR #2163, opened 2016-08-22
docs(NgModule): fix typo

SamVerschueren:
Changed

> To reference the _type_ we'd have to import the module, which loads the module **loads** immediately

to

> To reference the _type_ we'd have to import the module, which loads the module immediately


wardbell:
Thanks. Incorporating in another fix-up PR.

SamVerschueren:
🍻

PR #2165, opened 2016-08-22
chore: update karma to zone 0.6.15

Foxandxss:
Yet another update. I also updated the tests in ngUpgrade because I forgot the last time.

### DIFF
--- a/public/docs/_examples/karma-test-shim.js
+++ b/public/docs/_examples/karma-test-shim.js
@@ -28,22 +28,18 @@ System.config({
 });
 
 System.import('systemjs.config.js')
-  .then(function () {
-    return Promise.all([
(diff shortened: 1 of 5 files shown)


wardbell:
I don't quite understand this one.  There are changes to `1.js` versions but not the ones without `.1`.

Also the list of zone modules is incomplete. I'm adding what I believe to be the correct set in the next commit.

Foxandxss:
This change was for .15 which only needed a file rename, but in the same day .16 came out requiring all the files, I just hadn't time to come back and change it.

I also added the generic karma-shim. I know you want to remove it, but I am not sure if that is an option anymore, it is needed for the upgrade guide.

PR #1911, opened 2016-07-18
[Backup] docs(testing): chapter and sample rewrite for RC5

wardbell:
**Do NOT merge ... superseded by #2198**
 
Backup for #2198  which replaces it. Keeping for a brief while as 2198 started as the rebase and squash of this PR and I'm not confident that everything survived.

Delete soon ... and definitely after merging #2198

PR #2168, opened 2016-08-22
docs(ngmodule): fix broken link

### LINKED ISSUE #2166: NgModule: BrowserModule Vs. CommonModule not explained
The docs state the following

> We also replaced BrowserModule by CommonModule for reasons we explain [soon](https://angular.io/docs/ts/latest/guide/ngmodule.html#root-vs-feature-module).

Does `soon` mean it will be added in the near future, or that it will be explained later on in the documentation? Because currently, the link is not pointing to anything.

After reading the entire document, I believe this is the same reason as to why you have `SharedModule.forRoot()` and `SharedModule`. Because in this case, `BrowserModule` bootstraps application-wide providers. If you would use `BrowserModule` everywhere, lazy-loaded feature modules will create a new instance of those providers in their child-injector. 

I'm not sure though, just guessing here. Would be nice if it was explained more explicitly.

Foxandxss:
The link was pointing to nowhere.

I am not completely sure if the prose mentions this reasons, but the FAQ does it really nicely.

Fixes #2166

### DIFF
--- a/public/docs/ts/latest/guide/ngmodule.jade
+++ b/public/docs/ts/latest/guide/ngmodule.jade
@@ -548,7 +548,7 @@ a#feature-modules
     The `ContactModule` must import `FormsModule` explicitly so that
     `ContactComponent` can data bind with `ngModel`.
 :marked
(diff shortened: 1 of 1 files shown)


wardbell:
Thanks. Incorporated in #1911

PR #2171, opened 2016-08-22
chore(dart e2e): add --[no-]pub flag

chalin:
This flag (on by default) determines if `pub {upgrade,build}` is run
before launching e2e tests. It is useful to be able to disable when
tweaking the e2e tests themselves.

cc @kwalrath @Foxandxss


kwalrath:
LGTM

PR #2177, opened 2016-08-22
chore: update to rc6

filipesilva:
The suites still failing on CI are due to a `Type` from `@angular/core` now being a generic (`Type<T>`), and thus triggering a compilation failure on https://github.com/angular/in-memory-web-api. Here is a PR that addresses that: https://github.com/angular/in-memory-web-api/pull/18.

This is still missing package.json RC6 versions.


wardbell:
We'll fix the `in-memory-web-api` at the same time that we process this.  Just have to figure out the timing.

filipesilva:
*Breaking changes, not previously deprecated*

* `Type` -> `Type<T>`
* `ExceptionHandler` -> `ErrorHandler`, 
* `SanitizationService` -> `Sanitizer`,
* `DomSanitizationService` -> `DomSanitizer`
* `BaseException` has been removed. Use JS `new Error(...)`

Foxandxss:
We are fine bot.

Issue #2173, opened 2016-08-22
[Dart] doc links to API pages should land on Dart, not TS pages


chalin:
In the current docs, most links to API pages have a URI that is TS specific.

Now that Dart API docs are in, we need to stop excluding links to Dart API pages in the broken link checker options. Otherwise, there will be missing many broken links (e.g., #2126).
- [x] excluding links to Dart API pages in the broken link checker options
- Fix links to API pages
  - [x] QuickStart - no broken links
  - [x] tutorial - no broken links
  - [x] guide - #2185
  - [x] API pages - #2187. This is mainly the language-switcher pulldown menu.

While removing the regex that excludes Dart API links is easy, fixing all the broken links will be a challenge because all dev guide links to API pages currently use TS API doc page URIs.

cc @kwalrath

PR #2185, opened 2016-08-24
docs(*): fix links to Dart API entries

chalin:
Contributes to #2173.

### DIFF
--- a/public/docs/dart/latest/guide/forms.jade
+++ b/public/docs/dart/latest/guide/forms.jade
@@ -518,7 +518,7 @@ figure.image-display
     Recall from the previous section that `ngControl` registered this input box with the
     `NgForm` directive as "name".
 
-    We didn't add the **[`NgForm`](../api/angular2.common/NgForm-class.html) directive*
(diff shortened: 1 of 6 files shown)


public/docs/dart/latest/guide/forms.jade
```diff
@@ -518,7 +518,7 @@ figure.image-display
     Recall from the previous section that `ngControl` registered this input box with the
     `NgForm` directive as "name".
 
-    We didn't add the **[`NgForm`](../api/common/index/NgForm-directive.html) directive*
+    We didn't add the **[`NgForm`](../api/angular2.common/NgForm-class.html) directive*
```
kwalrath: You didn't change this, but... why does this line have `*`s? Plus, they're mismatched!

kwalrath:
LGTM. I didn't try to generate the docs.

public/docs/dart/latest/guide/forms.jade
chalin: Oh, that needs to be fixed. Good catch.

PR #2186, opened 2016-08-24
docs(resources): use ng-href to avoid false broken link reports

chalin:


kwalrath:
LGTM. I'll just wait for the Travis build to finish before merging this...

PR #2187, opened 2016-08-24
docs(api): fix cross-language dropdown menu

chalin:
- Re-enable dropdown on main API page.
- Remove Dart choice from TS & JS API entry pages dropdown.

Contributes to #2173.


kwalrath:
LvGTM!

PR #2188, opened 2016-08-24
docs(ngmodule): Update ngmodule.jade

shairez:
not sure if it's a mistake or not, but sending this just in case


Foxandxss:
Thanks but there is a fix for that in a different PR.

PR #2189, opened 2016-08-24
docs(ngModule): fix a potential confusion

shairez:
I know it's petty stuff, but technically it does `import` the `FormsModule`.
It just doesn't add it to the imports array.

So to minimize the potential confusion about the overloaded "import" thing, I suggest the following change to the sentence


wardbell:
Thanks, Shai. I see the potential for confusion. I incorporated your quest for clarity in PR #2179

PR #2196, opened 2016-08-25
better compression of images using zopfli

adhintz:


Foxandxss:
What's the result of this? Some "before" and "after" size or something, but at least I don't see visual difference, which is good.

adhintz:
They're roughly half the file size of the original images. The compression
should be lossless (no visual difference).

wardbell:
Thanks!  We'll be reshooting many of our images in the near future. Maybe you can revisit in a month or so and help us again.

PR #2200, opened 2016-08-26
since RC5 the webpack prod build is defect

Martin-Wegner:
Issue https://github.com/angular/angular/issues/10618 recommends 'mangle: { keep_fnames: true, screw_ie8: true } for the webpack UglifyJs plugin

### DIFF
--- a/public/docs/_examples/webpack/ts/config/webpack.prod.js
+++ b/public/docs/_examples/webpack/ts/config/webpack.prod.js
@@ -26,8 +26,7 @@ module.exports = webpackMerge(commonConfig, {
     new webpack.optimize.DedupePlugin(),
     new webpack.optimize.UglifyJsPlugin({ // https://github.com/angular/angular/issues/10618
       mangle: {
-        keep_fnames: true,
-        screw_ie8: true
+        keep_fnames: true
       }
     }),
     new ExtractTextPlugin('[name].[hash].css'),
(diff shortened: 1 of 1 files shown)


Foxandxss:
Ugh, I would like not doing this, but if the fix comes in 2.0.1, I guess there is no other way around.

Foxandxss:
You don't really need the `screw_ie8` to fix the issue, just keep_fnames

Martin-Wegner:
That's right but Angular 2 only supports evergreen browsers, that means IE10+ :) And the screw_ie8 option basically disables hacks used to retain compatibility with IE<9. The drawback is, it also introduces bugs, see e.g.: https://github.com/mishoo/UglifyJS/issues/485. So I think it is OK...

Foxandxss:
That gives me 404.

Martin-Wegner:
Ups sorry, just copied it from: https://github.com/gruntjs/grunt-contrib-uglify/issues/40 but they disables the issues section... But you will find examples:
https://github.com/mishoo/UglifyJS2/issues/1204
https://github.com/mishoo/UglifyJS2/issues/409#issuecomment-229355111

Foxandxss:
Still, I managed to fix that error on my side without `screw_ie8`.

Martin-Wegner:
So we should remove this option, right?

Foxandxss:
yes

Martin-Wegner:
Done :)

PR #2202, opened 2016-08-26
chore(site): support site build w/o recompiling API docs

chalin:
If the site has already been built once, then rebuild the site without re-harp-compiling the API docs use:
```
gulp check-deploy --lang=''
```
Otherwise specify the languages whose API docs are to be rebuilt; as usual, omitting the `--lang` flag is equivalent to `--lang=all` which is equivalent to `--lang=‘ts|js|dart’`.

Other changes to gulp tasks:
- Renamed `serve` to `harp-serve`.
- Removed `check-serve`. Use `harp-compile` followed by `serve-www’ instead.
- `harp-compile` now also defaults to `--lang=all`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,7 @@ _.*
 **/resources/zips
 public/docs/xref-*.*
 _zip-output
-www
+www*
 npm-debug*.log*
 *.plnkr.html
 plnkr.html
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -120,22 +120,35 @@ var _styleLessName = 'a2docs.less';
 //    or a regex pattern to match any one of 'ts', 'js', or 'dart'.
 //    Default: 'ts|js' except for the "full site build" tasks (see below),
 //    for which it is 'all'.
-//
-var lang, langs, buildDartApiDocs = false;
+
+// langs and skipLangs partition ['ts', 'js', 'dart'].
+var lang, langs, skipLangs, buildDartApiDocs = false;
 function configLangs(langOption) {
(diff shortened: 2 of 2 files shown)


chalin:
@wardbell @Foxandxss @kwalrath : ready for review.

wardbell:
I like the idea. Will try it!

Foxandxss:
We used (or use) to have 20 tasks doing similar jobs, if we can reduce a bunch of them, I will be all in.

We have 3-4 that harpCompile, another that serves, another that compiles and serves...

kwalrath:
So much faster!

There's one issue with the output, though: The "Deploy to Firebase? (y/n):" question is obscured by the 3 new lines of "Rename" output. Could you move those lines to be before the deploy prompt? Here's what I see:

```
...
[10:11:29] compile ok
[10:11:29] running live server ...
prompt: Deploy to Firebase? (y/n): :  [10:11:29] Rename: mv public/docs/ts/latest/_api-tmp-hide-from-jade public/docs/ts/latest/api
[10:11:29] Rename: mv public/docs/js/latest/_api-tmp-hide-from-jade public/docs/js/latest/api
[10:11:29] Rename: mv public/docs/dart/latest/_api-tmp-hide-from-jade public/docs/dart/latest/api
n
[10:13:00] Not deploying
[10:13:00] Finished 'check-deploy' after 2.13 min
```

kwalrath:
The newest version works great for me! OK if I merge this, @wardbell & @Foxandxss?

PR #2209, opened 2016-08-26
docs(ngmodule): add plunker images

wardbell:
Added plunker images rather than just the frame.
Extended live-example directive to accept an optional image


Foxandxss:
What if we set a convention on the images path so if the directive finds an image in that concrete path, will use it and if not, will default to "an empty one".

wardbell:
@Foxandxss How will it know which one to use?

PR #2212, opened 2016-08-27
api-builder: replace underscores from jade filenames

### LINKED ISSUE #2184: [TS] API entry missing for http._createDefaultCookieXSRFStrategy
Visit https://angular.io/docs/ts/latest/api/http/index/ and click on the link to `_createDefaultCookieXSRFStrategy` which is found towards the end of the page. Result: 404, page not found.

The relevant code excerpt from `modules/@angular/http/http.ts` is:
```
/**
 * @experimental
 */
export function _createDefaultCookieXSRFStrategy() {
  return new CookieXSRFStrategy();
}
```
Notice that the function is exported.

Is the page for this entry being skipped because of the leading underscore? If so, then it should not appear in the index summary.

@ericjim @petebacondarwin 
cc @Foxandxss @kwalrath 

This is one of the broken links reported in #2150. (There is also a similar issue with `platform-browser/index/_WORKER_UI_PLATFORM_PROVIDERS-let.html`, but let's see how this issue is resolved first).

petebacondarwin:
Closes #2184

### DIFF
--- a/tools/api-builder/angular.io-package/index.js
+++ b/tools/api-builder/angular.io-package/index.js
@@ -114,8 +114,8 @@ module.exports = new Package('angular.io', [basePackage, targetPackage, cheatshe
 
   computePathsProcessor.pathTemplates.push({
     docTypes: EXPORT_DOC_TYPES,
-    getPath: function(doc) { return doc.moduleDoc.moduleFolder + '/' + doc.name.replace(/^_+/, '-') + '-' + doc.docType + '.html'; },
-    getOutputPath: function(doc) { return doc.moduleDoc.moduleFolder + '/' + doc.name.replace(/^_+/, '-') + '-' + doc.docType + '.jade'; }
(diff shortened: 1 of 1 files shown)


chalin:
Ah, nice and simple fix. Maybe it would be better, though, to replace only the first leading underscore with a minus? Although unlikely, this could avoid potential clashes between an entry for `__foo` and `_foo`.

LGTM (though I haven't tried it).

chalin:
cc @Foxandxss

petebacondarwin:
I fixed it up to replace each leading '_' with the same number of leading '-'

wardbell:
Please follow up with the Angular team. They should **not** be exporting public API members with leading underscore. Therefore, this appears to be a solution looking for a non-existent problem.

petebacondarwin:
Although it is an "experimental" export so perhaps they wanted it to look a "bit private".

petebacondarwin:
I think we should probably close this one now (@chalin?) given that @IgorMinar closed the original issue.

chalin:
Yes, that's fine.

Issue #2129, opened 2016-08-17
Router: document `canLoad` for lazy loaded modules


wardbell:
See https://github.com/angular/angular/issues/10877

c/ @brandonroberts

PR #2214, opened 2016-08-29
docs(router): Added and organized more router dev guide content

### LINKED ISSUE #2148: Router: Add better explanation of how feature module routes get combined together
Seems to be a point of confusion. People think adding a feature module with routing to `imports` makes the routes children of the `AppModule` routes when they actually just get added to the configuration.

Example gist explains what actually happens: https://gist.github.com/brandonroberts/f17493b416092055615b8fc0830c127a

brandonroberts:
Moved all heroes functionality into milestone 2
Crisis Center initial functionality is milestone 3
Admin feature module as milestone 4 including route guard examples
Updated milestone 5 to lazy load admin feature module
Added examples for CanLoad, CanActivateChildren guard, component-less routes
Added section on explanation of ActivatedRoute
Added section on relative navigation
Added section on animating route components

Closes #2148 #2129


johnpapa:
Should https://github.com/angular/angular.io/pull/2214/files#diff-4da8fde79cf7bf8486e1ccc80130599eR123 also mention that the order is important? If this is mentioned later, perhaps just a link to that.

public/docs/ts/latest/guide/router.jade
```diff
@@ -122,7 +126,7 @@ include ../_util-fns
 
 :marked
```
johnpapa: Should all references to `router` be changed to `featuremodulenameRouterModule` ? I think we can skip this for now and schedule this for a future PR. Thoughts?

brandonroberts:
I wrote up a section about the route order mattering in another PR, but I can add it to this one instead.

public/docs/ts/latest/guide/router.jade
```diff
@@ -331,9 +337,11 @@ figure.image-display
   * including a route parameter (passing the Hero `id` while routing to the "Hero Detail")
   * child routes (the *Crisis Center* has its own routes)
   * the `CanActivate` guard (checking route access)
+  * the `CanActivateChild` guard (checking child route access)
   * the `CanDeactivate` guard (ask permission to discard unsaved changes)
   * the `Resolve` guard (pre-fetching route data)
-  * lazy loading feature area modules
+  * lazy loading feature modules
+  * the `CanLoad` guard (check lazy loaded feature module access)
```
johnpapa: Should we word this such that it explains that we are checking if we should indeed load the lazy loaded assets? As currently worded, it implies we have access to the module, but not that the files will be retrieved via XHR in the browser. Just a thought

johnpapa:
Is putting it in a separate file really paving the way for lazy loading? We can still lazy load if the routes are defined in the same file as the module. no? Or did I misunderstand.

https://github.com/angular/angular.io/pull/2214/files#diff-4da8fde79cf7bf8486e1ccc80130599eR404

public/docs/ts/latest/guide/router.jade
```diff
@@ -676,12 +684,11 @@ figure.image-display
 
 :marked
   Now that we have routes for our `Heroes` module, we'll need to register them with the *Router*.
-  We'll import the *RouterModule* like we did in the root NgModule, but there is a slight difference here.
+  We'll import the *RouterModule* like we did in the `AppModule`, but there is a slight difference here.
   In our root routing setup, we used the static **forRoot** method to register our routes and application level
-  service providers. Since we are in a submodule, we'll use **Router.forChild** method to only register additional routes. We do this
-  because the *Router* will combine all the provided routes from the submodule together and build our configuration. This allows us to continue defining
-  our feature-specific routes without modifying our main route configuration.
+  service providers. Since we are in a feature module, we'll use **Router.forChild** method to only register additional routes.
```
johnpapa: I recommend calling more attention to this `forChild`. Perhaps make it a callout note

public/docs/ts/latest/guide/router.jade
brandonroberts: Yes. I'll update that with a follow-up PR and sweep the rest of the docs

public/docs/ts/latest/guide/router.jade
```diff
@@ -770,6 +777,42 @@ a#get-route-parameter
...
+    **`parent`**: an `ActivatedRoute` that contains the information from the parent route when using [child routes](#child-routing-component).
+
+    **`firstChild`**: contains the first `ActivatedRoute` in the list of child routes.
+
+    **`children`**: contains all the [child routes](#child-routing-component) activated under the current route.
+
+:marked
+  We import the `Router`, `ActivatedRoute`, and `Params` tokens from the router package.
+
++makeExcerpt('app/heroes/hero-detail.component.1.ts (activated route)', 'imports')
```
johnpapa: Should this ngOnInit use guard logic to make sure it is indeed convertible  to a number?

```
ngOnInit() {
  this.route.params.forEach((params: Params) => {
     let id = +params['id']; // (+) converts string 'id' to a number
     this.service.getHero(id).then(hero => this.hero = hero);
   });
}
```

public/docs/ts/latest/guide/router.jade
```diff
@@ -868,23 +899,192 @@ a#nav-to-list
 
 +makeExcerpt('app/heroes/hero-detail.component.1.ts (excerpt)', 'gotoHeroes', '')
 
-h3#merge-hero-routes Import hero module into root NgModule
+.l-main-section#optional-route-parameters
 :marked
-  Our heroes feature is ready, but application doesn't know about our heroes module yet.
-  We'll need to import it into the root NgModule we defined in `app.module.ts`.
+  ### Route Parameters
```
johnpapa: I think this title should be "Optional Route Parameters". I didn't change it because I wanted to make sure you agreed. The anchor also says optional.

public/docs/ts/latest/guide/router.jade
```diff
@@ -868,23 +899,192 @@ a#nav-to-list
...
+  The `id` value appears in the URL as (`;id=15;foo=foo`), not in the URL path.
+  The path for the "Heroes" route doesn't have an `:id` token.
+
+  The optional route parameters are not separated by "?" and "&".
+  They are **separated by semicolons (;)**
+  This is *matrix URL* notation &mdash; something we may not have seen before.
+
+.l-sub-section
+  :marked
+    *Matrix URL* notation is an idea first floated
```
johnpapa: I like this note/callout. Good job here. Should we also address any concern that since it is not HTML standard that it will still work in all browsers that A2 supports as A2 handles this?

public/docs/ts/latest/guide/router.jade
```diff
@@ -868,23 +899,192 @@ a#nav-to-list
...
+  The optional `foo` route parameter is harmless and continues to be ignored.
+
+h3#merge-hero-routes Import Hero feature module into AppModule
+:marked
+  Our heroes feature module is ready, but our application doesn't know about our heroes module yet.
+  We'll need to import it into the `AppModule` we defined in `app.module.ts`.
 
   Update `app.module.ts` as follows:
 
 +makeExcerpt('app/app.module.2.ts (heroes module import)', 'hero-import')
```
johnpapa: the `routing` variable is shown, but there is no import statement showing where it came from.

public/docs/ts/latest/guide/router.jade
```diff
@@ -1041,61 +1241,50 @@ a#child-routing-component
...
+  by default, the `Crisis Detail` component will be re-used as we select different crises. In
+  our `Hero Detail` route, the component was recreated each time we selected a different hero.
+
+  The child paths *extend* the path of their parent route unless the are _empty_.
 
   Normally paths that begin with `/` refer to the root of the application.
   Here they are appended to the path to the `CrisisCenterComponent`.
 
-  To write an URL that navigates to the `CrisisListComponent`, we'd append its child route path,  `/`,
+  To write an URL that navigates to the `CrisisCenterHomeComponent`, we'd append its child route path, `/`,
```
johnpapa: I found myself re-reading this slash section a few times to follow it. I don't have any good ideas on how to clarify it though :(

Just pointing it out as this was the first part of the chapter I had to re-read.

Loving this chapter!

johnpapa:
Many of the examples show an empty `providers: [ ]` array. I recommend removing the unused property except when it is being used to serve a purpose.

public/docs/ts/latest/guide/router.jade
```diff
@@ -1236,30 +1423,77 @@ a#can-activate-guard
...
 :marked
-  Next, we add a child route to the `crisis-center.routes` with the path, `/admin`.
+  Our initial admin routing configuration:
 
-+makeExcerpt('app/crisis-center/crisis-center.routing.3.ts (admin route)', 'admin-route-no-guard')
++makeExcerpt('app/admin/admin.routing.1.ts (admin routing)', 'admin-routes')
+
+h3#component-less-route <i>Component-Less Route</i>: grouping routes without a component
+:marked
+  Looking at our child route under the `AdminComponent`, we have a route with a **path** and a **children**
```
johnpapa: In this example, I get the point ... but could we not have also put the `AdminDashboardComponent` as a parent of these children and achieved the same effect?

johnpapa:
I like that we are showing the signature for `canActivate`, but it makes this harder to read and more daunting to the learner. I think we should remove the return type. THoughts?

https://github.com/angular/angular.io/pull/2214/files#diff-4da8fde79cf7bf8486e1ccc80130599eR1548

public/docs/ts/latest/guide/router.jade
```diff
@@ -1339,6 +1582,28 @@ a#can-activate-guard
...
+  protect child routes within our feature module.
+
+  Let's extend our `AuthGuard` to protect when navigating between our `admin` routes. First we'll open our
+  `auth-guard.service.ts` and add `CanActivateChild` interface to our imported tokens from the router package.
+
+  Next, we'll implement the `canActivateChild` method with takes the same arguments as the `canActivate` method,
+  an `ActivatedRouteSnapshot` and `RouterStateSnapshot`. The `canActivateChild` behaves the same way the other
+  guards do, returning an `Observable<boolean>` or `Promise<boolean>` for async checks and `boolean` for sync checks.
+
++makeExcerpt('app/auth-guard.service.3.ts (excerpt)', 'can-activate-child')
```
johnpapa: suggest removing the explicit return type of the CanActivateChild

johnpapa:
Consider adding a header to all code examples as it makes it easier for the learner to know what the example is referring to (file name or some concept)

johnpapa:
Not sure what interface you are referring to

https://github.com/angular/angular.io/pull/2214/files#diff-4da8fde79cf7bf8486e1ccc80130599eR1774

johnpapa:
In some places we spread, some we dont. It works either way as the tools are smart. But maybe we should call this out?

```
const appRoutes: Routes = [
  ...loginRoutes,
  ...adminRoutes
];
```

public/docs/ts/latest/guide/router.jade
```diff
@@ -1768,47 +1869,46 @@ a#fragment
...
 
-  If we look closer at the `loadChildren` string, we can see that it maps directly to our `crisis-center.module` file where we previously built
-  out our `Crisis Center` feature area. After the path to the file we use a `#` to denote where our file path ends and to tell the `Router` the name
-  of our `CrisisCenter` NgModule. If we look in our `crisis-center.module` file, we can see it matches name of our exported NgModule class.
+  If we look closer at the `loadChildren` string, we can see that it maps directly to our `admin.module.ts` file where we previously built
+  out our `Admin` feature area. After the path to the file we use a `#` to denote where our file path ends and to tell the `Router` the name
+  of our `AdminModule`. If we look in our `admin.module.ts` file, we can see it matches name of our exported module class.
 
-+makeExcerpt('app/crisis-center/crisis-center.module.ts (export)', 'crisis-center-module-export')
++makeExcerpt('app/admin/admin.module.ts (export)', 'admin-module-export')
```
johnpapa: Should we point out that we can `default export` too?

johnpapa:
I tried the `Resolver` code and ran into some issues. One of which is that if I use the API for the interface as shown in the API docs like this, it works:

`resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot) : Observable<any>|Promise<any>|any`

In your plunk you omit the `state` argument and it works for you. In my example it never called the function if I omitted it. I am using RC5. Is this a bug that is fixed in RC6 that you were referring to?

johnpapa:
Great chapter!

public/docs/ts/latest/guide/router.jade
brandonroberts: I think this was leftover from before modules came along and the process of lazy loading was loading the routes in a separate file from the components. I'll remove this text.

public/docs/ts/latest/guide/router.jade
brandonroberts: I don't think so. The service does a strict equality check so if a letter is passed in, it won't find a match

public/docs/ts/latest/guide/router.jade
brandonroberts: Added an additional note about browser support for the matrix notation by the router

public/docs/ts/latest/guide/router.jade
brandonroberts: Fixed

public/docs/ts/latest/guide/router.jade
brandonroberts: Yes, we could. I have a separate `AdminDashboardComponent` to have a default route for the admin routing. Otherwise we'd have to use a redirect to one of the other routes.

public/docs/ts/latest/guide/router.jade
brandonroberts: I agree about removing the return type. We are explaining the types of turns it expects. It would make the signatures in the examples shorter also.

public/docs/ts/latest/guide/router.jade
brandonroberts: We are leaning towards providing the export name here as `default export` isn't supported by AoT compilation yet.

brandonroberts:
@wardbell @Foxandxss I made the edits @johnpapa suggested, so this one should be ready to go once travis goes green.

public/docs/ts/latest/guide/router.jade
wardbell: I took a crack at clarifying this particular passage which is a head scratcher.

public/docs/ts/latest/guide/router.jade
```diff
@@ -1200,32 +1400,28 @@ h2#guards Route Guards
...
 
   1. [CanActivate](../api/router/index/CanActivate-interface.html) to mediate navigation *to* a route.
 
-  2. [CanDeactivate](../api/router/index/CanDeactivate-interface.html) to mediate navigation *away* from the current route.
+  2. [CanActivateChild](../api/router/index/CanActivateChild-interface.html) to mediate navigation *to* a child route.
 
-  3. [Resolve](../api/router/index/Resolve-interface.html) to perform route data retrieval before *before* route activation.
+  3. [CanDeactivate](../api/router/index/CanDeactivate-interface.html) to mediate navigation *away* from the current route.
+
+  4. [Resolve](../api/router/index/Resolve-interface.html) to perform route data retrieval before *before* route activation.
```
wardbell: Remove first "before" in "before *before*"

public/docs/ts/latest/guide/router.jade
wardbell: @johnpapa Apparently the spread operator is still required in routes. I tried removing them all and it wouldn't compile.

We are NOT showing `export default` because we can't make it work with AoT (yet).

wardbell:
Closed by c15a68e

PR #2225, opened 2016-08-31
Kw doc edits

kapunahelewong:
This includes proofer edits, updated screenshot and a rogue "we" in Security. One "we" remains but is in the context of "we who work on Angular". I can reword if this last "we" doesn't make the cut.

Thanks!


kwalrath:
Yikes, this touches lots of files! I'll take a look ASAP so things don't get clobbered.

PR #2244, opened 2016-09-01
docs(browser-support): document browser support and polyfills

mlaval:
This is a first try at documenting browser support an polyfills.
The content needs to be reviewed and validated.

/cc @wardbell @IgorMinar @naomiblack

### DIFF
--- a/public/docs/ts/latest/guide/browser-support.jade
+++ b/public/docs/ts/latest/guide/browser-support.jade
@@ -1,0 +1,191 @@
+:marked
+  # Browser support #  
+  Angular supports most recent browsers. This includes the following specific versions:
+
+table
+  tr
+    th Chrome
+    th Firefox
+    th Edge
+    th IE
+    th Safari
+    th iOS
+    th Android
+    th IE mobile
+  tr
+    td latest
+    td latest
+    td 14
+    td 11
+    td 9
+    td 9
+    td Marshmallow (6.0)
+    td 11
+  tr
+    td
+    td
+    td 13
+    td 10
+    td 8
+    td 8
+    td Lollipop<br>(5.0, 5.1)
+    td
+  tr
+    td
+    td
+    td
+    td 9
+    td 7
+    td 7
+    td KitKat<br>(4.4)
+    td
+  tr
+    td
+    td
+    td
+    td
+    td
+    td
+    td Jelly Bean<br>(4.1, 4.2, 4.3)
+    td
+    
+    
+:marked
+  To ensure this compatibility, the unit tests of the framework are run on all of them for every pull request. 
+  This is done as part of the continuous integration setup of the project, using [SauceLabs](https://saucelabs.com/) and [Browserstack](https://www.browserstack.com).
+  
+  # Polyfills #
+  Angular is built on the latest standards of the web platform.
+  Targeting such a wide range of browsers can be challenging as they may not support all the features of modern browsers.
+  
+  
+  To fill the gaps, the best solution is to use polyfills which "upgrade" old browsers by adding missing features.
+  For each browser, some polyfill may be required to run any application, and some may be needed only when some features are used.
+
+.alert.is-helpful 
+  span As a developer, depending on the browsers you target and on the features you use, you'll need to load a set of polyfills.
+  span  The tables below will help you making the right choices.
+  
+.alert.is-important 
+  span The polyfills documented here are the ones which were identified to run full Angular application.
+  span  Others may be needed if you decide to use extra features which are not supported everywhere.
+  span  On top of that, they will not magically transform an old and slow browser into a modern and fast one.
+  
+:marked
+  ## Mandatory polyfills ##
+  For each supported browser, these are the polyfills which are required to run an application:
+  
+table
+  tr
+    th Browsers (desktop & mobile)
+    th Polyfills required
+  tr
+    td Chrome, Firefox, Edge, Safari 9+
+    td None
+  tr
+    td Safari 7 & 8, IE10 & 11, Android 4.1+
+    td ES6
+  tr
+    td IE9
+    td ES6<br>classList
+    
+    
+:marked
+  ## Optional browser features to polyfill ##
+  In addition to the mandatory polyfills, some features of Angular may require some extra polyfills.
+  
+  For example, the animations use the standard web animation API, which is only available in Chrome and Firefox today.
+  In order to use this feature in other browsers, the correct polyfill is needed.
+  
+  Here are the features which may require additional polyfills:
+
+table
+  tr
+    th Feature
+    th Polyfill
+    th Browsers (desktop & mobile)
+  tr
+    td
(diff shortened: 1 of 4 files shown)


public/docs/ts/latest/guide/browser-support.jade
```diff
@@ -0,0 +1,190 @@
...
+    td
+    td
+    td
+    td
+    td Jelly Bean<br>(4.1, 4.2, 4.3)
+    td
+    
+    
+:marked
+  To ensure this compatibility, the unit tests of the framework are run on all of them for every single commit.
```
IgorMinar: "every single commit" sounds a bit dramatic and it's not really true. we run tests on PRs which can contain multiple commits.

public/docs/ts/latest/guide/browser-support.jade
```diff
@@ -0,0 +1,190 @@
...
+:marked
+  To ensure this compatibility, the unit tests of the framework are run on all of them for every single commit. 
+  This is done as part of the continuous integration setup of the project, using [SauceLabs](https://saucelabs.com/) and [Browserstack](https://www.browserstack.com).
+  
+  # Polyfills #
+  Angular is built on the latest standards of the web platform.
+  Targeting such a wide range of browsers can be challenging as they may not support all the functionalities of modern browsers.
+  
+  
+  To fill the gaps, the best solution is to use polyfills which "upgrade" old browsers by adding missing functionalities.
```
IgorMinar: functionalities -> features

public/docs/ts/latest/guide/browser-support.jade
```diff
@@ -0,0 +1,190 @@
...
+  For each browser, some polyfill may be required to run any application, and some may be needed only when some features are used.
+
+.alert.is-helpful 
+  span As a developer, depending on the browsers you target and on the features you use, you'll need to load a set of polyfills.
+  span  The tables below will help you making the right choices.
+  
+.alert.is-important 
+  span The polyfills described here under are the ones which, others may be required for your application.
```
IgorMinar: "here under" -> below

IgorMinar: "are the ones which, others may be required for your application" - what do you mean by this? I don't understand. can you rephrase?

public/docs/ts/latest/guide/browser-support.jade
```diff
@@ -0,0 +1,190 @@
...
+    td Safari 7 & 8, IE10 & 11, Android 4.1+
+    td ES6
+  tr
+    td IE9
+    td ES6<br>classList
+    
+    
+:marked
+  ## Optional polyfills ##
+  On top of the mandatory polyfills, some features of Angular may require some extra polyfills.
```
IgorMinar: "on top" -> "In addition"

public/docs/ts/latest/guide/browser-support.jade
```diff
@@ -0,0 +1,190 @@
...
+      a(href="../api/common/index/DecimalPipe-class.html") decimal
+      span  and 
+      a(href="../api/common/index/PercentPipe-class.html") percent
+      span  pipes
+    td Intl API
+    td All but Chrome, Firefox, Edge and IE11
+  tr
+    td 
+      a(href="../api/common/index/NgClass-directive.html") NgClass
+      span  on SVG elements
```
IgorMinar: is the double space after span intentional?

public/docs/ts/latest/guide/browser-support.jade
```diff
@@ -0,0 +1,190 @@
...
+    td MIT / Unicode licence
+    td 13.5KB
+  tr
+    td Web Animations
+    td
+      a(href="https://github.com/web-animations/web-animations-js") https://github.com/web-animations/web-animations-js
+    td Apache
+    td 14.8KB
+  tr
+    td Typed Arrays
```
IgorMinar: really? what requires typed arrays?

public/docs/ts/latest/guide/browser-support.jade
```diff
@@ -0,0 +1,190 @@
...
+  tr
+    td Safari 7 & 8, IE10 & 11, Android 4.1+
+    td ES6
+  tr
+    td IE9
+    td ES6<br>classList
+    
+    
+:marked
+  ## Optional polyfills ##
```
IgorMinar: "Optional browser features to polyfill" - isn't this more accurate? the list above doesn't list polyfills but feature that should be polyfilled.

public/docs/ts/latest/guide/browser-support.jade
IgorMinar: the same for Blob and FormData - they are not really needed unless you are using particular browser features in your app at which point you are responsible for figuring out what polyfills you need. none of these should be used directly by angular's code unless I'm missing something.

public/docs/ts/latest/guide/browser-support.jade
mlaval: Yes, for formating purpose

public/docs/ts/latest/guide/browser-support.jade
mlaval: We use typed arrays in http module, and we have tests in the same module that use Blob and FormData. That's why I documented this feature here.
I could be removed if too specific.

mlaval:
Thanks for the suggestions, I updated the PR

IgorMinar:
LGTM!

@wardbell can we merge this in?

wardbell:
Ill have a chance to review when I return on Monday. @foxandxss maybe you can get a jump on this and make sure it is ready to go. Thanks everyone!

Foxandxss:
Yes sir, will do today.

Foxandxss:
I would change the "ES6" to say "ES6 shim" (I know that perhaps your way is more correct, but thinking about the user perspective, I think they are more familiar with the "ES6 shim" term.

On the other hand, the first time is a bit overcomplicated. Instead of listing all versions one by one, can't we say IE9+, Android4.1+, etc?

mlaval:
"ES6 shim" is also the name of a well-known library, it could be confusing, especially since we suggest to use core-js at the end (to shim ES6 :D).
Anyway, it should be ES2015. So it could be changed to "ES2015 shim".

About the first table, the intention was to follow the format of a SauceLabs badge. To me, it looks better than saying: 
Angular supports most recent browsers. This includes the Chrome (latest), Firefox (latest), Edge 13+, IE 9+, Safari 7+, iOS 7+, Android 4.1+ and IE mobile 11.
It can definitely be changed.

Foxandxss:
> "ES6 shim" is also the name of a well-known library, it could be confusing, especially since we suggest to use core-js at the end (to shim ES6 :D).
Anyway, it should be ES2015. So it could be changed to "ES2015 shim".

Agreed.

> It can definitely be changed.

It is only my opinion, let see what the others say

IgorMinar:
I'm merging this because we need other pages to link to it, we can improve it further next week if needed

Issue #2231, opened 2016-08-31
Tour of Heroes - HTTP - create method for HeroService never defined


jzissu:
In the Tour of Heroes, the HTTP page has a section called "Add a hero" where the user makes a new button that will create a new hero using a name entered in an input field on the Heroes page of the application. However when following the steps in the tutorial it doesn't work because they never tell you to add the create method to the HeroService class, which means that when the Add button is clicked the function being called doesn't exist. Now looking at the bottom of the page will show the method, and looking at the Plunkr page also has the method, but you should add the create method to the explanation for that section.

bignimbus:
If no one else is working on this, I can submit a fix for this issue.

PR #2256, opened 2016-09-02
Correct some small errors in tutorial examples

bignimbus:
Addresses: https://github.com/angular/angular.io/issues/2231

And also a small issue where an `<h1>` tag is being included in a template where it was previously deleted.

At https://angular.io/docs/ts/latest/tutorial/toh-pt5.html:

Near the top, user sees:

screen shot 2016-09-02 at 6 08 15 pm

Since the user has moved the `<h1>` tag to another template, we should not see it here, below:

screen shot 2016-09-02 at 6 07 37 pm

Thanks in advance for your consideration of these tweaks.

### DIFF
--- a/public/docs/_examples/toh-5/ts/app/heroes.component.1.ts
+++ b/public/docs/_examples/toh-5/ts/app/heroes.component.1.ts
@@ -1,0 +1,96 @@
+// #docplaster
+// #docregion
+import { Component, OnInit } from '@angular/core';
+
+import { Hero } from './hero';
+// #docregion hero-service-import
+import { HeroService } from './hero.service';
+// #enddocregion hero-service-import
+
+@Component({
+  selector: 'my-app',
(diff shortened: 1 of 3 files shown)


Foxandxss:
Looks good to me, but that is a big new file just for the template.

bignimbus:
Makes sense, I can cut that down to just the essentials.

wardbell:
Thanks for taking this on. I appreciate the effort and your care.

I don't think we should add an entire file just to delete the `<h1>` . The maintenance burden is not worth it. You can see the problems already. For example, your version still has the `title` property which shouldn't be there. And all those irrelevant `#docregion` tags.

I think the world will survive an extra `<h1>`, especially if we tell folks to delete it.

Your other change to ToH-6 is important.

I incorporated these thoughts into #2264. Closing yours. Thanks.

PR #2265, opened 2016-09-03
chore: improve UX of build w/o API

chalin:
Just do a full build if no API docs currently exist rather than building the site and then aborting with an error message at the end of it all.


chalin:
cc @kwalrath @wardbell @Foxandxss - minor PR over the gulpfile.

kwalrath:
LGTM. It'd be nice if I could generate the Dart docs into a shared place, so that I'm the only one who has to generate those docs...

PR #2272, opened 2016-09-06
fix(footer): Footer logo is overflowing

LayZeeDK:
Footer logo was overflowing at ~960px wide viewport. Added `background-size: contain` (shorthand), and `max-width: 100%` to contain wrapper, and background image within grid.

### DIFF
--- a/public/resources/css/module/_footer.scss
+++ b/public/resources/css/module/_footer.scss
@@ -27,9 +27,10 @@
   }
 
   .logo-inverse-large {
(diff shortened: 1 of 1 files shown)


alexwolfe:
@LayZeeDK I've checked out this branch and compared before/after but don't see any difference. Can you include a screen shot of what exactly this is fixing using the current site vs. your branch? Thanks!

LayZeeDK:
@alexwolfe 

Chrome 53 beta 64-bit before:
image

Chrome 53 beta 64-bit after:
image


Edge 13 before:
image

Edge 13 after:
image

PR #2274, opened 2016-09-07
Update package.webpack.json

jgatjens:
Missing reflect-metadata module on package.json.


jgatjens:
I signed it!

Foxandxss:
Is not really missing, we are not using it in the example.

jgatjens:
if you try to run the test you will get the an error.

Foxandxss:
That is right. I will ask and then change it on test or revert the core-js.

Foxandxss:
This got fixed.

PR #2285, opened 2016-09-08
docs(quickstart): remove compiler-cli from dependencies

Foxandxss:
This seems to confuse people and we are actually not using it.

I know we are going to use it in the future for AoT but we can either bring it back when TS 2 is here (so new users don't get a lot of warnings) or just add the instructions to add it on the AoT cookbook.

### DIFF
--- a/public/docs/_examples/quickstart/ts/package.1.json
+++ b/public/docs/_examples/quickstart/ts/package.1.json
@@ -13,7 +13,6 @@
   "dependencies": {
     "@angular/common": "2.0.0-rc.7",
     "@angular/compiler": "2.0.0-rc.7",
-    "@angular/compiler-cli": "0.6.1",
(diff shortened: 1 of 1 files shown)


public/docs/_examples/quickstart/ts/package.1.json
```diff
@@ -13,7 +13,6 @@
   "dependencies": {
     "@angular/common": "2.0.0-rc.6",
     "@angular/compiler": "2.0.0-rc.6",
-    "@angular/compiler-cli": "0.6.0",
```
wardbell: What about the same line in `_examples/package.json`?

Foxandxss: Sure, why not. Doesn't hurt the readers, only us, but doesn't give us anything by having it there right now.

PR #2287, opened 2016-09-08
doc-gen(mergeDecoratorDocs): capture all the metadata docs

petebacondarwin:
Iterate through the docs and merge all the of metadata docs
for each decorator doc into the decorator doc and remove it
so that it is not rendered in a page of its own.

This is a first step in tidying up the decorator API docs.
With this change you will no longer see the "auxiliary" classes, such as
`ComponentMetadata`, `ComponentType` and so on in the API docs.

Instead these docs are now attached to the decorator doc, which can then
be accessed in templates for rendering the description and so on.


petebacondarwin:
CC @IgorMinar

IgorMinar:
looks good to me. now we need to adjust the templates to render the mergerged docs. anything else?

petebacondarwin:
Let's see if how far we can get by updating the templates. Obviously there are still issues around NgModule, etc

IgorMinar:
the pipe commit looks good.

tools/api-builder/docs-package/processors/mergeDecoratorDocs.js
```diff
@@ -21,7 +21,11 @@ module.exports = function mergeDecoratorDocs() {
         var makeDecorator = getMakeDecoratorCall(doc);
         if (makeDecorator) {
           doc.docType = 'decorator';
+          // get the type of the decorator metadata
           doc.decoratorType = makeDecorator.arguments[0].text;
+          // clear the symbol type named (e.g. ComponentMetadataFactory) since it is not needed
+          console.log(doc.symbolTypeName);
```
petebacondarwin: oops that needs to go

IgorMinar:
landed as  13878efca75a5c45f07ca06d42f8e030dd6693c4

PR #2289, opened 2016-09-08
chore(cheatsheet): use Jade to generate mirror page

chalin:
The page `docs/ts/latest/guide/cheatsheet.html` is a mirror of `docs/ts/latest/cheatsheet.html`. Use Jade instead of Harp to create the mirror.

cc @kwalrath - note that this has no visible effect on the generated page.


chalin:
cc @ericjim

kwalrath:
LGTM

PR #2296, opened 2016-09-09
Fix inline tag problems

### LINKED ISSUE #2294: `TypeError: Cannot read property 'baseName' of undefined` when linking to the lifecycle event dev guide
@petebacondarwin @IgorMinar When I want to link to the lifecycle event dev guide, I am getting the error below.

This is strange as other parts in our sources already refer to this guide successfully (see e.g. https://github.com/angular/angular/blob/master/modules/@angular/core/src/metadata.ts#L294).

Note: My links contain a `#` at the end, but the error is also present when I remove the `#`.

To reproduce: Clone [this PR](https://github.com/angular/angular/pull/11465) and replace `\@ with `@` in the file `lifecycle_hooks.ts`

```
TypeError: Cannot read property 'baseName' of undefined
    at Object._self.handler (/Users/tbosch/projects/angular.io/tools/api-builder/links-package/inline-tag-defs/linkDocs.js:49:76)
    at /Users/tbosch/projects/angular.io/node_modules/dgeni-packages/jsdoc/processors/inline-tags.js:229:41
```

petebacondarwin:
Fixes #2294

### DIFF
--- a/tools/api-builder/angular.io-package/processors/addJadeDataDocsProcessor.js
+++ b/tools/api-builder/angular.io-package/processors/addJadeDataDocsProcessor.js
@@ -95,7 +95,7 @@ module.exports = function addJadeDataDocsProcessor() {
               });
 
               // Remove line breaks, there should only be one tag
-              howToUse = howToUseArray[0].description.replace(/(\r\n|\n|\r)/gm,"");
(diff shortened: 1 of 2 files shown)


tbosch:
Landed in master.

PR #2301, opened 2016-09-09
chore: update to node 6

filipesilva:
The new `harp` version should allow us to move to the latest LTS node version, 6.

See https://github.com/sintaxi/harp/issues/571


Foxandxss:
There is also a file in the root for nvm.

filipesilva:
Good catch, updated.

PR #2305, opened 2016-09-10
update conference links - add ng-europe; remove AngularConnect

katcita:
This links the top banner to ng-europe instead of AngularConnect. Now that AngularConnect is sold out, I think it would be really helpful to link to ng-europe until the conf. Let me know if this could work. Thanks!

How it looks like right now:

oldbannerangulario

How it looks like with the ng-europe update:

image


katcita:
I signed it!

naomiblack:
merged, live.

PR #2307, opened 2016-09-11
fix(build-compile): allow skipping langs in a fresh build

filipesilva:
Currently, running ` gulp build-compile --dgeni-log=error --lang=''` with no previously full (all langs) build will result in these errors:

```
[17:52:57] Error: ENOENT: no such file or directory, stat 'D:\work\angular.io\www-backup\docs\ts\latest\api'
[17:52:57] Error: ENOENT: no such file or directory, stat 'D:\work\angular.io\www-backup\docs\js\latest\api'
[17:52:57] Error: ENOENT: no such file or directory, stat 'D:\work\angular.io\www-backup\docs\dart\latest\api'
```

This PR fixes that behaviour. This is especially relevant for setups where a full build is never created.

/cc @chalin

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1044,8 +1044,10 @@ function restoreApiHtml() {
     const relApiDir = path.join('docs', lang, vers, 'api');
     const wwwApiSubdir = path.join('www', relApiDir);
     const backupApiSubdir = path.join('www-backup', relApiDir);
-    gutil.log(`cp ${backupApiSubdir} ${wwwApiSubdir}`)
-    fs.copySync(backupApiSubdir, wwwApiSubdir);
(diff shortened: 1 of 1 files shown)


chalin:
As we discussed elsewhere: this failure is by design to ensure that @naomiblack and @kwalrath don't deploy a partial site unknowingly.

I'm ok with this fix if we we guard it with an extra command line flag, say:
```js
    if (fs.existsSync(backupApiSubdir) || argv.forceSkipApi !== true) {
     // ...
   }
```
So developers building a partial site (say, w/o Dart API docs) would use the `--force-skip-api` flag.

filipesilva:
Ah I didn't quite get that it was by design. I'll update with your recommendation.

filipesilva:
@chalin updated per your comment.

chalin:
LGTM

PR #2318, opened 2016-09-12
API doc-gen updates

petebacondarwin:
- Add an `@ngModule @angular/core` dgeni tag that can be used to document in what NgModule a directive is exported. This is a "multi" tag in that it can be used multiple times in a single doc, to indicate that a directive is exported from more than one class. This means that the property `ngModule` on the doc will be an array of strings, rather than a simple string.
- Fix the the paths to the API docs to remove the unwanted `index` and `-doctype` parts.

cc @IgorMinar 

TODO:
- at the very least we will need to update the directive template to display the ngModule
- think about if the change to the URLs of the API docs will have an impact from external sites linking to the docs

### DIFF
--- a/tools/api-builder/angular.io-package/index.js
+++ b/tools/api-builder/angular.io-package/index.js
@@ -101,7 +101,7 @@ module.exports = new Package('angular.io', [basePackage, targetPackage, cheatshe
   computePathsProcessor.pathTemplates.push({
     docTypes: ['module'],
     getPath: function computeModulePath(doc) {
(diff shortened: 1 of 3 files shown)


IgorMinar:
I cherry-picked ebc3757 to master.

the second commit will have to wait a bit because it requires all of the guide pages with links to be modified.

petebacondarwin:
@wardbell this can be closed because the changes are already in the new `angular/angular` doc gen tooling.

PR #2324, opened 2016-09-13
fix(api-builder): escape double-quotes in JSON output

petebacondarwin:
Fixes #2321 

@naomiblack and @IgorMinar

### DIFF
--- a/tools/api-builder/angular.io-package/templates/api-list-data.template.html
+++ b/tools/api-builder/angular.io-package/templates/api-list-data.template.html
@@ -7,7 +7,7 @@
       "docType": "{$ item.docType $}",
       "stability": "{$ item.stability $}",
       "secure": "{$ item.security $}",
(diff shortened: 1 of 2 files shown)


tools/api-builder/angular.io-package/templates/api-list-data.template.html
```diff
@@ -7,7 +7,7 @@
       "docType": "{$ item.docType $}",
       "stability": "{$ item.stability $}",
```
IgorMinar: both stability and security fields can contain free-form text. we should escape those too.

IgorMinar:
@naomiblack you missed the comment above. this solution is incomplete and the site will still break eventually.

IgorMinar:
oh.. I see. the json actually doesn't contain the user provided text, that is only used in non-json contexts. this change is fine then.

PR #2325, opened 2016-09-13
fix(api-builder): escape double-quotes in JSON output

petebacondarwin:
This builds on top of #2324 and fixes the problem mentioned at https://github.com/angular/angular.io/pull/2321#issuecomment-246597648

It does this by updating to a newly released version of dgeni-packages that copes with mulitple declarations for a symbol.


naomiblack:
merged.

PR #2330, opened 2016-09-14
chore(home): fix joyful-development.png import statement

### LINKED ISSUE #2085: Outdated image on main site.
In angular.io, on the Incredible Tooling section, the images shows Angular code before the RC. The package in the image shows angular2/core but this is no longer true, now it is @angular/core and it is missing a semicolon. Also the file is called app.ts, when according to the style guide it should be app.component.ts, finally there is no selector.

wardbell:
fixes #2085  Just the linked png, not the svg, jpg, or gif
c/ @naomiblack


naomiblack:
merged.

PR #2332, opened 2016-09-14
Fix github links

petebacondarwin:
The update to dgeni-packages led to a breaking change in the Nunjucks
templates, where macros are now isolated from their calling site.

This means that we must pass the `versionInfo` object through to the
macro rather than expecting it to be in scope already.


naomiblack:
merged.

PR #2336, opened 2016-09-14
chore(api-builder): use Jade include rather than Harp partial

chalin:
This change is in support of ng2.io. There is no net effect on ng1.io.


chalin:
@petebacondarwin - can you review?

cc @alexwolfe @ericjim @kwalrath @Foxandxss

naomiblack:
merged, live.

PR #2345, opened 2016-09-14
docs(aot): remove typescript instruction

Foxandxss:
Now that we have TS2 everywhere in our docs (and quickstart) we don't need to tell people to install typescript 2 by hand.

The good part is that installing TS2 by hand ends with a big red INVALID that can scare people off.


filipesilva:
lgtm

naomiblack:
merged, live.

PR #2353, opened 2016-09-15
chore: here we are

filipesilva:


filipesilva:
@Foxandxss is ok with me commiting these changes in his name.

Foxandxss:
I agree with this commits.

PR #2360, opened 2016-09-15
Update 2 Twitter Handels in harp.json

Rudimental:
Currently, the links to Twitter on Kapunahele Wong's and Ward Bell's bios both point to https://twitter.com/undefined. These links are accessible when their bios are opened. This PR adds their proper Twitter handle, so their respective links point to the right place.


Rudimental:
I signed it, googlebot.

PR #2308, opened 2016-09-11
fix(harp): temporarily use PR commit for js fix

filipesilva:
https://github.com/sintaxi/harp/pull/579 introduced a bug related to the new harp version where the `resources/js` wasn't copied on `gulp build-compile`.

This PR uses the commit in https://github.com/sintaxi/harp/pull/579 to fix this while Harp is being updated.

/cc @chalin

PR #2376, opened 2016-09-16
chore(harp): use official pre-release 0.21 of harp

chalin:
@filipesilva's harp patch was just merged so we can start using the official pre-release again (as we had tried to do in #2301, but were forced to revert in #2308).

cc @naomiblack

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "gulp-tslint": "^5.0.0",
     "gulp-util": "^3.0.6",
     "gulp-watch": "^4.3.4",
-    "harp": "git://github.com/filipesilva/harp.git#8da8d3497ddbfcbcbadd8be63e0fd731d7310cc4",
+    "harp": "0.21.0-pre.1",
     "html2jade": "^0.8.4",
     "indent-string": "^2.1.0",
     "jasmine-core": "^2.3.4",


package.json
```diff
@@ -48,7 +48,7 @@
     "gulp-tslint": "^5.0.0",
     "gulp-util": "^3.0.6",
     "gulp-watch": "^4.3.4",
-    "harp": "git://github.com/filipesilva/harp.git#8da8d3497ddbfcbcbadd8be63e0fd731d7310cc4",
+    "harp": "^0.21.0-pre.0",
```
filipesilva: I believe it's pre.1 instead, see https://github.com/filipesilva/harp/commit/bf9030cb1ac2a6035bae3da9afbdab4990011370#commitcomment-19049717

I would also advise to not use the caret, semver resolution falls back to alphabetical on these non-semver ones.

chalin: Yes, it is `-pre.1`.
> semver resolution falls back to alphabetical

Good to know. Though, I wouldn't expect more than 9 pre subreleases. You still want the caret removed?

chalin: Done.

filipesilva:
Lgtm

PR #2377, opened 2016-09-16
docs(search): proposal for a new search experience using DocSearch 📄

vvo:
Hi,

This PR is a proposal to change the default search feature of https://angular.io to [DocSearch](https://community.algolia.com/docsearch).

We believe this change offers a better search experience for AngularJS developers than the
current search feature.

DocSearch is built by [Algolia](https://www.algolia.com/). It's free. You do not have to maintain it. We built it as a community effort
to bring good documentation search to developer communities. It is installed on 160+ websites,
maybe you already used it while browsing https://facebook.github.io/react/ or http://vuejs.org/api/.

@maxiloc and @vvo (me) did the DocSearch integration for angular.io, we work at @algolia.

Here's a demo of how it looks:
angular io

Differences with the current search:
- Isolates searches by language. If you are on `/docs/dart` then you only get the relevant search results for this language.
- Typo tolerance. Write `ngnodule`, we correct that to `ngmodule`
- Direct access to corresponding paragraph if a query matches one. Example: `architecture components` will directly scroll to `guide/architecture.html#components`
- Wider search results style allowing for a more comfortable experience
- Performance. This is subjective but you can maybe tell from the GIF that we show results at the first keystroke which is not the case today

angular.io search index is replicated in 10 regions in the world like any DocSearch index, bringing good search performance to most users.

If the ranking does not meet your expectations, **we can tune it**. I am sure, as AngularJS developers, you already know what would be a perfect ranking/display for a good search. We can tune it in many ways but we believe it could be ok for a first start.

I hope you appreciate this PR and the work we put on making the search on https://angular.io/ AWESOME :)

### DIFF
--- a/public/_includes/_head-include.jade
+++ b/public/_includes/_head-include.jade
@@ -40,6 +40,7 @@ link(href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesh
 link(rel="stylesheet" href="/resources/css/vendor/icomoon/style.css")
 link(rel="stylesheet" href="/resources/css/vendor/animate.css")
 link(rel="stylesheet" href="/resources/css/main.css")
+link(rel="stylesheet" href="https://cdn.jsdelivr.net/docsearch.js/2/docsearch.min.css")
 
 <!-- MOBILE ICONS -->
 <link rel="apple-touch-icon" sizes="57x57" href="/resources/images/favicons/apple-touch-icon-57x57.png">
--- a/public/_includes/_scripts-include.jade
+++ b/public/_includes/_scripts-include.jade
@@ -37,6 +37,7 @@ script(src="/resources/js/directives/if-docs.js")
 script(src="/resources/js/directives/live-example.js")
 script(src="/resources/js/directives/ngio-ex-path.js")
 script(src="/resources/js/directives/scroll-y-offset-element.js")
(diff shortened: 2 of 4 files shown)


naomiblack:
cc'ing some interested parties... we've been looking at tuning search which has sorely languished these past few months. thanks for the PR. we'll have a look.

vvo:
Thanks @naomiblack, any feedback yet? You must be all busy with the official release eheh :)

xorgy:
This looks beautiful and a solution like this is the first thing I see people groan about when they visit the docs.

vvo:
Ping @naomiblack @alexwolfe @robwormald. I think you recently upgraded search with https://github.com/angular/angular.io/commit/5732b07423a4dedde76150bae90e4e08bcbcbd83. If this PR is no more relevant, feel free to close it.

Thanks

vvo:
If you also want a result page with DocSeach, I can provide it. Right now the result page on angulario shows both autocompletion menu and results at the same time:

2016-10-21-085254_1391x1154_scrot

Also when switching languages in the left menu, it does not refine the left search.

vvo:
Hey hi, any news here? Do you want to maybe either close or merge this? Thanks!

vvo:
Closing, if anyone from angular is interested in restarting this, just ping me here

PR #2378, opened 2016-09-16
docs(typescript): Fix name of typings config file

karptonite:
I believe name of the typings config file is typings.json, not typings.config.


karptonite:
I signed it!

PR #2380, opened 2016-09-16
udpate colors and contrast for accessibility improvements

alexwolfe:
@naomiblack @ericjim Updates to improvements to contrast for accessibility. Ready for review.


alexwolfe:
I signed it!

ericjim:
Very nice, thank you!

Foxandxss:
The sidebar is way way better. Did you change the h1 colors on teh guides? I can't see any changes using the devtools but I appreciate a change:



Left is new change. Those h1 doesn't pass the accessibility for the colors. Still, I don't see it changing in the browser console :/

Also, I don't see any color change on the text color (paragraphs) but I feel this one way better.

This [page](http://localhost:3000/docs/ts/latest/) is way better than before. The text is way small and the contrast doesn't work yet, but is a step in the right direction.

naomiblack:
merged, live. leaving open for discussion cc @alexwolfe

alexwolfe:
@Foxandxss Are you talking about the page titles in the hero? I made the opacity 100% instead of 87% which should hopefully solve the issue. But we could be talking about a different thing? Send me a screen shot of the culprit in slack and I'll make short work of it ;) 👍

alexwolfe:
@Foxandxss ahh, I see the problem. Something didn't come through right. Going to fix this and push again to this branch. They should def not be that light, something got messed up with the merge with master.

alexwolfe:
@Foxandxss ok, fixed :) they look the way the are supposed to now (like headings you can see). Thanks for the heads up, not sure how/why my original changes got squashed but we should be good to go now. 👍

Foxandxss:
it looks great now.

wardbell:
Made moot by new angular/angular/aio.

PR #2383, opened 2016-09-17
fix(about): dialog should not show twitter if empty

DevVersion:
- The bio directive accidentally shows the Twitter link in the dialog, even if it's undefined.
  
  > The directive is not able to track it, because Jade emits a `"undefined"` string, instead of an empty string.
  > 
  > `!!"undefined" ==> true` // Shows Twitter button
- Fix deprecation warning of Angular Material, where `md-dialog-actions` should be used instead of the class.

### DIFF
--- a/public/about/index.jade
+++ b/public/about/index.jade
@@ -22,7 +22,7 @@
     for person, name in bios
       if person.type == type
         .c3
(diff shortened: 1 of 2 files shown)


gkalpak:
To make things worse, there is an actual twitter account withe the username `undefined` :smiley:: https://mobile.twitter.com/undefined

chalin:
cc @alexwolfe @ericjim

naomiblack:
merged! sorry this took me so long. Thank you Paul!!!

DevVersion:
Thanks for merging :)

PR #2389, opened 2016-09-17
docs(lang/latest): parameterize landing page Jade by language

chalin:
ng2.io preparatory work.

cc @ericjim @Foxandxss


Foxandxss:
Are all the new links that appear on the js side (say AoT) having a "this is not here yet" ? Instead of a 404.

wardbell:
@Foxandxss Yes ... we don't want 404s when you toggle from one lang to another.

chalin:
FYI, added missing cookbook AoT chapter placeholders for JS and Dart in #2418.

PR #2387, opened 2016-09-17
docs: fix {ts,dart,js}/latest/_data.json

chalin:
Angular 2.0.0 is out so also cleaned up:
- Removed "Angular is a candidate release".
- Changed "API Preview" to "API Reference"

Fixes #2384.
Fixes #2385.
Contributes to #2407.

PR #2402, opened 2016-09-18
chore: rename Angular 2 to simply Angular

Foxandxss:
None of the dart files has been modified nor the _cache files (at least not intentionally).

I also left the _data.json files because there is a PR for them #2387


public/docs/dart/latest/guide/architecture.jade
```diff
@@ -6,7 +6,7 @@ block includes
   - var _at_angular = 'angular2'
```
wardbell: This should be "angular" as well, right?

PR #2418, opened 2016-09-20
docs(cookbook): don't use shredder on intro page

chalin:
Also added missing cookbook AoT chapter placeholders for JS and Dart, fixing 404s mentioned in #2389.


wardbell:
How did `// #docregion cookbook` creep into Jade pages? What was the intention?

Merging

chalin:
> How did // #docregion cookbook creep into Jade pages? What was the intention?
That was the "old way" that we shared prose (before using Jade extends and blocks).

Issue #2408, opened 2016-09-19
Transparent sidenav in narrow resolutions overlays with content


ericjim:
There appears to be a slight regression with the sidenav that is only noticeable when the view port is in a narrow resolution.

![screen shot 2016-09-19 at 10 48 31 am](https://cloud.githubusercontent.com/assets/1329167/18636405/aa51cc52-7e56-11e6-8752-b2e648414064.png)

I'll tag @alexwolfe, since I believe we changed the transparency of the site recently.

cc @chalin

PR #2427, opened 2016-09-21
fix: boost transparency of sidenav

ericjim:
Fixes: https://github.com/angular/angular.io/issues/2408

### Preview
screen shot 2016-09-20 at 8 55 39 pm

cc @chalin

### DIFF
--- a/public/resources/css/module/_side-nav.scss
+++ b/public/resources/css/module/_side-nav.scss
@@ -10,7 +10,7 @@
 */
 
 $sidenav: '.sidenav'; // Classname
-$sidenav-background: rgba($blue-grey-50, .56);
+$sidenav-background: rgba($blue-grey-50, 1);
 $sidenav-background-dark: $blue-grey-200;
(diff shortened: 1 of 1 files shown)


alexwolfe:
Believe it or not it will actually reduce the contrast ratio (since bg color will be a darker gray). So for accessibility, the color is better at .56. 

Any specific reason to change it (preference is a valid option btw :) )

ericjim:
The reason for change was to fix: https://github.com/angular/angular.io/issues/2408, though I get what you mean with the contrast ratio, it's really subtle.

alexwolfe:
@ericjim Ahhh, good catch. Can you simply apply it to the mobiles styles? Otherwise I can do it today (will be tinkering around today anyway). Let me know.

ericjim:
@alexwolfe How's that?

alexwolfe:
@ericjim perfect 👍 Thanks for fixing this!

PR #2439, opened 2016-09-22
docs(toh): Replaced window.history with location service

brandonroberts:
Also updated references of "Component Router" to "Router"

### DIFF
--- a/public/docs/_examples/toh-5/ts/app/hero-detail.component.ts
+++ b/public/docs/_examples/toh-5/ts/app/hero-detail.component.ts
@@ -1,10 +1,11 @@
 // #docplaster
 // #docregion , v2
-import { Component, OnInit } from '@angular/core';
+import { Component, OnInit }      from '@angular/core';
 import { ActivatedRoute, Params } from '@angular/router';
+import { Location }               from '@angular/common';
 
-import { Hero } from './hero';
-import { HeroService } from './hero.service';
+import { Hero }         from './hero';
+import { HeroService }  from './hero.service';
 
 @Component({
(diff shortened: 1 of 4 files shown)


public/docs/_examples/toh-5/ts/app/hero-detail.component.ts
```diff
@@ -38,7 +40,7 @@ export class HeroDetailComponent implements OnInit {
 
   // #docregion goBack
   goBack(): void {
-    window.history.back();
+    this.location.back();
```
chalin: Nice.

public/docs/_examples/toh-6/ts/app/hero-detail.component.ts
```diff
@@ -29,11 +31,11 @@ export class HeroDetailComponent implements OnInit {
   // #docregion save
   save(): void {
     this.heroService.update(this.hero)
-      .then(this.goBack);
+      .then(() => this.goBack());
```
chalin: Why this change? You are replacing a function g by an anonymous function which does nothing but call g. I.e., g and `() => g()` are equivalent, though the former is shorter.

brandonroberts: I was getting an error before that it `Cannot call location of null` error. Maybe it was a fluke on my part but I just tried it again and it worked correctly. I'll change it back to what it was.

chalin: Oops, your change is actually the right thing to do to ensure that `this` gets bound properly in `goBack`.

PR #2465, opened 2016-09-24
docs(lifecycle-hooks): make it easier to get to call sequence

wardbell:
Per Ben Lesh suggestion. Also converts away from "we"


blesh:
❤️

PR #2495, opened 2016-09-28
design(style): fix ignored bottom padding on Firefox

g0tsimid:
The bottom of the sidebar is obstructed by the TS/JS/Dart language picker in Firefox because the bottom padding amount is ignored. This is fixed by moving the padding into the margin of a child element.

### DIFF
--- a/public/resources/css/module/_side-nav.scss
+++ b/public/resources/css/module/_side-nav.scss
@@ -26,7 +26,6 @@ $sidenav-width: 240px;
 #{$sidenav} {
   background: $sidenav-background;
   box-shadow: 3px 0px 6px rgba($black, .24);
-  padding-bottom: 72px;
   width: $sidenav-width;
   bottom: 0px;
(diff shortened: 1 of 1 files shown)


chalin:
cc @ericjim

flx-l:
+1

Since padding-bottom is ignored by many non-webkit browsers on elements with overflow:auto, the "Cookbook" section is not reachable if the "Advanced" section is selected.

jddxf:
+1

And do you guys think the sidebar scrolling seems wired? When the scrollbar is at the top or the bottom, a part of it is covered by the top nav or the bottom dropdown. 

Perhaps we can remove both `padding-top` and `padding-bottom` of `.sidenav` and set `top` and `bottom` instead. That way we don't even need to care about the ignored bottom padding.

wardbell:
Made moot by replacing this site with angular/angular/aio

PR #2233, opened 2016-08-31
chore(test): add protractor4

filipesilva:
This PR updates Protractor to from version 3 to 4, which is made inTypescript and includes typings.

Since Protractor4 includes typings, this solves an issue we had with the Upgrade Adapter chapter where Jquery's `$` typings conflicted with Protractors `$`. This issue prevented us from having them in the same node package and was the main motivator for the `_protractor` folder.

Thus, this PR removes the `_protractor` folder and adds Protractor configuration and helpers directly to `_examples`, which solves some editor tooling issues we had because of the indirect path structure for e2e tests.

Boilerplate files were moved from `_examples` to `_examples/_boilerplate` as well, so that they are contained there and also to be able to use a `tsconfig.json` for Protractor in `_examples` (this was needed for editor tooling purposes). Gulp tasks and plunker generation were adjusted to use this directory.

Some of the custom logic we had in `protractor.config.js` was moved into `protractor-helpers.ts` since we were attaching it to the `browser` variable, which is now typed and thus threw errors when accessing these custom members. Our custom `sendKeys` was also removed since it's not needed anymore.

End result is that we should have full type safety and editor tooling support in Protractor.

Note: previous tooling added a git ignored `tsconfig.json` besides each `e2e-spec.ts` file, which should now be removed otherwise editor tooling will not work properly. Overall, after this PR it is a good to run `git clean -fdx` to 'reset' the repo and re-setup our infrastructure.

/cc @wardbell @chalin @Foxandxss this PR includes some structural changes that I'd like you to look at and see if you approve.

PR #2496, opened 2016-09-28
docs(toh-6-aot): add aot to toh-6

thelgevold:
Tor says: "... AoT CB is ready to go"


thelgevold:
This is still pending https://github.com/angular/in-memory-web-api/pull/36, but here is my first stab at supporting both AoT and JiT in the same code base. We can improve on this, but this is at least a starting point.

AoT
I have added a new project with no source code under toh-6-aot. I went in this direction to allow both compilation modes to be active at the same time (with unit tests).

There is no source code in the new project, but there is a new index.html file pointing to the generated bundle. The bundle is generated from the source in the original toh-6 folder. Trigger a build by running `build:toh-aot` from the `toh-6-aot/ts`

JiT
The JiT code still lives in its original location, but I have added two sets of main.ts and tsconfigs over there. Trigger a JiT build from `toh-6/ts` by running `build:toh-jit`.

@wardbell @johnpapa @brandonroberts @Foxandxss @filipesilva

thelgevold:
I made a few tweaks based on discussions with @wardbell 

Instead of adding a new tsconfig.json I have made a few tweaks to the shared config
This also includes an upgrade of in-mem-web-api

The remaining part for this cookbook is to add a section to the prose.

filipesilva:
@thelgevold I've reviewed your files and have a few questions:
- you mentioned that after talking with Ward, you had opted to tweaking the shared tsconfg instead of adding a new one. In this PR I still see 3 tsconfig total, can you elaborate?
- can this approach handle lazy loaded modules?

I'm also thinking of generalizing this approach to run AoT over all of our examples in a separate e2e CI instance.

thelgevold:
@filipesilva The third tsconfig is obsolete now, so I have deleted it.
Currently this does not address lazy loading of modules. I was thinking we could address that in a subsequent revision.

filipesilva:
Addressing lazy loading separately is fine by me, just wanted to confirm.

thelgevold:
@wardbell I have updated the prose and removed the WIP part of the title. 
This PR is ready for review.

public/docs/_examples/toh-6/ts/app/app.module.ts
```diff
@@ -12,7 +12,7 @@ import { HttpModule }    from '@angular/http';
 
 // #enddocregion v1
 // Imports for loading & configuring the in-memory web api
-import { InMemoryWebApiModule } from 'angular-in-memory-web-api';
+import { InMemoryWebApiModule } from 'angular-in-memory-web-api/in-memory-web-api.module';
```
brandonroberts: Is there a reason we have to import this way instead of just importing from `angular-in-memory-web-api`?

public/docs/ts/latest/cookbook/aot-compiler.jade
```diff
@@ -301,3 +302,130 @@ a#source-code
...
+  `tsconfig.json (JiT),
+   tsconfig-aot.json (AoT)`
+)   
+
+:marked
+  ### External AoT Libraries
+
+  Tour of Heroes has an external dependency on the `angular-in-memory-web-api` library.
+
+  We can't expect all libraries to be released with AoT in mind, but `angular-in-memory-web-api` has already been AoT compiled and released with the necessary dependencies to integrate it with another AoT compiled application.
```
brandonroberts: I think we should expect this since one non-AoT provided library prevents AoT altogether. Maybe that is addressed in a separate cookbook on creating a third party library.

thelgevold: Hm..Does this mean that any third part angular lib has to be AoT-ed for AoT to work? We are integrating rxjs which is non-AoT, but that is not Angular....

public/docs/_examples/toh-6/ts/app/app.module.ts
brandonroberts: Ok. It should be noted as a caveat to rollup if this is not the way it has to be done normally unless we figure out a way to use the barrel export.

thelgevold: I wonder if this is happening because the code is already converted to JavaScript by the time it enters the rollup world. I guess the barrel is transpiled as well, but maybe Rollup isn't considering the convention of using barrels...

public/docs/ts/latest/cookbook/aot-compiler.jade
thelgevold: I think I agree that we should have a separate cookbook talking about creating third party libraries. I removed the sentence that said we shouldn't expect AoT since it seems to be a prerequisite.

wardbell:
I am rebasing and reviewing and will force push so beware

wardbell:
I reviewed it and cleaned up some prose. There are outstanding issues:

* [ ] Doesn't compile anymore since the Protractor PR changed our setup. Mostly duplicate typing errors
* [ ] No obvious way for readers to get the samples.
* [ ]  I don't think Rollup works in Windows when its configuration file is named `rollup.js`; as I recall it should be named something else like `rollup.config.js`.

filipesilva:
@thelgevold there were a few setup changes in https://github.com/angular/angular.io/pull/2233, which went in yesterday. 

When rebasing pay special attention to `public/docs/_examples/_boilerplate/tsconfig.json`, which is the new master tsconfig for example. It contains the `"types": []` property that forces `tsc`/`ngc` to ignore any `@types` present in `node_modules`. 

Since you create new tsconfigs for your examples, you must add this property, otherwise you will get a lot of duplicate identifier errors.

thelgevold:
The only item I am not clear on is the `No obvious way for readers to get the samples` part. 
Have to think about improvements here.

thelgevold:
I have fixed the typings issue and renamed the rollup.js file to rollup-config.js

johnpapa:
should it be `rollup.config.js.` or `rollup-config.js` ? which works in @wardbell 's case?

thelgevold:
I think either will work. My impression is that on windows it can't be named the same as the binary (rollup)

thelgevold:
I switched to using the AppRouterModule and there were no issues with AoT

johnpapa:
We should make sure the cookbook prose and code samples are updated too.

public/docs/_examples/package.json
```diff
@@ -74,6 +74,7 @@
     "raw-loader": "^0.5.1",
     "rimraf": "^2.5.4",
     "rollup-plugin-commonjs": "^4.1.0",
+    "source-map-explorer": "^1.3.2",
```
johnpapa: do we need this in the examples? its cool but are we using it here?

thelgevold: Yeah, the prose talks about how to use it, so I think it makes sense to install it with the sample

wardbell:
I've extended the gulpfile's e2e facility so it can run the same e2e.specs against the AoT build of an app when `~/aot/index.html` is defined.

It's primitive and rigid but it works as seen in the toh-6 in his  PR. That means no need for a separate aot project folder (toh-6-aot is gone).

The "trick" is that the AoT version of the `index.html` is in the `/aot` folder and to run it you serve from within `/aot`!

Still don't have a good way to demonstrate this to our readers. The instructions in the prose describe what to do if you've cloned the repo.

PR #2516, opened 2016-09-30
docs(rc4-to-rc5): change cleanup

Foxandxss:


wardbell:
I think we just remove this file altogether.

PR #2526, opened 2016-10-02
docs: fixed path typo in tutorial(Routing, HTTP)

timxzz:
The path to `app.component.css` should start from current directory, not `app/`, since the file is already in `app/`


timxzz:
I signed it!

wardbell:
Good catch!

PR #2563, opened 2016-10-07
rename Developer Guides to Guide

ericjim:
We are using these files to generated the sidenav entries in ng2.io, however, they don't seem to have a net effect on the sidenav in ng1.io

cc @chalin @kwalrath


chalin:
As follow-up: note that "Developer Guides" isn't used in the sidenav anymore. Since the site refresh by Alex, we've been using "Guide".

PR #2566, opened 2016-10-07
Wolfe api filter updates

alexwolfe:
@naomiblack This one is ready to go, please review and merge. Thanks!

PR #2569, opened 2016-10-07
docs(api): fix Dart API reference page

chalin:
Dart API reference page is blank following the merge of #2566. This minor fix adapts to the new API filter criteria.

Also, for Dart we don’t show the statuses at all since none of the API entries are tagged with a status.

cc @kwalrath @naomiblack @alexwolfe @ericjim

### DIFF
--- a/public/resources/js/directives/api-list.js
+++ b/public/resources/js/directives/api-list.js
@@ -28,7 +28,7 @@ angularIO.directive('apiList', function () {
       '    <div class="overlay" ng-class="{ visible: $ctrl.showTypeMenu === true }" ng-click="$ctrl.toggleMenu(\'type\')"></div>' +
       '  </div>' +
       '        ' +
-      '  <div class="form-select-menu">' +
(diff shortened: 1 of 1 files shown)


public/resources/js/directives/api-list.js
```diff
@@ -76,7 +76,7 @@ angularIO.directive('apiList', function () {
         { cssClass: 'function', title: 'Function', matches: ['function'] },
         { cssClass: 'enum', title: 'Enum', matches: ['enum'] },
         { cssClass: 'type-alias', title: 'Type Alias', matches: ['type-alias'] },
-        { cssClass: 'const', title: 'Const', matches: ['var', 'let', 'const'] }
+        { cssClass: 'const', title: 'Const', matches: ['const', 'var', 'let'] }
```
chalin: This change is necessary so that `'const'` elements get picked up in line 215; because `'const'` is used as the canonical representative for the matches array elements.

PR #2588, opened 2016-10-12
Upgrade swiftype to 2.0 library and added a results page for searches

alexwolfe:
@naomiblack Ok, this should upgrade swiftype and create a results page for us. Please note, Angular conflicts with the in page modal popup which is why I chose to create a separate layout and page for the results.


naomiblack:
merged. live!

PR #2596, opened 2016-10-13
docs(attribute-directives): reword a bit

### LINKED ISSUE #2459: Typos in attribute directives guide
TS version:

> We'll add an `import` statement to fetch the 'Highlight' directive and, added that class to the `declarations` NgModule metadata so that Angular will recognize our directive when it encounters `myHighlight` in the template.

"and, added" should probably be "and add". Also the phrasing of "`declarations` NgModule metadata" seems off - maybe "NgModule metadata's `declarations`" instead? 

Dart version has some similar issues.

Edit:

Up a little there's also this:

> After the `@Directive` metadata comes the directive's controller class, which contains the logic for the directive. We export \`HighlightDirective\` to make it accessible to other components.

Seems like the backticks might have been escaped by accident?

Foxandxss:
Fixes #2459

### DIFF
--- a/public/docs/ts/latest/guide/attribute-directives.jade
+++ b/public/docs/ts/latest/guide/attribute-directives.jade
@@ -99,7 +99,7 @@ block highlight-directive-1
 p
   | After the #[code @Directive] metadata comes the directive's controller class, which contains the logic for the directive. 
   +ifDocsFor('ts')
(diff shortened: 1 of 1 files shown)


Foxandxss:
Oh, #2508 has some of this changes, so after that gets merged, I will refactor this.

wardbell:
Closed because incorporated in Kapunahele's PR #2508

PR #2601, opened 2016-10-13
Update styles.css link in Routing tutorial

tiangolo:
The current link in the Routing tutorial leads to a 404 error.

This update fixes that to point to the correct GitHub "raw" file link. Nevertheless, in the "Quickstart" the link takes to the normal GitHub page, not the "raw" file.

---

Also, it isn't very clear that for relative "asset" loading (html and css files) it's needed to set the `moduleId` variable in every `@Component` declaration to `module.id`. Otherwise the loading fails. (It took me some time to discover that).

As in:

```ts
@Component({
  moduleId: module.id,
  selector: 'my-dashboard',
  templateUrl: 'dashboard.component.html',
  styleUrls: ['dashboard.component.css']
})
export class DashboardComponent implements OnInit {

...
```

Should I try to add that to the guide to make it more clear in another PR?

### DIFF
--- a/public/docs/ts/latest/tutorial/toh-pt5.jade
+++ b/public/docs/ts/latest/tutorial/toh-pt5.jade
@@ -878,7 +878,7 @@ block css-files
 
 +makeExcerpt('styles.css (excerpt)', 'toh')
 
-- var styles_css = 'https://raw.githubusercontent.com/angular/angular.io/master/public/docs/_examples/styles.css'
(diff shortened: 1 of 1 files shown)


tiangolo:
I signed it!

Foxandxss:
Thanks. The moduleId is being fixed in a different PR.

tiangolo:
Thanks! That was a fast merge :tada: ! 

And thanks for the clarification about the `moduleId`.

PR #2610, opened 2016-10-15
removed navigation when only once announcement is present

alexwolfe:
@naomiblack Ok, here's a fix for the banner when only one item is present. :)


naomiblack:
live, merged! thanks :)

PR #2612, opened 2016-10-15
Rephrase the confusing text for modules loader

vadimk:


Foxandxss:
Old, Quickstart is completely different now.

PR #2613, opened 2016-10-16
docs(glossary): fix app module filename

simonxca:
Small error I noticed while going through docs.
As stated throughout angular.io, `AppModule` should be in `app.module.ts`


simonxca:
Signed it

Foxandxss:
Thank you

PR #2629, opened 2016-10-18
docs(animation): add note about aliases

Foxandxss:


public/docs/ts/latest/guide/animations.jade
```diff
@@ -179,6 +179,14 @@ figure
   are different on enter and leave: the element enters from the left
   and leaves to the right.
 
+.l-sub-section
+  :marked
+    This two animations transitions are really common, so Angular provides an alias for them:
```
wardbell: These two common animations have their own aliases:

PR #2632, opened 2016-10-19
[WIP] docs(cb-http-interception): add Http interception cookbook

wardbell:
**Work in Progress - do not merge**

Todo:
- [ ] Confirm this is a good approach with DevRel
- [ ] Write prose

### DIFF
--- a/public/docs/_examples/cb-http-interception/ts/app/http-interceptor.ts
+++ b/public/docs/_examples/cb-http-interception/ts/app/http-interceptor.ts
@@ -1,0 +1,127 @@
+/////// Experimental Generic HttpInterceptor ////////
+
+import { Injectable } from '@angular/core';
+
+import { Connection, ConnectionBackend, Http,
+         ReadyState, Request, RequestOptions,
+         Response, XHRBackend } from '@angular/http';
+
+import 'rxjs/add/observable/throw';
+
+import { of }         from 'rxjs/observable/of';
+import { Observable } from 'rxjs/Observable';
+import { Subject }    from 'rxjs/Subject';
+
+type interceptorFn = (act: HttpInterceptorActivity) => HttpInterceptorActivity;
+
+@Injectable()
+/**
+ * Http Backend Interceptor raises request and response activity events.
+ * Request interceptors can modify the request before it is sent (e.g, add headers) and can
+ * set the activity response to bypass XHR and return the response immediately.
+ * Response interceptors post-process responses before passing them to`http`.
+ */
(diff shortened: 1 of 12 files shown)


bennadel:
@wardbell so I've been staring at this code for a while and I think I'm having some issues reconciling the NG2 approach here with the existing NG1 approach. If we look at the NG1 code, there are a few differences:

1. Request / Response interceptors are different. Or at least, they can be defined as separate properties on the interceptor, which means the user-code itself doesn't need to differentiate between the incoming/outgoing when processing actions.
2. The interceptors are setup during the "config" phase.

Now, I know we don't really have a "config" phase in NG2, so maybe putting the setup in the root-component is the way to go. After all, it is the root of the application. 

The other thing that strikes me is the use of `.subscribe()` as a means to *alter* an action. Something about that feels strange. When I think about "subscribe" in other aspects, such as RxJS streams, it makes me think that I have access to the "result" of something. In this case, however, the subscribe is to the in-process workflow. Just feels a little odd.

I'm still noodling on this, but just some thoughts I had.

bennadel:
Going back to the config phase, I almost wish we had some sort of DI token for these things. Like:

```js
{
    provide: HTTP_INTERCEPTORS,
    multi: true,
    useClass: MyHttpInterceptor
}
```

... where that class could implement something more akin to the NG1 API:

```js
class MyHttpInterceptor implements HttpIncterceptor {
    public function request( config ) {}
    public function requestError( rejection ) {}
    public function response( config ) {}
    public function responseError( rejection ) {}
}
```

Then, maybe this class could also add another "Subject" for the root-node to subscribe to for the logging.

wardbell:
@bennadel Ok ... I agree that my previous approach was unintuitive.  So I've completely rewritten it. What do you think now?

wardbell:
Angular is adding HTTP interception; no point in this one.

public/docs/_examples/cb-http-interception/ts/app/http-interceptor.ts
```diff
@@ -0,0 +1,127 @@
...
+  addRequestInterceptor(fn: interceptorFn) {
+    this.requestInterceptors.push(fn);
+  }
+
+  addResponseInterceptor(fn: interceptorFn) {
+    this.responseInterceptors.push(fn);
+  }
+
+  addResponseErrorInterceptor(fn: interceptorFn) {
+    this.responseInterceptors.push(fn);
```
d-nation: For anyone who is using this until HTTP Interceptors land... this line should read:
`this.responseErrorInterceptors.push(fn);`

PR #2652, opened 2016-10-22
docs(security): improve xsrf description and add it to http chapter as well

wardbell:
Improves the English description of XSRF in the Security chapter and adds a shorter version (with references) to the Http chapter where developers would expect to find it..

@mprobst Please review and confirm that the `XSRFStrategy` override, described in the original text, actually works in AoT:
```
{ provide: XSRFStrategy, useValue: new CookieXSRFStrategy('myCookieName', 'My-Header-Name') }
```
I fear that it might not because it involves executing code (newing the class).  It may be necessary to do this:
```
export const xsrfStrategy  = () =>  new CookieXSRFStrategy('myCookieName', 'My-Header-Name');
...
{ provide: XSRFStrategy, useFactory: xsrfStrategy  }
```
Please confirm which approach is best.


johnpapa:
Perhaps it would be worth mentioning why the malicious code cant simply read the cookie and do the same thing.

public/docs/ts/latest/guide/security.jade
```diff
@@ -200,11 +200,12 @@ h2#http HTTP-level vulnerabilities
 
 h3#xsrf Cross-site request forgery
 :marked
-  In a cross-site request forgery, an attacker tricks the user into visiting a
-  _different_ page and has them, for example, submit a form that sends a request to your application's
-  web server. If the user is logged into your application, the browser will send authentication
-  cookies, and the attacker could&mdash;for example&mdash;cause a bank transfer in the user's name with
-  the right request.
+  In a cross-site request forgery (CSRF or XSRF), an attacker tricks the user into visiting
+  a different web page with malicious code that secretly sends a request to your application's web server.
```
mprobst: this might be clearer if you give the websites names, e.g. `evil.com` and `example.com`.

public/docs/ts/latest/guide/security.jade
```diff
@@ -200,11 +200,12 @@ h2#http HTTP-level vulnerabilities
...
 h3#xsrf Cross-site request forgery
 :marked
-  In a cross-site request forgery, an attacker tricks the user into visiting a
-  _different_ page and has them, for example, submit a form that sends a request to your application's
-  web server. If the user is logged into your application, the browser will send authentication
-  cookies, and the attacker could&mdash;for example&mdash;cause a bank transfer in the user's name with
-  the right request.
+  In a cross-site request forgery (CSRF or XSRF), an attacker tricks the user into visiting
+  a different web page with malicious code that secretly sends a request to your application's web server. 
+  If the user is logged into your application, the browser sends the request along with authentication cookies.
```
mprobst: it sends authentication cookies along with the request, right?

public/docs/ts/latest/guide/security.jade
```diff
@@ -200,11 +200,12 @@ h2#http HTTP-level vulnerabilities
...
+  An _unprotected_ server can't tell the difference between a legitimate request and a forged request.
+  
+  In this manner, for example, a knowledgable attacker could transfer money from the user's account to the attacker's account.
```
mprobst: ... `, if the bank's website does not protect itself against XSRF.`?

public/docs/ts/latest/guide/security.jade
```diff
@@ -213,27 +214,36 @@ h3#xsrf Cross-site request forgery
...
+  sets a header named `X-XSRF-TOKEN` with the value of that cookie. 
+  
+  The server must do its part by setting the
+  initial `XSRF-TOKEN` cookie and confirming that each subsequent state-modifying request
+  includes a matching `XSRF-TOKEN` cookie and `X-XSRF-TOKEN` header.
 
-  CSRF tokens should be unique per user and session, have a large random value generated by a
-  cryptographically secure random number generator, and expire.
+  XSRF/CSRF tokens should be unique per user and session, have a large random value generated by a
+  cryptographically secure random number generator, and should expire after a reasonably short period of time.
```
mprobst: maybe suggest a day or so

public/docs/ts/latest/guide/security.jade
```diff
@@ -213,27 +214,36 @@ h3#xsrf Cross-site request forgery
   each API request, the server then validates the client by checking that the token is sent back,
   usually in an HTTP header called `X-XSRF-TOKEN`.
 
-  The Angular `http` client has built-in support for this technique. The default
-  `CookieXSRFStrategy` looks for a cookie called `XSRF-TOKEN` and sets an HTTP request header named
-  `X-XSRF-TOKEN` with the value of that cookie on every request. The server must set the
-  `XSRF-TOKEN` cookie and validate the response header for each state-modifying request.
+  Angular's `http` has built-in support for the client-side half of this technique in its `XSRFStrategy`. 
+  The default `CookieXSRFStrategy` is turned on automatically.
+  During each HTTP request, the `CookieXSRFStrategy` looks for a cookie called `XSRF-TOKEN` and
```
mprobst: The `during` is a bit misleading, this all happens before the request is sent. Maybe:

```
When sending an HTTP request, `CookieXSRFStrategy` looks for a cookie called `XSRF-TOKEN` and sets a header named `X-XSRF-TOKEN` with the value of that cookie. 
```

public/docs/ts/latest/guide/security.jade
```diff
@@ -213,27 +214,36 @@ h3#xsrf Cross-site request forgery
...
-  { provide: XSRFStrategy, useValue: new CookieXSRFStrategy('myCookieName', 'My-Header-Name')}
-  { provide: XSRFStrategy, useClass: MyXSRFStrategy}
+  { provide: XSRFStrategy, useClass: MyXSRFStrategy }
 
 :marked
   For information about CSRF at the Open Web Application Security Project (OWASP) see
-  [Cross-Site Request Forgery (CSRF)](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_%28CSRF%29) and
-  [Cross-Site Request Forgery (CSRF) Prevention Cheat Sheet](https://www.owasp.org/index.php/CSRF_Prevention_Cheat_Sheet). The Stanford University
-  paper [Robust Defenses for Cross-Site Request Forgery](https://seclab.stanford.edu/websec/csrf/csrf.pdf) is also a rich source of detail.
+  <a href="https://www.owasp.org/index.php/Cross-Site_Request_Forgery_%28CSRF%29" target="_blank">Cross-Site Request Forgery (CSRF)</a> and
```
mprobst: any particular reason why you're changing the link syntax here?

public/docs/ts/latest/guide/security.jade
```diff
@@ -213,27 +214,36 @@ h3#xsrf Cross-site request forgery
...
+  <a href="https://www.owasp.org/index.php/CSRF_Prevention_Cheat_Sheet" target="_blank">Cross-Site Request Forgery (CSRF) Prevention Cheat Sheet</a>.
+  The Stanford University paper 
+  <a href="" target="_blank">Robust Defenses for Cross-Site Request Forgery</a> is a rich source of detail.
```
mprobst: no link?

public/docs/ts/latest/guide/security.jade
```diff
@@ -213,27 +214,36 @@ h3#xsrf Cross-site request forgery
...
+
+  See also Dave Smith's masterful, easy-to-understand
```
mprobst: I'm happy that you like this, but maybe a more neutral tone of voice would be appropriate in our docs?

public/docs/ts/latest/guide/server-communication.jade
```diff
@@ -628,6 +629,26 @@ block wikipedia-jsonp+
       You added the `debounceTime`, `distinctUntilChanged`, and `switchMap` operators to the RxJS `Observable` class
       in `rxjs-operators` as [described above](#rxjs).
 
+a#xsrf
+.l-main-section
+:marked
+  ## Guarding against Cross-Site Request Forgery
+
+  In a cross-site request forgery (CSRF or XSRF), an attacker tricks the user into visiting
```
mprobst: why do we repeat the same content here? should this link to the original doc, so we don't have two sources of truth that can diverge?

mprobst:
Re whether this works with AoT, it should. The static reflector understands class instantiations and string literals, AFAIK.

public/docs/ts/latest/guide/security.jade
wardbell: Yes. We want links to external sites to appear in a new browser tab by default. Can't specify that in markdown

public/docs/ts/latest/guide/security.jade
wardbell: oops

public/docs/ts/latest/guide/server-communication.jade
wardbell: Thought about that. We need some context here.  I only repeat the intro and this is material that shouldn't change even if our implementation does. I didn't think making up a new summary description would make a difference.

wardbell:
@johnpapa I didn't want to get into this depth in our Angular docs. That's why we provide links. The real answer is that it has nothing to do with reading the cookie. In fact, the key to the XSRF attack is that the browser can be counted on to forward the cookie.

The key to the technique, as I understand it, is that the XSRF attack cannot set the request header! Knowing the cookie is useless; the server rejects the request if the cookie (whatever it is) doesn't match the header ... and the attacker can't set the header at all.

In any case, I really don't want to go into it.

wardbell:
@mprobst I wasn't sure about AoT's ability to instantiate with string literals. Not sure what the rules are. Good to know. Thanks. And thanks for your other comments which I will incorporate.

public/docs/ts/latest/guide/server-communication.jade
wardbell: @mprobst @johnpapa Changes have  been made. I'll merge in a day or two if you have no more objections.

wardbell:
@mprobst @johnpapa Changes have  been made. I'll merge in a day or two if you have no more objections.

wardbell:
Thanks @mprobst @johnpapa. I you have more changes, time for a new PR

mprobst:
@wardbell please note that most folks don't work on weekends, so might be nice to leave a workday at least for comments. I think we also do mandatory code reviews in most Angular repos now, whether that's useful here for docs is probably your call.

PR #2657, opened 2016-10-23
Typo fix 'nestled'=>'nested'

SamuelMarks:
Technically 'nestled' still makes sense in this context, but is a little too Christmas-y! - I assume you meant 'nested' :P


wardbell:
I meant "nestled" but I have to agree, in hindsight, that my verb choice was a bit to intimate :)  Accepting your PR

PR #2669, opened 2016-10-25
updates to improve search results in Swiftype

alexwolfe:
@naomiblack This should improve a few things: 
- Exclude all navigation from being included in search results
- Label results clearly with the section the come from (ex. API, GUIDE, COOKBOOK). 

Please review and merge. Thanks!


alexwolfe:
One this is merge, give me a heads up and I can do a reindexing.

PR #2691, opened 2016-10-28
chore: remove bootstrap package

Foxandxss:
The main reason behind this change is to not confuse the end user with "Why do we have bootstrap as a dependency"? "Does Angular depend on bootstrap"?

The idea was to replicate the minimum css needed from bootstrap into those two forms guides but I think it is better for the time being to use a cdn (I saw that phonecat, AKA upgrade guide uses a cdn with bootstrap already) and then change it later.


wardbell:
I merged companion PR #270 in angular/quickstart

PR #2695, opened 2016-10-30
docs(quickstart): quickstart text rev + drop app.module.1

wardbell:
The main text change is to create the component _before_ the ngModule so that the reader isn't going ngModule -> component -> ngModule to add the declarations. This refactor also eliminates need for `app.module.1.ts`

### DIFF
--- a/public/docs/ts/latest/quickstart.jade
+++ b/public/docs/ts/latest/quickstart.jade
@@ -4,7 +4,7 @@ block includes
   - var _prereq = 'Node.js and npm'
   - var _angular_browser_uri = '@angular/platform-browser-dynamic'
   - var _angular_core_uri = '@angular/core'
-  - var _stepInit = 3 // Step # after NgModule step
+  - var _stepInit = 1 
   - var _quickstartSrcURL='https://github.com/angular/quickstart/blob/master/README.md'
 
 //- TS/Dart shared step counter
@@ -42,10 +42,10 @@ h2 Build this application!
 
 :marked
   - [Prerequisite](#prereq): Install #{_prereq}.
-  - [Step 1](#create-and-configure): Create and configure the project.
-  <span if-docs="ts"><li>[Step 2](#ngmodule): Create your application.</span>
-  - [Step !{step++}](#root-component): Create a component and add it to your application.
-  - [Step !{step++}](#main): Start up your application.
+  - [Step !{step++}](#create-and-configure): Create and configure the project.
+  - [Step !{step++}](#root-component): Create a component.
+  <span if-docs="ts"><li>[Step !{step++}](#ngmodule): Create the application's _NgModule_.</span>
+  - [Step !{step++}](#main): Start with _main_.
   - [Step !{step++}](#index): Define the web page that hosts the application.
   - [Step !{step++}](#build-and-run): Build and run the application.
   - [Step !{step++}](#make-changes): Make some live changes.
@@ -64,7 +64,7 @@ block setup-tooling
     in a terminal window.
 
 .l-main-section
-h1#create-and-configure Step 1: Create and configure the project
+h1#create-and-configure Step !{step++}: Create and configure the project
 
 - var _package_and_config_files = _docsFor == 'dart' ? 'pubspec.yaml' : 'configuration files'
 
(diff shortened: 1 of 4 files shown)


naomiblack:
Do not merge -- discussing with Ward.

public/docs/ts/latest/quickstart.jade
```diff
@@ -154,89 +154,90 @@ block install-packages
...
-    as well as [`RouterModule`](../latest/api/router/index/RouterModule-class.html) and
-    [`HttpModule`](../latest/api/http/index/HttpModule-class.html). These are introduced in the
-    [Tour of Heroes Tutorial](./tutorial/).
 
 .l-main-section
 h1#root-component Step !{step++}: Create a component and add it to your application
 :marked
   Every Angular application has at least one component: the _root component_, named `AppComponent`
-  here.
+  by convention.
```
naomiblack: Using "here" is a deliberate choice (earlier drafts did say "by convention". We don't want to impose a convention on the user that Angular itself doesn't enforce. 

What a developer calls the root component is their call.

public/docs/ts/latest/quickstart.jade
```diff
@@ -154,89 +154,90 @@ block install-packages
...
 
         - a *selector* that specifies a simple CSS selector for an HTML element that represents
           the component.
         - a *template* that tells Angular how to render the component's view.
+      +ifDocsFor('ts')
+        - *backticks* (`) surrounding the template HTML.
+        Use backticks instead of quote marks (' or ") 
+        so the HTML can span multiple lines as the template grows.
+
     * **A component class** that controls the appearance and behavior of a view
```
naomiblack: This is useful to know, but belongs in the TOH rather than here. The QuickStart is just a quick limited scope demo. We'll teach concepts later.

public/docs/ts/latest/quickstart.jade
```diff
@@ -154,89 +154,90 @@ block install-packages
...
-    from `@angular/platform-browser` to the `imports` array.
-
-    This is the smallest amount of Angular that is needed for a minimal application to run in the
-    browser.
-
-    The QuickStart application doesn't do anything else, so you don't need any other modules. In a real
-    application, you'd likely import [`FormsModule`](../latest/api/forms/index/FormsModule-class.html) 
-    as well as [`RouterModule`](../latest/api/router/index/RouterModule-class.html) and
-    [`HttpModule`](../latest/api/http/index/HttpModule-class.html). These are introduced in the
-    [Tour of Heroes Tutorial](./tutorial/).
```
naomiblack: @robwormald @StephenFluin Ward is proposing a restructure to create the component before the module in this doc flow. I think we did it the other way for pedagogical reasons... Can you have a look and comment please?

StephenFluin: To conceptual journey we want to take developers on is very important. It's important that developers separate the steps of creating an "app" (create a blank module), then create the component and add it to their module. You have to visit the module twice as part of the quickstart, but it trains the developer that "create a component, add it to your module" which they will be expected to do hundreds of times. We should keep 1. create blank module, 2. create component and add it to module

public/docs/ts/latest/quickstart.jade
wardbell: >What a developer calls the root component is their call.
The docs, the styleguide and the cli all create and refer to `AppModule` and `AppComponent`. These are not _my_ conventions. They are _our_ conventions. 

As with _any_ conventions you are free to create your own, but you would be wise to do so only with a good reason.

public/docs/ts/latest/quickstart.jade
wardbell: As I have been told, the QuickStart is not a "conceptual journey". It is supposed to be a clear, straightforward path to a start.

This is a confusing path here, creating first one module and then another.

public/docs/ts/latest/quickstart.jade
wardbell: Frankly this one sentence doesn't "teach concepts" more than _any of the many other sentences around it_.  I don't understand why the other statements are valuable and this one is not.

We can delete the comment. If we do I suggest we remove the backticks too  ... even though experience teaches us that it is easier to add HTML to the template when you start with backticks.

public/docs/ts/latest/quickstart.jade
naomiblack: agreed, but we would rather not make it a requirement called out this strongly.

wardbell:
We're going a different path.

PR #2701, opened 2016-10-31
clicking on Books & Training sends user to Education

ikilled:
When user clicks in Books & Training item in footer the website should take them to Education section (anchor) in the middle of the page, not to the top where Development section is.


ikilled:
I signed it!

PR #2711, opened 2016-10-31
feat: ng2 announcement bar

ericjim:


public/_includes/_scripts-include.jade
```diff
@@ -43,7 +43,6 @@ script(src="/resources/js/directives/scroll-y-offset-element.js")
 
 if ng2
   app-root
-  if current.path[0] !== 'index'
     script(src="/inline.js")
     // script(src="/styles.bundle.js")
     script(src="/main.bundle.js")
```
chalin: Pls unindent lines 46-48 so that they are at the same indentation level as `app-root`.

chalin: Actually, this might actually be ok. Leave as is for now. I'll experiment later.

Issue #998, opened 2016-03-22
ToH should use anchors + routerLink instead of router.navigate almost everywhere


IgorMinar:
For example in https://angular.io/docs/ts/latest/tutorial/toh-pt5.html

The example code right now looks like this:

``` html
<div *ngFor="#hero of heroes" (click)="gotoDetail(hero)" class="col-1-4">
```

``` js
gotoDetail(hero: Hero) {
  let link = ['HeroDetail', { id: hero.id }];
  this._router.navigate(link);
}
```

This code is problematic because of the following reasons:
- the navigation via the div is not accessible (to screenreaders and users using keyboard for navigation)
- it's not possible to left-click+cmd to open the link in a new tab/window
- it's not possible to right-click on an item and copy the link
- SEO also suffers because crawlers don't see a link so they can't index the transition to the new state

All these problems can be avoided (and the code can be made shorter) if `routerLink` directive is used instead:

``` html
<a *ngFor="#hero of heroes" [routerLink]="['HeroDetail', { id: hero.id }]" class="col-1-4">
```

This code will generate a link that has all the good properties listed above, while fully preserving the functionality of the app.

wardbell:
That is great information and I'd like to capture that in the router chapter as well!

The accessibility arguments for `<a>` are winners (and I wonder what the material design people are doing about this).

### RouterLink or (click) binding?

I need to discuss with John why we used `(click)` to delegate to a component navigation method rather than using `RouterLink`. That could be intentional under _some circumstances_: maybe you do NOT want the user to easily open it in a different tab; I've seen resistance to that AND to deep linking among some of our IdeaBlade clients.  

If that's what you want, fine. But you should KNOW that you're making that choice deliberately ... and we'd say so in the docs.

### Plunker sadness

I'm somewhat challenged by my present inability to demonstrate the "open in another tab" feature from within plunker. Works find when the app is run independently.  *Within plunker*, we seem to be generating a URL with a base path that is different from the base path of the source tab. 

In this snapshot, the top url is the source tab and the bottom url is the generated url for "Narco"

![image](https://cloud.githubusercontent.com/assets/129061/14119807/cb745048-f5a4-11e5-8e60-32dfb34e3541.png)

Therefore opening the link in a new tab results in a 404.  Works fine if I click-in-place. Maybe someone has an idea?

IgorMinar:
Can you share the link to the plunker that is broken in this way?

IgorMinar:
[...]

PR #2718, opened 2016-11-01
docs(toh-5): dashboard uses [routerLink] bindings #998

wardbell:
Per issue #998, switch from imperative routing to anchors with `[routerLink]` bindings in the `DashboardComponent`.

Kept the button-click imperative navigation in `HeroesComponent` mainly to show how to do that when you need imperative navigation but also because this use case is more button-like than anchor-like.

@chalin Modified the DART example to match. I _think_ I did it right but would be great if you confirmed.


chalin:
googlebot, you are mistaken, I did author the commit.

chalin:
I did author the commits. Googlebot!

wardbell:
Thanks, @chalin. Merging. You'll come back latter with the Dart fixes and remove the entry from ` bad-code-excerpt-skip-patterns.txt`.

chalin:
Follow-up is #2744.

PR #2721, opened 2016-11-01
[ng2] Fix announcement thrashing

ericjim:
fixes: https://github.com/angular/angular.io/issues/2712

Note: the slide indicators are tied to ngOnInit, which means they won't appear until the component is initialized.

### DIFF
--- a/public/_includes/_hero-home.jade
+++ b/public/_includes/_hero-home.jade
@@ -6,8 +6,8 @@ header(class="background-sky l-relative")
         a(href="/docs/ts/latest/quickstart.html" class="hero-cta md-raised button button-large button-plain" md-button) Get Started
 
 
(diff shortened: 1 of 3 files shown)


public/_includes/_hero-home.jade
```diff
@@ -6,8 +6,8 @@ header(class="background-sky l-relative")
         a(href="/docs/ts/latest/quickstart.html" class="hero-cta md-raised button button-large button-plain" md-button) Get Started
 
 
-    announcement-bar
-      .announcement-bar-slide.clearfix
+    announcement-bar.announcement-bar
```
chalin: Given that the styles have been changed. Could we just drop the `.announcement-bar` and adjust the styles to apply to the `announcement-bar` tag?

Issue #2709, opened 2016-10-31
[ng2] resources page: port resources page to angular 2


ericjim:

ericjim:
I will close this issue since we've ported the page into angular 2, however, we will be able to make this page a static page. Opens https://github.com/angular/angular.io/issues/2734

PR #2726, opened 2016-11-01
[ng2] feat: ng2 resources page

ericjim:
fixes https://github.com/angular/angular.io/issues/2709, but we will not use the ng2 version yet. We will attempt to build this page statically.


src/app/resources-page/resources-page.component.ts
```diff
@@ -0,0 +1,71 @@
...
+  ngOnInit() {
+    window.addEventListener('scroll', this.changeScrollPos.bind(this), false);
+  }
+
+  ngOnDestroy() {
+    window.removeEventListener('scroll', this.changeScrollPos.bind(this), false);
+  }
+}
+
+function flattenFirebaseObject(result): ResourceGroup[] {
```
ericjim: Before we were leveraging the Firebase SDK to extract the data (and a pipe to parse the data). Rather than doing that, I simply fetch the firebase json file on npm postinstall, and flatten the file to conform to `ResourceGroup[]`.

ericjim: Use this function for implementing the page statically.

PR #2732, opened 2016-11-01
[ng2] add support menu toggling; exclude all A1 & md1 scripts

### LINKED ISSUE #2703: [ng2] sidenav language drop-down doesn't work

chalin:
Support menu toggling w/o using A1 controller. Menus affected:
- Main menu
- Sidenav
- Language delection menu.

With this commit, we are able to exclude all A1 and MD1 scripts.

Fixes #2703.

### DIFF
--- a/public/resources/js/util.js
+++ b/public/resources/js/util.js
@@ -83,10 +83,9 @@ var NgIoUtil = (function () {
         return NgIoUtil._exampleName;
     };
 
-    NgIoUtil.showHideMenu = function (menu) {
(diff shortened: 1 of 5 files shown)


public/resources/js/util.js
```diff
@@ -83,5 +83,14 @@ var NgIoUtil = (function () {
         return NgIoUtil._exampleName;
     };
 
+    NgIoUtil.toggleMenu = function (menu) {
```
ericjim: Very nice

Issue #2730, opened 2016-11-01
[ng2] restylize language dropdown menu


chalin:
The language dropdown menu works w/o md2 but the formatting is off.
With md1 the menu appears like:
![screen shot 2016-11-01 at 3 57 59 pm](https://cloud.githubusercontent.com/assets/4140793/19910802/8cbf6d1c-a04c-11e6-83a4-0ae0450893b2.png)

W/o md it appears like:
![screen shot 2016-11-01 at 3 57 49 pm](https://cloud.githubusercontent.com/assets/4140793/19910797/7dc382da-a04c-11e6-9afe-6780822be143.png)

PR #2737, opened 2016-11-02
[ng2] add missing uppercase text in language selection dropdown

ericjim:
fixes https://github.com/angular/angular.io/issues/2730


chalin:
Was that not needed before because of your use of md1?

PR #2741, opened 2016-11-02
use SimpleChanges instead of SimpleChange

maxisam:
` {[propertyName: string]: SimpleChange}` ->  `SimpleChanges`


wardbell:
Cool. That didn't exist when the chapter was first written. But you missed a few :-)  Got 'em all in #2742.  Thanks!

PR #2747, opened 2016-11-04
chore: don't shred _boilerplate files nor dart build files

chalin:


chalin:
@filipesilva - finally, I think this will fix my problems with the shredder processing Dart build files! W/o this change, some folders would have 1000 files processed in parallel.

gulpfile.js
```diff
@@ -808,7 +808,7 @@ gulp.task('_harp-compile', function() {
 
 gulp.task('_shred-devguide-examples', ['_shred-clean-devguide', '_copy-example-boilerplate'], function() {
   // Split big shredding task into partials 2016-06-14
-  var examplePaths = globby.sync(EXAMPLES_PATH+'/*/', {ignore: ['/node_modules']});
+  var examplePaths = globby.sync(EXAMPLES_PATH+'/*/', {ignore: ['**/_boilerplate']});
```
filipesilva: Why is `/node_modules` not ignores anymore?

filipesilva: We discussed offline and turns out the folder still exists.

PR #2758, opened 2016-11-05
docs(cb-third-party-lib) create a third party lib

thelgevold:
WIP. Do not merge


thelgevold:
Here is my Third-Party-Library cookbook.

The library is published with support for AoT and JiT builds. 

AoT files:: JS with esm modules, .d.ts and metadata.json, html, and css files

JiT: files: umd bundle with inlined templates and css. 

The html/css inlining approach is currently using a script borrowed from material 2.

The sample library is published to node_modules via a custom script triggered from add-example-boilerplate 

@wardbell @robwormald @filipesilva @Foxandxss

public/docs/ts/latest/cookbook/third-party-lib.jade
```diff
@@ -0,0 +1,247 @@
...
+
+  Publishing plain JavaScript with typings and meta data allows the consuming application to remain agnostic of the library's build environment.
+
+<a id="tree-shaking"></a>
+:marked
+  ## Preparing the library for Tree Shaking
+
+  In addition to supporting AoT, the library code should also be "Tree Shakable". 
+  
+  Thee Shakers work best with `ES2015` JavaScript.
```
Foxandxss: Tree

public/docs/ts/latest/cookbook/third-party-lib.jade
```diff
@@ -0,0 +1,247 @@
...
+
+  The entry point to the module is configured using the `module` attribute in `package.json`. In this case `module` is defined as `"module": "index.js"`.
+
+<a id="jit"></a>
+:marked
+  ## Supporting JiT
+
+  AoT compiled code is the prefered format for production builds, but due to the long compilation time, it may not be practical to use AoT during development.
+
+  To create a more flexible developer experience, a JiT compatible build of the library should be published as well. The format of the JiT bundle is `umd`, which stands for Universal Module Definition. Shipping the bundle as `umd` ensures compatability with most common module loading formats.
```
Foxandxss: compatibility

public/docs/ts/latest/cookbook/third-party-lib.jade
```diff
@@ -0,0 +1,247 @@
...
+)(format='.')  
+:marked
+  Generate the `umd` bundle by running `node_modules/.bin/rollup -c rollup-config.js`.
+
+<a id="publish"></a>
+:marked
+  ## Publish
+
+:marked
+  `Rollup` wil create the `umd` bundle, but but prior to bundling, all external templates or css files must be inlined.
```
Foxandxss: No will (or wil :P)

Foxandxss: Double but.

Foxandxss:
I am more worried in teaching Webpack as well. There are a lot of libraries being published with a Webpack tooling and the would need guidance as well.

thelgevold:
@Foxandxss Thanks for the review. I have made tweaks based on your feedback.

I am unsure about the Webpack part though. My thinking is that the most important take-away from this is learning what is required to be AoT compatible. Specifically, which files to include etc. 

I do agree that webpack seems to be very popular these days though. I have also seen feedback that people would like to see more webpack stuf in our AoT docs. 

What I find appealing about Rollup is that the configuration is simpler and requires less intensive tech stack explanations.

However, I want to try to offer the most useful content to the users...

@wardbell @robwormald Any feedback on whether is makes sense to include webpack here vs updating our existing webpack guide?

Foxandxss:
Leave the webpack aside I guess. I would like to think about some ideas but I need to run all of them through Ward first.

public/docs/ts/latest/cookbook/third-party-lib.jade
```diff
@@ -0,0 +1,247 @@
...
+  
+  A version of the library intended for JiT compiled applications will also be included. 
+
+  The code for the Hero-Profile library can be found below.
+
+  To support both AoT and JiT compilation there are two different tsconfig files.
+  
+  `tsconfig.json` for JiT compilation and `tsconfig-aot.json` for AoT compilation.
+
++makeTabs(
```
filipesilva: `index.ts` is mentioned elsewhere in the prose but not shown here.

public/docs/ts/latest/cookbook/third-party-lib.jade
```diff
@@ -0,0 +1,247 @@
...
+  `cb-third-party-lib/hero-profile/hero-profile.module.ts,
+   cb-third-party-lib/hero-profile/tsconfig-aot.json,
```
filipesilva: Would it be possible to actually only use one `tsconfig.json`, but invoke it with overrides for the changed props? 

For instance `ngc --module=es2015`.

public/docs/_examples/cb-third-party-lib/hero-profile/hero-profile.component.ts
```diff
@@ -0,0 +1,14 @@
+// #docregion
+import { Component, Input } from '@angular/core';
+
+import { Hero } from './hero';
+
+@Component({
+  moduleId: module.id,
+  selector: 'hero-profile',
+  templateUrl: 'hero-profile.component.html',
+  styleUrls: ['hero-profile.component.css']
```
filipesilva: https://github.com/angular/angular.io/pull/3071 changed all paths to be relative, these should be as well.

public/docs/ts/latest/cookbook/third-party-lib.jade
```diff
@@ -0,0 +1,247 @@
...
+   cb-third-party-lib/ts/tsconfig.json,
```
filipesilva: Can you make `tsconfig.1.json` for this chapter with the config you want to show and use it here?

Our internal tsconfigs have some properties that shouldn't be shown to users:
```
    "typeRoots": [
      "../../node_modules/@types/"
    ]
```
```
  "exclude": [
    "node_modules/*",
    "**/*-aot.ts",
    "app-aot"
  ]
```

public/docs/_examples/cb-third-party-lib/hero-profile/publish.js
```diff
@@ -0,0 +1,59 @@
+// #docregion
+function publish() {
+  var inlineResources = require('./inline-resources');
+
+  // AoT compile
+  var spawnNgc = require( 'child_process' ).spawnSync;
+  var ngc = spawnNgc('./public/docs/_examples/node_modules/.bin/ngc', ['-p', './public/docs/_examples/cb-third-party-lib/hero-profile/tsconfig-aot.json']);
```
filipesilva: Although this is the correct path for our tooling, it shouldn't be the path shown to users.

public/docs/_examples/cb-third-party-lib/hero-profile/publish.js
```diff
@@ -0,0 +1,59 @@
...
+
+  // Copy to node_modules
+  var fs = require('fs');
+  var del = require('del');
+
+  var node_modules_root = './public/docs/_examples/node_modules/hero-profile/';
```
filipesilva: Same as above.

public/docs/_examples/cb-third-party-lib/hero-profile/publish.js
```diff
@@ -0,0 +1,59 @@
...
+
+  del.sync(node_modules_root, {force:true});
```
filipesilva: What is this step for? It's not explained in the prose.

public/docs/ts/latest/cookbook/third-party-lib.jade
```diff
@@ -0,0 +1,247 @@
...
+  JiT applications load the `umd` bundle using `SystemJS`. This requires a minor tweak to `systemjs.config.js` to register the `umd` bundle with `SystemJS`.
+
+  Simply add `'hero-profile': 'npm:hero-profile/bundles/hero-profile.umd.js'` to the map section of `systemjs.config.js`. 
+
+.l-main-section
+<a id="final-app"></a>
+:marked
+   ## Final Application 
+
+   If you have cloned the `Angular.io` repo, all the steps described in this cookbook can be executed by running the following command:
```
filipesilva: I think we need to give people instructions on how to build a sample lib outside of angular.io. Something like a `angular/quickstart-lib`, side by side with `angular/quickstart`.

This is something I can do later and should not block the release of this chapter.

thelgevold: I think that makes sense. Some of the complexity comes from having to simulate an npm dependency, and at the same time, make it work with the angular.io workflow. 

I think most of the value is in teaching what types of files are necessary to make the library AoT compatible.

thelgevold: I think it's also useful to describe the umd part of the library and its use.

filipesilva:
Superseded by #3411

PR #2760, opened 2016-11-05
docs(quickstart): app says "Hello Angular!" instead of "My first..."

wardbell:
A small step toward the QuickStart reboot


chalin:
Nice!

Maybe "Hello Again Angular!" rather than "Hello Angular 2!" as the greeting for the second app?

wardbell:
@chalin Sure.. Maybe later. I'm in the midst of a total reboot of the Quickstart.  First pass will be TS only after which I strongly suspect you'll want to follow suit in Dart. I need to focus on that right now.

PR #2771, opened 2016-11-08
chore(deploy): don't name project in firebase deploy

chalin:
Naming the project would sometimes cause gulp to report `An unexpected error has occurred` with exit code 2.


kwalrath:
LGTM. Thanks for the fix!

PR #2792, opened 2016-11-12
Fix display of heritage

### LINKED ISSUE #2791: API Reference - Missing Http Body documentation, no base classes or implemented methods
see https://github.com/angular/angular/issues/12784

Problem is: `abstract class Response extends abstract Body {...}`

The doc from `Body` is not propagated to `Response`.

@petebacondarwin [says it's because of the `abstract](https://github.com/angular/angular/issues/12784#issuecomment-260081779)

petebacondarwin:
This is a first step to fixing the problem that classes like `Response` in `@angular/http` does not show that it derives from `Body`.  A heritage string is already extracted by the `typescript` dgeni-package. So this PR includes this information in the relevant templates for classes and interface.

But this is probably not enough since the heritage processing does not:
a) show type parameters for generic super types (e.g. `RouterState` extends `Tree<ActivatedRoute>`).
b) these super types are not explicitly exported from any modules and so there are no API docs pages for them.

I'll put together a fix for a). But I am not sure what the best solution for b) is...

Closes #2791

### DIFF
--- a/tools/api-builder/angular.io-package/processors/matchUpDirectiveDecorators.js
+++ b/tools/api-builder/angular.io-package/processors/matchUpDirectiveDecorators.js
@@ -5,7 +5,7 @@ var _ = require('lodash');
  * @description
  *
  */
-module.exports = function matchUpDirectiveDecoratorsProcessor(aliasMap) {
+module.exports = function matchUpDirectiveDecoratorsProcessor() {
 
   return {
     $runAfter: ['ids-computed', 'paths-computed'],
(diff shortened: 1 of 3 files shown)


petebacondarwin:
That was quick @Foxandxss

vicb:
I'm impressed by the response time here (for the and fix and for the review) !
Great job guys, Thanks !

wardbell:
Why haven't we merged this yet? I don't know enough about it to do it (except to blindly click the button)

wardbell:
Merged on Pete's request to me in Slack

vicb:
I'm less impressed by the response time now :(

PR #2815, opened 2016-11-16
fix(live-example): correctly match example name

filipesilva:
Currently, when visiting a page with a relative link e.g. https://angular.io/docs/ts/latest/guide/testing.html#!#top, `NgIoUtil.getExampleName` fails to match the example name. 

This results on `live-example` directives on that page missing the example name in the link: instead of linking to https://angular.io/resources/live-examples/testing/ts/plnkr.html it will link to https://angular.io/resources/live-examples//ts/plnkr.html (notice the missing `testing` and the double slashes).

/cc @Foxandxss


chalin:
Sorry, my bad for introducing that problem when I added the optional `.html` suffix match. Thanks for the fix! LGTM

PR #2817, opened 2016-11-17
Update router.jade: Fix missing anchor id

kennyjacob:
The link *Importing from the router library* (https://angular.io/docs/ts/latest/guide/router.html#import) does not work.

I supposed that this was the corresponding section to be linked to.


kennyjacob:
I signed it!

PR #2821, opened 2016-11-17
docs(guide & api): feedback/issue links for Dart should refer to Dart repos

chalin:
- Also added links to chapters mentioned in table.
- Other minor copyedits.

### DIFF
--- a/public/docs/dart/latest/api/index.jade
+++ b/public/docs/dart/latest/api/index.jade
@@ -1,9 +1,9 @@
 include ../_util-fns
 
 :marked
-  > **Known issues:** The angular.io issue tracker contains [all known
-  issues][api-issues]; if you notice others, please [report
-  them][new-issue]. Thanks!
+  > **Known issues:** The Angular issue tracker contains
+  [all known issues][api-issues]; if you notice others, please 
+  [report them][new-issue]. Thanks!
 
   [new-issue]: !{_ngRepoURL}/issues/new?title=%5BAPI%20docs%5D
   [api-issues]: !{_ngRepoURL}/issues
--- a/public/docs/dart/latest/api/index.jade
+++ b/public/docs/dart/latest/api/index.jade
@@ -1,9 +1,12 @@
+include ../_util-fns
+
 :marked
   > **Known issues:** The angular.io issue tracker contains [all known
   issues][api-issues]; if you notice others, please [report
   them][new-issue]. Thanks!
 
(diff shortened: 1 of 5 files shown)


public/docs/ts/latest/guide/index.jade
```diff
@@ -80,37 +81,28 @@ table(width="100%")
...
-
-  Use the [angular.io Github repo](https://github.com/angular/angular.io) for **documentation** issues and pull requests.
+  - The [Cheat Sheet](cheatsheet.html) lists Angular syntax for common scenarios.
+  - The [Glossary](glossary.html) defines terms that Angular developers should know.
+  - The [API Reference](../api/) is the authority on every public-facing member of the Angular libraries.
 
+  ## We welcome feedback! 
 
-  Use the [Angular Github repo](https://github.com/angular/angular) to report issues with **Angular** itself.
+  - Use the [site Github repo](!{_ngDocRepoURL}) for **documentation** issues and pull requests.
```
kwalrath: Github -> GitHub

(Global. Not introduced in this PR, but while you're there...)

public/docs/ts/latest/guide/index.jade
```diff
@@ -22,7 +21,7 @@ table(width="100%")
   col(width="15%")
   col
   tr(style=top)
-    td <b>QuickStart</b>
+    td <b><a href="../quickstart">QuickStart</a></b>
```
kwalrath: It's a little weird to have links on QuickStart but not on Guide & Advanced... But it *is* definitely handy to have these links, especially for those times when the window is too narrow to display the sidenav.

This is a case where having everything in the guide/ directory is coming back to bite us.

Maybe we should try this out and see if anyone objects?

public/docs/dart/latest/api/index.jade
```diff
@@ -1,9 +1,12 @@
+include ../_util-fns
+
 :marked
   > **Known issues:** The angular.io issue tracker contains [all known
```
kwalrath: angular.io -> Angular

public/docs/ts/latest/guide/index.jade
chalin: Fixed

public/docs/ts/latest/guide/index.jade
kwalrath: No, I don't think a self-link is helpful. We could maybe briefly mention in the text that that's where they are.

chalin: (I was thinking about that, but wasn't sure how to adjust the text.)

PR #2822, opened 2016-11-17
chore: remove deprecated `clear=` and use clear layout helpers

ericjim:
This should ease the transition of Dart out of angular.io

cc @chalin @kwalrath


kwalrath:
Thanks!

Issue #1683, opened 2016-06-17
Document `ngProjectAs` in content projection (advanced)


vicb:
See https://github.com/angular/angular/blob/f39c9c9e75671a7e235734b6b8aef263f6dff254/modules/%40angular/compiler/test/template_parser_spec.ts#L1082

Please ping me if you have any question

ospatil:
As promised on angular/angular#12930, here is the pull request #2851. Let me know if it looks good.

PR #2851, opened 2016-11-21
docs - lifecycle-hook - add ngProjectAs documentation

ospatil:
Add documentation and example for selective content projection and _ngProjectAs_ attribute usage. For #1683.


wardbell:
It's an interesting example but I fail to see what this has to do with **Lifecycle Hooks** which is the focus of this guide. Unless you make a compelling argument very soon, I will be obliged to close.

ospatil:
I wasn't sure about the place either. The only page I could find in the entire documentation mentioning something about **Content Projection** is the **Lifecycle Hooks** page and hence I added it there as this feature is related to it.

I would be happy to submit a new pull request if you could point me to the right place in the documentation.

wardbell:
As you pointed out, there is no place where we talk about content projection. We haven't documented that yet and don't have it scheduled as yet. When we do write it, we will talk about this of course.

I'm sorry but I'm going to have to close this PR as it can't possibly go here.

PR #2856, opened 2016-11-22
Corrected RouterOutlet description

RodrigoQuesadaDev:
<router-outlet> does not wrap the views, the views are inserted next to its closing tag, not within its content.


RodrigoQuesadaDev:
I signed it!

PR #2872, opened 2016-11-24
Update styles.css

freezestudio:
Delete duplicate content

PR #2877, opened 2016-11-26
examples(styles): drop duplicate `items` classes but keep `li.selected`

chalin:
Supersedes #2872 but ensures that we don’t loose `.items li.selected`.


filipesilva:
Closed via https://github.com/angular/angular.io/pull/2877

PR #2879, opened 2016-11-27
docs(testing): simplify and accelerate path to Angular component testing

wardbell:
## First commit
Rob Wormald recognized that we had no plunker for a simple component test.
Added the banner-specs for that purpose and also a quickstart-specs in the setup folder
Adjusted prose in testing and setup-systemjs-anatomy to call these out
Supporting adjustments to plunker builder and gulpfile
Test plunkers now open on a spec instead of on index.html

## Second commit
Recognized that there was too much material getting in the way of learning to write Angular app tests.
Removed some stuff. Moved other stuff down.
Removed some of the "open on a spec" instructions because plunkers open in embedded-style

## Third commit
Remove `karma-htmlfile-reporter` from karma.conf and package.json because no longer used.
Adjust a few other karma.conf settings.


naomiblack:
Review on dev server @900f724 via https://ngdocsdev.firebaseapp.com/docs/ts/latest/guide/testing.html

PR #2885, opened 2016-11-28
docs(devguide): template tags are replaced by comments now

fmalcher:
Since 79399e1c511fb on `angular/angular` the template tag `<template></template>` is no longer replaced by `<script></script>` by default. Instead, a comment string is being inserted.

https://github.com/angular/angular/commit/79399e1c511fb9bf3f26259af43d6ad51e3fdef7#diff-3512bd2ab31e1763a0bc918f63a75835L121


wardbell:
Sorry for the long delay in response.  Of course you are correct. We are rewriting the section that includes this passage so we will incorporate your finding in a different way.  Thanks for pointing it out!

wardbell:
Closed by PR #3110

PR #2892, opened 2016-11-28
docs(webpack): fix for tsconfig.json provided in example

sergeome:
Pull request for this [issue](https://github.com/angular/angular.io/issues/2890).

### DIFF
--- a/public/docs/_examples/webpack/ts/tsconfig.1.json
+++ b/public/docs/_examples/webpack/ts/tsconfig.1.json
@@ -7,7 +7,6 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "lib": ["es2015", "dom"],
-    "removeComments": false,
     "noImplicitAny": true,
     "suppressImplicitAnyIndexErrors": true
   }
--- a/public/docs/_examples/webpack/ts/tsconfig.1.json
+++ b/public/docs/_examples/webpack/ts/tsconfig.1.json
@@ -6,8 +6,15 @@
     "sourceMap": true,
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
+    "lib": ["es2015", "dom"],
     "removeComments": false,
     "noImplicitAny": true,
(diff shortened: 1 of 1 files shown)


sergeome:
I signed it!

sergeome:
I signed it!

Foxandxss:
We don't need `exclude` without `files` and why do we need `typeRoots`? I am totally in for the `lib`, I wanted to add that with an update but forgot.

sergeome:
Hello @Foxandxss ,
1. I've added `exclude` in order to not process all files inside `node_modules`.
2. I thought that `typeRoots` actually can help solve the issue. But I've just verified and add `"lib": ["es2015", "dom"],` is enough.

Sorry. for this. 
As far as I understand we just need to add `"lib": ["es2015", "dom"],`. Can I modify it and submit updated version in this pull request?

Best regards,
Siarhei

Foxandxss:
Sure you can.

The node_modules exclusion is only needed if you specify the files by hand, since we do not, node_modules is excluded automagically.

sergeome:
@Foxandxss,
Thank you for the explanation. I didn't know about that.
I've updated my changes.

Best regards,
Siarhei

public/docs/_examples/webpack/ts/tsconfig.1.json
```diff
@@ -6,8 +6,9 @@
     "sourceMap": true,
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
+    "lib": ["es2015", "dom"],
     "removeComments": false,
```
wardbell: Would you please remove this `removeComments` setting too. That is the default setting and we don't want it here. I thought I'd removed it everywhere but I guess I missed this one.

Then I can merge your PR.

sergeome:
Hello @wardbell ,

I've just removed and pushed my changes.
Sorry for the delay with this.

Best regards,
Siarhei

Foxandxss:
Thank you.

PR #2907, opened 2016-11-29
docs(toh): fix code-example format and Dart code excerpt

chalin:
In the Dart toh-5: ensure `HeroesComponent` doc region `renaming` doesn't show a router.

### DIFF
--- a/public/docs/_examples/toh-5/dart/lib/heroes_component.dart
+++ b/public/docs/_examples/toh-5/dart/lib/heroes_component.dart
@@ -26,7 +26,11 @@ class HeroesComponent implements OnInit {
   Hero selectedHero;
 
   // #docregion renaming
-  HeroesComponent(this._heroService, this._router);
+  HeroesComponent(this._heroService,
+  // #enddocregion renaming
+                  this._router
+  // #docregion renaming
+                  );
   // #enddocregion renaming
 
   Future<Null> getHeroes() async {
--- a/public/docs/dart/latest/tutorial/toh-pt1.jade
+++ b/public/docs/dart/latest/tutorial/toh-pt1.jade
@@ -38,7 +38,7 @@ include ../_util-fns
   ## Keep the app compiling and running
   We want to start the Dart compiler, have it watch for changes, and start our server. We'll do this by typing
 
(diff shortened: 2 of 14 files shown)


public/docs/_examples/toh-5/dart/lib/heroes_component.dart
```diff
@@ -26,7 +26,11 @@ class HeroesComponent implements OnInit {
   Hero selectedHero;
 
   // #docregion renaming
-  HeroesComponent(this._heroService, this._router);
+  HeroesComponent(this._heroService,
```
kwalrath: It's OK to terminate an arg list with a ","? (Seems like it should be, but not sure whether all the SDKs the pubspec allows actually support this.)

kwalrath:
How does the second commit change the visible formatting? Also, does webdev handle this OK?

public/docs/_examples/toh-5/dart/lib/heroes_component.dart
wardbell: It's technically OK in TS although misleading and especially w/o the trailing `)`.  Not loving it. There should be another way to convey the message.

chalin: The trade-off is to enable plaster /* ... */ in the code excerpt, or to leave it as is. IMHO, even if a particular SDK doesn't support it, readers should understand that the code excerpt is suggestive, given that the heading is "_lib/heroes_component.dart (showing renamings only)_". Use of plaster just makes it too cluttered.

chalin:
Yes, webdev can handle all these changes (and uses all those formatings already).

public/docs/_examples/toh-5/dart/lib/heroes_component.dart
chalin: @wardbell - btw, this is only for the Dart code excerpt.

chalin: Btw, I was trying to avoid having a `heroes_component_1.dart` file (which would be another way of solving this).

PR #2915, opened 2016-11-30
chore(README): development setup - add deploy-install step

chalin:


README.md
```diff
@@ -29,15 +29,24 @@ This site relies heavily on node and npm.
 1. Make sure you are using at least node v.5+ and latest npm; 
 if not install [nvm](https://github.com/creationix/nvm) to get node going on your machine.
 
-1. Install npm `gulp` package *globally*: `npm install -g gulp`
+1. Install global npm packages by running `./scripts/before-install.sh`
```
filipesilva: Can you revert this line? Not all windows users use a console that can run bash scripts.

chalin: Sure, but wouldn't that apply to all scripts mentioned in the dev setup?

I'm just trying to keep this all DRY. We know that the scripts are kept up-to-date because they are run on Travis. If we duplicate the script commands here, then they run the risk of becoming out of step ... as they did this morning.

If a Windows user can't run a script I'd assume that they'd look inside and run the equivalent commands under Windows. Would that do?

filipesilva: There's only one other mention of the scripts in the readme (the one for resetting the environment). 

I guess it's fine, but will keep an eye out to see if new authors or users have having trouble.

chalin: sgtm, ty

PR #2918, opened 2016-12-01
docs(api/cheatsheet): show ng version + lang in page header

chalin:
Continuation of #2796, "duplicate header cleanup".

This PR also streamlines the cheatsheet page: the language and version are now shown in the header rather than in the page content.
> screen shot 2016-12-01 at 8 09 32 am

The Angular version is also shown on the API Reference page.

cc @kwalrath


chalin:
(Oups, concurrent update problem over the labels it would seem. Kathy approved so I've re-instated the proper labels.)

Issue #2930, opened 2016-12-05
complete migration of Dart to separate repo & reorg JS docs


chalin:
**Dart items**:
- [x] firebase redirects: all `/dart` --> Dart site, #2924
- [ ] **cache sync** for Dart docs before breaking single-source links
  - [x] architecture - #2990, https://github.com/dart-lang/site-webdev/pull/244
  - [x] pipes - #2937
  - [x] toh - #2936
  - [x] lifecycle-hooks - #2952
  - [ ] other pages, see #2077
- [x] finalize move of Dart Jade sources to site-webdev - #2956
- [x] move `ts/_cache` to site-webdev - #2974
- [x] move Dart `_examples` to site-webdev - #2976
- [ ] build scripts: clean out (remove Dart elements)
  - [x] minimal cleanup so that check-deploy doesn't require or deploy Dart docs - #2931
  - [ ] travis
  - [ ] gulpfile
    - [x] api + cheatsheet build elements - #2970
    - [ ] doc build elements
    - [ ] example build elements
  - [x] tools
    - `tools/dart-api-builder` - #2970
    - `tools/api-builder/dart-package` - #2970
- [ ] TS Jade sources (under `/docs`): clean out single-source elements
  - [ ] `_util-fns.jade`: mixins, functions, var defs.
  - [ ] TS Jade: clean out single-source elements:
    - [ ] Jade extends/blocks; mixins
    - [ ] inline text (via variables like `!{_array}`)
    - [ ] A1 directives like `ngio-ex`

    from:
      - [ ] quickstart
      - [ ] glossary
      - [ ] guide pages
      - [ ] tutorial pages
- [ ] A1 directive code & NgioUtils (under `/resources/js`): clean out Dart elements
- [x] migrate Dart issues to Dart repo

**JS items**
- [ ] firebase redirect to Why TS.
- [ ] Clean out JS elements from TS sources (there shouldn't be many of these but I do recall at least one situation where the Jade includes were in the opposite direction from what I expected: i.e., TS Jade file was including a JS Jade file).
- [ ] **remove language drop down**
  - [ ] perform any necessary styling / page layout adjustments
- [ ] clean out JS issues

tilwinjoy:
Does the "drop JS" in title mean that there won't be a JavaScript documentation at all..?

chalin:
@tilwinjoy - the title is indeed misleading, so I've adjusted it. What will be dropped will be the language submenu, not JS. Development of apps using JS will still be documented.

PR #2931, opened 2016-12-05
chore(gulp/travis): don't build Dart docs

chalin:
Contributes to #2930

cc @naomiblack @juleskremer


chalin:
In this PR:
- All gulp tasks will now build for `ts|js` by default (it used to be `ts|js|dart`).
- I've pulled out the build safeguards that previously would force a build of Dart API docs on a `check-deploy`.
- Deleted the build scripts used to install Dart repo and tooling.
- Travis: removed reference to the deleted build scripts (only affected `deploy-install.sh`).
- Added an `ignore` rule to `firebase.json` so that `docs/dart/**` is ignored (even if present).

Note that Dart doc pages will still be generated by Harp (but those take a trivial amount of time to generate).

Travis is green.

cc @filipesilva @wardbell

PR #2932, opened 2016-12-06
[WIP] feat(ru): style guide

valorkin:


wardbell:
Thanks for the effort!  Unfortunately, I must close because translation efforts need to be coordinated with Angular management. Please contact @juleskremer if you'd like to follow up.

valorkin:
@wardbell we are part of ng l10n, but there are no movement in RI direction

PR #2937, opened 2016-12-06
docs(pipes/dart): improve FetchJsonPipe discussion

chalin:
Followup to #2923

```
Suites passed:
  public/docs/_examples/pipes/dart
```

cc @wardbell - as usual, this is mainly updates to Dart files but with minor tweaks to the TS Jade. There was actually a `:marked` missing so some text wasn't showing up. I've fixed that.

### DIFF
--- a/public/docs/ts/_cache/guide/pipes.jade
+++ b/public/docs/ts/_cache/guide/pipes.jade
@@ -162,14 +162,16 @@ figure.image-display
 figure.image-display
     img(src='/resources/images/devguide/pipes/power-booster.png' alt="Power Booster")
 
+- var _decls = _docsFor == 'dart' ? 'pipes' : 'declarations';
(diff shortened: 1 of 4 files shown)


public/docs/ts/_cache/guide/pipes.jade
```diff
@@ -361,7 +369,7 @@ h3#async-pipe The impure #[i AsyncPipe]
   The component doesn't have to subscribe to the async data source,
   it doesn't extract the resolved values and expose them for binding,
   and the component doesn't have to unsubscribe when it is destroyed
-  (a potential source of memory leaks).
+  (a potent source of memory leaks).
```
kwalrath: This change (potential -> potent) seems questionable, but since it's not yours, Patrice...

@kapunahelewong maybe you can fix it?

PR #2940, opened 2016-12-07
[WIP] docs(why-ts): new "Why TypeScript?" page

wardbell:
**WIP** - still missing the "Angular Tools" section. Could merge now and add that later.

### DIFF
--- a/public/docs/ts/latest/why-ts.jade
+++ b/public/docs/ts/latest/why-ts.jade
@@ -1,0 +1,124 @@
+include _util-fns
+
+:marked
+  # Why Typescript?
+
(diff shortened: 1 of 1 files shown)


public/docs/ts/latest/why-ts.jade
```diff
@@ -0,0 +1,125 @@
+block includes
+  include _util-fns
```
chalin: You can drop the `block includes` and unindent the `include ...` line.

wardbell: Thanks.

PR #2941, opened 2016-12-07
docs(compiler): add doc + examples for i18n explicit ID option

maxime-allex:


vicb:
:+1:

wardbell:
I was touching up the examples and prose when I discovered that this depends upon an Angular 4.x feature. We do not have a way to handle those as I write. We'll figure it out.

Meanwhile, I've incorporated your head start in PR #3074 which closes this one.

Thanks!

PR #2956, opened 2016-12-09
docs: remove public/docs/dart

chalin:
cc @kwalrath


chalin:
Travis is fine with the changes, and I ensured that doing local builds still work w/o dart docs.

PR #2960, opened 2016-12-13
chore: update to in-memory-web-api v.0.2.0

wardbell:
So that `Observable<Response>` returned by `Http` calls involving the in-mem-web-api are "cold".
That means they don't occur until someone subscribes.

See [in-mem repo's change log for 0.2.0](https://github.com/angular/in-memory-web-api/blob/master/CHANGELOG.md#020-2016-12-11).


wardbell:
Oops. Shouldn't have Kapunahele's Reactive Forms changes

PR #2965, opened 2016-12-13
docs(toh-5): fix HeroesDetailComponent typo

portify:
Looks like this was supposed to refer to the `HeroDetailComponent` instead (there is no `HeroesDetailComponent`)


portify:
I signed it!

wardbell:
Good catch!

PR #2976, opened 2016-12-14
chore: delete Dart _examples

chalin:


kwalrath:
This reminds me... We need to update https://github.com/angular/dart-doc-syncer (and the README.md files that it produces).

PR #3004, opened 2016-12-19
fix typos

shairez:


wardbell:
This doc has been rewritten so your fixes, worthy as they are, are gone.

PR #3034, opened 2016-12-24
docs(router): warn about routing modules order

jiayihu:
If `AppRoutingModule` is imported in `imports` before feature modules, their routes will be ignored. I think this should be added in the docs for clarity.


jiayihu:
I signed it!

wardbell:
The PR#3060 incorporates this advice and much else. Thanks, but closing as a duplicate.

p.s.:  Perhaps you were the reason we added that advice. Don't remember. Thanks anyway.

PR #3040, opened 2016-12-27
docs(forms): ngSubmit isn't an Angular directive.

lightningtgc:
ngSubmit just an output property. 
See:
https://github.com/angular/angular/blob/master/modules/%40angular/forms/src/directives/ng_form.ts#L63-L64


lightningtgc:
@Foxandxss Yep,it is  directive's  event (FormGroupDirective and NgForm),maybe we can change the wording.

wardbell:
Thanks. I felt the need to further refine your observation in PR #3056. Closing this PR in favor of that one. But do much appreciate your contribution.

PR #3043, opened 2016-12-28
feat(prod-deploy): add prod deploy chapter

filipesilva:
This PR adds the Production Deployment Cookbook.

/cc @wardbell @Foxandxss @DeborahK


public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+
+.l-main-section
+:marked
+  ## Simplest deploy possible
+
+  The simplest possible way to deploy your app is to use your development environment.
+  It is already working locally after all, so it should work live.
+
+  1. Copy over your local project folder to your server.
+  2. In your server, edit `index.html` to have the right `<base href="/">` tag. If you are
```
DeborahK: Consider using "correct" instead of "right" here since right could mean right/left.

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  payload size, plus Angular and server configuration.
+
+.l-main-section
+:marked
+  ## Simplest deploy possible
+
+  The simplest possible way to deploy your app is to use your development environment.
+  It is already working locally after all, so it should work live.
+
+  1. Copy over your local project folder to your server.
```
DeborahK: In my daughter's case, she had a limited amount of allocated space on the server and could not copy all of the node_module folder. Consider mentioning using `npm install --production` here?

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+.l-main-section
+:marked
+  ## Minimizing your payload
+
+  One of the easiest optimizations you can do is to reduce the code users have to download.
+  
+  There are several techniques for achieving a smaller payload for your app:
+
+  - Bundling: concatenating several modules into a single file (bundle). 
+  - Inlining: add html and css as strings inside components.
```
DeborahK: Should be "adding" to match the other "-ing" words used here.

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  - Minification: removing all unnecessary whitespace and optional tokens.
+  - Uglification: rewriting code to use shorter variable/function names.
+  - Dead Code Elimination: removing code/variables that will never used.
```
DeborahK: Missing word here "that will never used". Consider changing to: "are not used" to be more active.

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  - Tree shaking: removing unused module exports.
+  - Ahead-of-Time (AoT) Compilation: pre-compiling Angular templates.
+
```
DeborahK: Great list!

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  Each of these do different things overall, but they all work together and their effects
+  compound on each other.
+
+  ### Less files
+  Bundling and inlining reduce your app's load time smaller by serving less files and thus
```
DeborahK: There is something wrong with this sentence. An extra word ("smaller") maybe? And the "less" here should be "fewer".

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  - Minification: removing all unnecessary whitespace and optional tokens.
+  - Uglification: rewriting code to use shorter variable/function names.
+  - Dead Code Elimination: removing code/variables that will never used.
+  - Tree shaking: removing unused module exports.
+  - Ahead-of-Time (AoT) Compilation: pre-compiling Angular templates.
+
+  Each of these do different things overall, but they all work together and their effects
+  compound on each other.
+
+  ### Less files
```
DeborahK: Consider changing "Less" to "Fewer".

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  Bundling and inlining reduce your app's load time smaller by serving less files and thus
+  less round trips. 
+  For small files, the overwhelming majority of time it takes to get it is not spent downloading,
+  but rather communicating with the server and coordinating the transfer. 
+  You can read more about resource timings in the 
+  [Chrome DevTools Network Performance page](https://developers.google.com/web/tools/chrome-devtools/network-performance/understanding-resource-timing)   
+
+  ### Smaller files
+  Minification, Uglification and Dead Code Elimination all contribute to reduce the total size
+  of your code to the smallest possible amount of bytes, which will result in a smaller
```
DeborahK: Consider changing "will result" to "results" to make it more active.

public/docs/ts/latest/cookbook/production-deploy.jade
DeborahK: In -> On?

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  transfer time.
+
+  Tree shaking also reduces the total size but does so by removing whole exports from modules.
+  It was popularized by [Rollup](http://rollupjs.org/) and you can read more about it in 
+  [this post](https://medium.com/@Rich_Harris/tree-shaking-versus-dead-code-elimination-d3765df85c80#.15ih9cyvl).
+
+  ### Template compilation
+  AoT compilation reduces render time by pre-compiling Angular templates that otherwise would be
+  compiled at runtime. 
+  It also reduces payload size since the Angular compiler itself is no longer needed, and nor are
```
DeborahK: Consider clarifying "is no longer needed". Maybe something like: "does not need to be downloaded to the client"  (or user)

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  directives that are never used in templates. This allows Tree-Shaking to further remove exports.
+
+  You can read more about AoT Compilation in our [dedicated cookbook chapter](aot-compiler.html),
+  where you'll also find instructions on how to use Rollup to do all the optimizations shown here.
```
DeborahK: "to do all" -> "to perform all"?

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  Another great choice is [Webpack 2](https://webpack.js.org/) together with the 
+  [Angular Ahead-of-Time Webpack Plugin](https://github.com/angular/angular-cli/tree/master/packages/%40ngtools/webpack).
```
DeborahK: Consider adding another sentence here on why this is a great choice. "Webpack makes it easier ..." or "Webpack combines ..." or whatever.

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  
+  You can use any build system you like however, the only important thing is to optimize your build.
```
DeborahK: Consider reducing the number of words here "you like however, the only important thing..." to "you like; the important thing...".

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+.l-main-section
+:marked
+  ## Angular configuration
+
+  ### The `base` tag
+
+  There is an important configuration item in the Angular Router that needs to be adjusted for 
+  different environments: 
+  [the `base` tags `href` property](https://angular.io/docs/ts/latest/guide/router.html#!#base-href).
+  This property tells the router how what the app root URL is so that the router can match routes.
```
DeborahK: Replace "how what" with "what".

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+
+  On your development server, your app is most likely served at the root: `http://localhost:3000/`.
+  Here you use `<base href="/">`, since `/` is the root of your app.
+  
+  But depending on the server setup that you use, it can also be served in a subfolder like
```
DeborahK: "it can also be served in" -> "the app can be served from"

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  `http://localhost:3000/myapp/` or `http://www.mysite.com/myapp/`, where `/myapp/` is the subfolder.
+  In this case you should use `<base href="/myapp/">`, since `/myapp/` is the root of your app and 
+  not `/`.
+
+  When your `base` tag is misconfigured your browser console will show several console, from
```
DeborahK: "will show several console" -> "will show several errors in the console"?

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  missing files on wrong paths to router errors about missing routes.
```
DeborahK: Is there an easy way to set this up so that the same code works in dev and deployment? Like adding an environment variable? Or somehow changing this in a build step?

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+
+  ### enableProdMode
+
+  You can also configure your Angular app to start on 
+  [production mode](https://angular.io/docs/ts/latest/api/core/index/enableProdMode-function.html), 
+  which will make it faster by disabling development specific checks.
+
+  Angular apps default do development move, as you can see by the following message on the browser
```
DeborahK: "do" -> "to"

DeborahK: "move" -> "mode"

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+code-example(format="nocode").
+  Angular 2 is running in the development mode. Call enableProdMode() to enable the production mode.
+
+:marked
+  To enable production mode, add this code to the start of your `main.ts`:
+
+code-example().
+  import { enableProdMode } from '@angular/core';
+  enableProdMode();
+
```
DeborahK: Is there an easy way to toggle this for dev vs production?

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  are needed.
+
+  A common mistake to make while lazy loading a module is to *also* have that module imported via
+  a ES6 import. 
+  If you do that, the module will never really be lazy loaded since the import statement will cause
+  it to be imported directly instead of being imported via the Router. 
+
+  It's important to note that your bundling configuration must take lazy loading into consideration.
+  Since there is no ES6 import for that module, normal bundlers don't know that they should also
+  separately bundle to module listed in the router config.
```
DeborahK: Missing a word here? "separately bundle to module listed"

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  You have to manually create additional bundles for each lazy loaded route.
```
DeborahK: Would this be more correct to say "lazy loaded Angular module" instead of "route" here?

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+
+  If you are using Webpack, the 
+  [Angular Ahead-of-Time Webpack Plugin](https://github.com/angular/angular-cli/tree/master/packages/%40ngtools/webpack)
+  will automatically recognize lazy routes.
```
DeborahK: "lazy" to "lazy loaded" for clarity?

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+
+
+.l-main-section
+:marked
+  ## Server configuration
+
+  Angular apps are 
+  [Single Page Applications](https://en.wikipedia.org/wiki/Single-page_application) (SPAs),
+  which makes them the perfect candidates to be served by simple static HTML server.
```
DeborahK: "by a simple" (missing "a")

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  No preprocessors required!
+
+  There is only one configuration item required of a server hosting an Angular app:
+  it needs to be able to fallback to `index.html` when being asked for a file the server
```
DeborahK: "it needs to be able to fallback" -> "it needs to fallback"

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  So, while the server has the responsability of serving the files, it is always the Angular app
+  that will determine which routes are valid and construct the page accordingly.
+  
+  Imagine you have the `admin/` route for your `www.my-awesome-app.com` Angular app. 
+  When you visit `www.my-awesome-app.com` most servers will respond with
+  the `index.html` located at `www.my-awesome-app.com/index.html`. 
+  Then you click the link for `admin/` and the your address bar will change 
+  to `www.my-awesome-app.com/admin/` and you're taken to the admin page.
+
+  But if you don't have fallback to `index.html` configured and refresh the page, you will
```
DeborahK: To make this more clear, consider rearranging this sentence. "If you then refresh the page, you will get a 404 - Not Found page unless you have the fallback to index.html configured."

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  Which doesn't make a lot of sense since you just saw that the admin page was there.
+
+  When you first clicked the link to `admin/` there was no server request for 
+  `www.my-awesome-app.com/admin/` because Angular manipulated the browser URL directly.
+
+  But when you refresh the `www.my-awesome-app.com/admin/` url, you are not asking *Angular*
+  for the admin route but rather *the server*.
+  And the server does not have a `www.my-awesome-app.com/admin/index.html` to serve, so it instead
+  shows you a 404.
+
```
DeborahK: Great to see this explained!

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  That is why the server needs to be configured to fallback to `index.html`.
+  When a user asks the server for any route (that isn't a file), you want the server to *always*
+  return default page that boots up Angular.
```
DeborahK: "return the" (missing "the")

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  After Angular is loaded and bootstrapped the router will look at the address bar and take
```
DeborahK: "will look" -> "looks" and "take" -> "takes"

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  the user where he needs to go.
```
DeborahK: Should we be avoiding he/she? Either "he" -> "they" or change to "takes the user to the appropriate route".

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+
+  ### Fallback configuration examples
+
+  Different servers are configured in different ways so there is no single configuration that
+  works everywhere.
+
+  Following are configurations for some of the most popular servers. 
+  The list is by no means exhausting, but should provide you with a good starting point.
```
DeborahK: exhausting -> exhaustive

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+  A common issue developers come across when working on SPAs is 
+  [Cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing)
+  errors.
+  These are not due to the SPA application itself but rather due to the configuration of APIs with
+  which the SPA communicates.
+
+  These APIs need to send the `Access-Control-Allow-Origin: *` header on their HTTP responses,
+  otherwise the SPA will not be able to call them.
+  You can read more about how to enable CORS for specific servers in 
+  [enable-cors.org](http://enable-cors.org/server.html)
```
DeborahK: Really good info here! Great job pulling this together!

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+    * [Smaller files](#smaller-files)
+    * [Template compilation](#template-compilation)
+  * [Angular configuration](#angular-configuration)
+    * [The `base` tag](#the-base-tag)
+    * [enableProdMode](#enableprodmode)
+    * [Lazy loading](#lazy-loading)
+  * [Server configuration](#server-configuration)
+    * [Why fallback to `index.html`?](#why-fallback-to-index-html-)  
+    * [Fallback configuration examples](#fallback-configuration-examples)  
+
```
filipesilva: Thanks!

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,316 @@
...
+a#overview
+.l-main-section
+:marked
+  ## Overview
+
+  After creating your brand new Angular app you'll want to put it online for the world to see.
+
+  Your development setup is optimized for build speed and rapid iteration, but when
+  deployment to production you'll want to optimize for loading speed and payload size.
```
filipesilva: Right you are, fixing.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Added note about production installs.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done

filipesilva: Done

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: That one was added later and I forgot to do just that, good catch!

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Right you are, I left a word in while rewriting that paragraph. Fixed.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Clarified.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Perfect place for a semicolon!

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Fixed

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: I'm not aware of a super easy way to do it right now, but am planning on incorporating a production deploy script in the quickstart that would do all this.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Fixed

filipesilva: Fixed

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: I don't think so. We had to add in environment specific files in the CLI to have this functionality. I think there's an AoT option in the works that forces it when compiling in AoT, but for non-aot it has to be manual.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Fixed.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Yeap, fixed.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Fixed

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Fixed

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Reworded

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Fixed

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Fixed

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Fixed

filipesilva:
Thanks for the review @DeborahK, I've incorporated all your feedback now. 

I think I touched everything you once listed as questions your users had. If there's something missing let me know.

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+  * [Server configuration](#server-configuration)
+    * [Why fallback to `index.html`?](#why-fallback-to-index-html-)  
+    * [Fallback configuration examples](#fallback-configuration-examples)  
+
+a#overview
+.l-main-section
+:marked
+  ## Overview
+
+  After creating your brand new Angular app you'll want to put it online for the world to see.
```
johnpapa: Suggest removing the word *brand* and just saying *After creating your new Angular app*.

johnpapa: I may not want it to be online for the world to see. I may be building a corporate internal app. Recommend just saying *deploy it*.

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+
+  Your development setup is optimized for build speed and rapid iteration, but when
+  deploying to production you'll want to optimize for loading speed and payload size.
+
+  In this chapter you'll see how to deploy your app right now, techniques to reduce your
```
johnpapa: remove *right now*

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+
+  In this chapter you'll see how to deploy your app right now, techniques to reduce your 
+  payload size, plus Angular and server configuration.
+
+.l-main-section
+:marked
+  ## Simplest deploy possible
+
+  The simplest possible way to deploy your app is to use your development environment.
+  It is already working locally after all, so it should work live.
```
johnpapa: Suggest *It is working locally, so it will work when published to a web server, too*

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+
+  1. Copy over your local project folder to your server.
+  2. On your server, edit `index.html` to have the correct `<base href="/">` tag. If you are
+  using a subfolder (e.g. `www.mysite.com/myapp/`) it should be `/myapp/` otherwise leave it as `/`.
+  [More on this later](#angular-configuration).
+  3. Configure your server to redirect requests for missing files to `index.html` instead.
+  [More on this later](#server-configuration).
+
+  And this is all you need to publish your app!
```
johnpapa: Remove *And*

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+
+  It's not a very good production deploy though.
```
johnpapa: Instead of negative prose, I suggest ... *You've deployed your app, but we can do better for production to optimize the app! Let's learn how.*

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+  [More on this later](#angular-configuration).
+  3. Configure your server to redirect requests for missing files to `index.html` instead.
+  [More on this later](#server-configuration).
+
+  And this is all you need to publish your app!
+
+  It's not a very good production deploy though. 
+
+  For one, it has a lot of files that aren't even used when serving your site to users since it has
+  all the development only packages.
```
johnpapa: Maybe add a link to the `package.json` to show the dev packages list

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+
+  For one, it has a lot of files that aren't even used when serving your site to users since it has
+  all the development only packages. 
+  You could cut down the number of files by doing a production install (`npm install --production`),
+  but only *after* compiling your Typescript files.
+
+  It also has a lot of small files. Each of those is going to take a whole round trip to get
+  delivered to the browser, so it's not going to be that fast to load.
+
+  There's more you can do to make our site fast for users.
```
johnpapa: Suggest adding ... *Plus, there's more ...*

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+ 
+.l-main-section
+:marked
+  ## Minimizing your payload
+
+  One of the easiest optimizations you can do is to reduce the code users have to download.
+  
+  There are several techniques for achieving a smaller payload for your app:
```
johnpapa: Great list! We all know what these do, but I think we should add a sentence here that explains succinctly the net effect of these tasks. Perhaps add ... `which will reduce the load time and improve the user experience.`

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+  - Inlining: adding html and css as strings inside components.
+  - Minification: removing all unnecessary whitespace and optional tokens.
+  - Uglification: rewriting code to use shorter variable/function names.
+  - Dead Code Elimination: removing code/variables that are not used.
+  - Tree shaking: removing unused module exports.
+  - Ahead-of-Time (AoT) Compilation: pre-compiling Angular templates.
+
+  Each of these do different things overall, but they all work together and their effects
+  compound on each other.
+
```
johnpapa: Suggest adding `Let's take a moment to better understand what each of these mean, and the value that each of these bring.`

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+  ### Template compilation
+  AoT compilation reduces render time by pre-compiling Angular templates that otherwise would be
+  compiled at runtime. 
+  It also reduces payload size since the compiled app does not depend on the Angular compiler,
+  nor on declared directives that are never used in templates. 
+  This allows Tree-Shaking to further remove exports.
+
+  You can read more about AoT Compilation in our [dedicated cookbook chapter](aot-compiler.html),
+  where you'll also find instructions on how to use Rollup to perform all the optimizations shown here.
```
johnpapa: I suggest removing the word `Rollup` from here. Why? Because we may use something else later and it would be easier to edit just that doc later and not this one too.

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+code-example().
+  import { enableProdMode } from '@angular/core';
+  enableProdMode();
+
+:marked
+  ### Lazy loading
+
+  The Angular Router allows you to configure NgModules as being 
+  [Lazy Loaded](https://angular.io/docs/ts/latest/guide/router.html#!#asynchronous-routing),
+  which means that particular NgModule (and all it's code) is not loaded on the very first load.
```
johnpapa: `it's` when possessive it `its`

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+  This allows your app to have a smaller initial load time by not loading modules until they
+  are needed.
+
+  A common mistake to make while lazy loading a module is to *also* have that module imported via
+  a ES6 import.
```
johnpapa: `a` --> `an`

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+  This allows your app to have a smaller initial load time by not loading modules until they
+  are needed.
+
+  A common mistake to make while lazy loading a module is to *also* have that module imported via
+  a ES6 import. 
+  If you do that, the module will never really be lazy loaded since the import statement will cause
+  it to be imported directly instead of being imported via the Router. 
+
+  It's important to note that your bundling configuration must take lazy loading into consideration.
+  Since there is no ES6 import for that module, bundlers don't know that they should also
```
johnpapa: Perhaps we should just use `ES import` instead of `ES6` import. That safeguards us for future versions ... it is also what I do for all of my material :)

public/docs/ts/latest/cookbook/production-deploy.jade
```diff
@@ -0,0 +1,327 @@
...
+
+
+.l-main-section
+:marked
+  ## Server configuration
+
+  Angular apps are 
+  [Single Page Applications](https://en.wikipedia.org/wiki/Single-page_application) (SPAs),
+  which makes them the perfect candidates to be served by a simple static HTML server.
+  No preprocessors required!
```
johnpapa: I know what this means, but will the reader know what a `preprocessor` is? either explain or remove.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: That will vary wildly from project to project, and will be fairly long as well here. I will add some examples though.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Done, across the doc.

public/docs/ts/latest/cookbook/production-deploy.jade
filipesilva: Reworded to
>   There is no need to have a server-side language composing the pages because
  Angular will do it on the client-side.

WDYT?

public/docs/ts/latest/cookbook/production-deploy.jade
johnpapa: ok

public/docs/ts/latest/cookbook/production-deploy.jade
johnpapa: LGTM

wardbell:
I very much like the tone, direction, and **brevity** of this cookbook!

I'll put copy edits elsewhere. Here are some stream-of-consciousness questions.

>I ask a lot of "why" questions below. I'm not suggesting that we burden this page with a lengthy discourse about the "why" of everything. But we should know "why" ... and may want to condense the "why" into a sentence here and there.

* I think we need a different title as "Production deploy" is a slang noun. Suggestions welcome. So far the best I have is "Deploy to production".

* I'm not sure that this is a cookbook. There is a lot of high level advice but relatively little in the way of instructions for what to do. The notable exception is that highly detailed "Server configuration" section.

  Can we be more specific?

* We should separate "optimization" from "deployment". You do that yourself by offering a path that pushes the dev environment to the web: "Simplest deploy possible" ... love that title.

* I think we should be more clear about what we're optimizing for.  Is it faster launch time? That seems to be what you're after. So what is the metric there? I've seen discussions that call out the time from click-to-app-usability. Maybe we should reference those?

* Let's make a strong case for a published dev build; as @DeborahK as observed, that's often a crucial step in winning (and retaining) managment/stakeholder support.

* In several places you say "optimize for loading speed and payload size".  Is payload size a goal on its own? You suggest it makes a difference in transfer time and startup time. In what environments is that true? I haven't seen numbers that support this claim for desktop or mobile apps over 4G networks. I'm a skeptic.

* In any case, I think of a smaller payload as _just one factor_ that contributes to faster loading.   It's significance in that respect depends a great deal on the target environment, yes? In my (limited) experience, the far bigger factor, both in desktop and over mobile, is the number of trips to the server.  

* In several places, you **conflate reduced payload size and number of files**. They are separate issues and may be seen as inversely related: bundling, for example, reduces the number of files while increasing the payload size for each request. 

  Which is more important? When? 

  In my (limited) experience, the number of files matters a lot whereas (total net) payload size makes comparatively little difference in app launch and startup time ... at least for enterprise apps. Is that your experience?

  There may be good reasons to **do both**. My point is that **a few optimizations** (bundling and inlining?) **may be more significant** than the others combined in improving perceived app launch time. Bundling may be much easier to achieve than some or all of the extraordinary efforts to reduce total payload size.

* I confess to wondering about the nearly maniacal effort of the Angular team to shrink the payload. Where are the metrics for the benefit of that? @johnpapa may be able to weigh in from his corporate experience.

* You wrote:

  >For one, it has a lot of files that aren't even used when serving your site to users since it has all the development only packages. 

  Why should I care? What is the cost of having extra files on the server now that storage is almost free? 

* I thought _Tree Shaking_ was a technique for _Dead Code Elimination_. You list these as separate items. Why?

* Why is the following statement under the "Template compilation" heading?

  >Another great choice is Webpack 2 together with the Angular Ahead-of-Time Webpack Plugin.

  Perhaps you should say what it is.

* Many people wonder if using SystemJS is "wrong" or means that you can't optimize for production. I don't think that is true, even if we are leaning to Webpack ourselves. What can we say to help people not worry about their use of SystemJS? Or should they worry?


*  What would be the perf benefit of switching to umd scripts for Angular and other 3rd party libraries. Aren't these a good choice if you can't go the webpack way? That's what we do for plunkers. Or is this thought just a distraction?

* The `base tag` section raises a concern but doesn't say how to ameliorate it.  There is a line of `index.html` code that works _everywhere_  I think: in local development and on any server. It's the one we use in generated plunkers:
  ```
  <script>document.write('<base href="' + document.location + '" />');</script>
  ```
  If that's true, should we recommend it?

*  Re: `enableProdMode` - why do you recommend setting that in `main.ts`. That defeats its purpose during development. Shouldn't it be inside a conditional that enables production mode in production builds?

* The `Lazy Loading` section is a great start. I don't think we're as clear as we could/should be on how to recognize and overcome the mistake:

  >A common mistake to make while lazy loading a module is to also have that module imported via an ES import. If you do that, the module will never really be lazy loaded since the import statement will cause it to be imported directly instead of being imported via the Router.

  That's something we can look at in copy editing.

* You made several references to the [Angular Ahead-of-Time Webpack Plugin](https://github.com/angular/angular-cli/tree/master/packages/%40ngtools/webpack).  You don't say how I can use that independent of the CLI. Can I?

* The _Server Configuration_ section is excellent and unusually detailed compared to anything else on this page.  I like that. But it feels a bit strange to be reading high-level advice and suddenly diving into details.

* Thanks for the CORS section!

filipesilva:
@wardbell I edited your comment with my replies so as to not bloat this thread unnecessarily. My replies look like this:
> FS: reply here

johnpapa:
I can't edit the comment so I will respond here ...

> Ward: In several places you say "optimize for loading speed and payload size". Is payload size a goal on its own? You suggest it makes a difference in transfer time and startup time. In what environments is that true? I haven't seen numbers that support this claim for desktop or mobile apps over 4G networks. I'm a skeptic.

I've seen many examples of this making a huge difference even on 4G. A large portion of our users use desktop/laptop on WiFi and the difference in 2 seconds to usability vs 3 has quantifiable effects on abandonment and revenue (per some companies I have seen stats from). We don't need to state this. 2 seconds is the bar that most folks I have spoken to have to deal with for public sites.

Even with internal apps, there is a focus (though more relaxed) on the speed. I'm sure many companies dont care ... but just as many do.

DeborahK:
I would *really* like to see this come out sooner rather than later for several reasons:
1) There is really good content here. We can edit/enhance over time.
2) Just yesterday I received another "now how do I deploy" question. It would be great to have something to point them to.
3) Talked to Joe and he did a chapter on deployment in the Fundamentals course. I would expect this to be coming out any day now. It would be great if we had something in the docs prior to then.

johnpapa:
I agree its not a cookbook .. but the content is very valuable. I think we are have a very confusing organization of docs anyway ... we should focus on good content and simply create a master list of the topics. 

My short term point is that we should publish this sooner rather than later wherever it goes. I highly doubt anyone knows the diff between cookbook and advanced and regular docs :)

wardbell:
Well _I know the difference_ between a cookbook and guide :-)

For sure a cookbook contains detailed instructions you can follow. For sure it contains sample code. This PR offers neither.

>I'm not judging. This is all I had reason to expect from the author, given the constraints I set for him. But I am stating facts.

Imagine a cookbook recipe called "Boil an egg" that talked about how chickens come from eggs (it might or might not mention that eggs come from chickens), and the relative merits of cooking in hot water versus steam, and how most people used gas stoves ... but never told you what to do.  That is a _guide_ at best, not a cookbook recipe at all.

At some point we have to go beyond musing about the factors and give people some real instructions. That isn't happening here to my satisfaction. There is one place that comes close to giving actual instruction for deploying to a server - the "Simplest possible deploy" - and even that omits crucial details or an example for even one server.

There is no accompanying code, nothing you can run, nothing you can see that looks like a production app.

Having said that, so pressing is the need for _something_ that I agree to completing the review and copy-editing _this week_ ... as a preliminary guide.  **Hold me to it!**

public/docs/ts/latest/cookbook/production-deploy.jade
wardbell: Yes there is ... and I've included it in my commit:
```
// Enable production mode unless running locally
if (!/localhost/.test(document.location.host)) {
  enableProdMode();
}
```

filipesilva: That's a pretty good way!

public/docs/ts/latest/guide/production-deploy.jade
```diff
@@ -0,0 +1,415 @@
...
+  Development files are full of comments and whitespace for easy reading and debugging.
+  The browser downloads entire libraries, instead of just the parts the app needs.
+  The volume of code passed from server to client (the "payload") 
+  can be significantly larger than is strictly necessary to execute the application.
+
+  The many requests and large payloads mean
+  your app takes longer to launch than it would if you optimized it.
+  Several seconds may pass (or worse) before the user can see or do anything userful.
+  
+  Does it matter? That depends upon business and technical factors you must evaluate for yourself.
```
filipesilva: I like this bit. A lot of people think that it's all or nothing but it's actually depends on knowing what you need.

public/docs/ts/latest/guide/production-deploy.jade
```diff
@@ -0,0 +1,415 @@
...
+  If it _does_ matter, there are tools and techniques to reduce the number of requests and the size of responses.
+
+  - Ahead-of-Time (AOT) Compilation: pre-compiles Angular component templates.
+  - Bundling: concatenates modules into a single file (bundle). 
+  - Inlining: pulls template html and css into the components.
+  - Minification: removes excess whitespace, comments, and optional tokens.
+  - Uglification: rewrites code to use short, cryptic variable and function names.
+  - Dead code elimination: removes unused code.
+  - Tree shaking: removes unreferenced modules.
+  - Performance measurement tools.
```
filipesilva: Not sure this should be a bullet point. Perhaps a paragraph like this:

```
You cannot optimise what you cannot measure. Be sure to become familiar with the performance measuring tools present in any modern-day browser to objectively assess the difference between different approaches.
```

public/docs/ts/latest/guide/production-deploy.jade
```diff
@@ -0,0 +1,480 @@
...
+
+a#overview
+.l-main-section
+:marked
+  ## Overview
+
+  This guide describes techniques for preparing and deploying an Angular application to a server running remotely.
+  The techniques progress from _easy but suboptimal_ to _more optimal and more involved_.
+
+  * The [simple way](#dev-deploy "Simplest deployment possible") is to copy the development environment to the server.
```
robwormald: This is probably not something we want to recommend, ever?

public/docs/ts/latest/guide/production-deploy.jade
```diff
@@ -0,0 +1,480 @@
...
+  `<base href="/myapp/">`.
+  Otherwise, leave it alone.
+  [More on this below](#base-tag).
+
+  1. Configure the server to redirect requests for missing files to `index.html`.
+  [More on this below](#fallback).
+
+  1. Enable production mode as [described below](#enableprodmode) (optional).
+
+  That's all it takes to publish your app!
```
robwormald: See above. Concerned people will stop here.

public/docs/ts/latest/guide/production-deploy.jade
```diff
@@ -0,0 +1,480 @@
...
+
+a#node-modules
+:marked
+  ### Load npm packages from the web (SystemJS)
+
+  The `node_modules` folder of _npm packages_ is huge and slow to upload to the server.
```
robwormald: Lets avoid pejorative words like "huge" and "slow: :)

public/docs/ts/latest/guide/production-deploy.jade
```diff
@@ -0,0 +1,480 @@
...
+  It's typically 20,500+ files and 180+ MB.
+  The application itself requires a tiny fraction of that to run.
+
+  Load the _npm packages_ from the web instead.
+
+  (1) Make a copy of `index.html` for deployment and replace all `node_module` scripts
```
robwormald: I think here, we want to use SystemJS builder instead. UMD bundles are a less optimal option and straight out won't work with AoT

public/docs/ts/latest/guide/production-deploy.jade
```diff
@@ -0,0 +1,480 @@
...
+  with versions that load from the web. It might look like this.
+
++makeExample('index.html', 'node-module-scripts', '')(format=".")
+:marked
+  (2) Replace the `systemjs.config.js` script with a script that 
+  loads `systemjs.config.server.js`.
++makeExample('index.html', 'systemjs-config', '')(format=".")
+:marked
+  (3) Add `systemjs.config.server.js` (shown in the code sample below) to the root folder.
+  This alternative version configures _SystemJS_ to load _UMD_ versions of Angular
```
robwormald: ditto. UMD !== deployment.

public/docs/ts/latest/guide/production-deploy.jade
```diff
@@ -0,0 +1,480 @@
...
+  Development files are full of comments and whitespace for easy reading and debugging.
+  The browser downloads entire libraries, instead of just the parts the app needs.
+  The volume of code passed from server to client (the "payload") 
+  can be significantly larger than is strictly necessary to execute the application.
+
+  The many requests and large payloads mean
+  the app takes longer to launch than it would if you optimized it.
+  Several seconds may pass (or worse) before the user can see or do anything userful.
+  
+  Does it matter? That depends upon business and technical factors you must evaluate for yourself.
```
robwormald: I have a hard time with this. I can't see a scenario where it's beneficial to deploy to production without these things. Why even ask the question?

public/docs/ts/latest/guide/production-deploy.jade
```diff
@@ -0,0 +1,480 @@
...
+  
+  If it _does_ matter, there are tools and techniques to reduce the number of requests and the size of responses.
+
+  - Ahead-of-Time (AOT) Compilation: pre-compiles Angular component templates.
+  - Bundling: concatenates modules into a single file (bundle). 
+  - Inlining: pulls template html and css into the components.
+  - Minification: removes excess whitespace, comments, and optional tokens.
+  - Uglification: rewrites code to use short, cryptic variable and function names.
+  - Dead code elimination: removes unused code.
+  - Tree shaking: removes unreferenced modules.
```
robwormald: Dead code elimination is basically the same as Treeshaking, for the purposes of this doc

filipesilva: The `Tree shaking with rollup` section talks a bit more about it, and links to an article that elaborates on the differences.

public/docs/ts/latest/guide/production-deploy.jade
filipesilva: On the docs meetings a couple of real world scenarios came up. Deploying demos was one, and for some projects there is no time/feature budget to actually produce an optimized deployment since the business does not need it.

public/docs/ts/latest/guide/production-deploy.jade
filipesilva: There is value is describing how to deploy your dev code with the minimum set of changes. This is a question that comes up a couple of times every week.

I'll add a note that this isn't recommended per se, but rather a fallback.

public/docs/ts/latest/guide/production-deploy.jade
filipesilva: This section is meant to show the simplest possible deploy. Adding a build step would go against that. I'll clarify that further.

There is also the concern of adding another interim build step that isn't further detailed or documented. It's better to jump directly to the rollup+aot configuration for optimization.

public/docs/ts/latest/guide/production-deploy.jade
filipesilva: Will fix.

public/docs/ts/latest/guide/production-deploy.jade
filipesilva: Will reword to avoid that.

filipesilva:
@robwormald I added a commit addressing your suggestions:
- added section about importing RxJs operators
- added WebPageTest example
- removed 'huge' and 'slow' in favor of a better description
- added more encouragement to further pursue optimizations

robwormald:
This PR LGTM. couple of notes, more to @wardbell than anyone:

> in any case, I think of a smaller payload as just one factor that contributes to faster loading. It's significance in that respect depends a great deal on the target environment, yes?

As of today, script parsing time (which is a function of how much JS you ship to browser) is the #1 (by far) cost of getting angular apps booted, especially on mobile devices. This is worse with mega-vendor bundles.
>  In my (limited) experience, the far bigger factor, both in desktop and over mobile, is the number of trips to the server.
> In my (limited) experience, the number of files matters a lot whereas (total net) payload size makes comparatively little difference in app launch and startup time ... at least for enterprise apps. Is that your experience?

This is more server dependent (does your server support HTTP2 (push?) - if it does, somewhat counterintuitively, its much faster to load a bunch of smaller scripts over the single connection than one large script (!). This whole pattern is changing and I want to be cautious about getting overly prescriptive about these sorts of things while stuff is in flux.

:shipit:

wardbell:
@robwormald Thanks for the perf feedback and for demonstrating these points to me personally (Stephen did so too). I understand much better. I think we're silent on which factors are "more important" ... and silence is golden.

Thanks for your help!

PR #3067, opened 2017-01-03
Fix update new Year

wKoza:
Update date 2016 -> 2017


wardbell:
The CI sample tests had failures but for reasons that can't possibly relate to this PR. Merging anyway.

PR #3068, opened 2017-01-03
Docs(Style Guide) Add unsubscribe on Observable

wKoza:
Add unsubscribe on Observable in ngOnDestroy in Style Guide.


wardbell:
No. This is incorrect. There is no need to unsubscribe from an `http` call.

We will discuss when and how to manage Observable subscriptions in our forthcoming RxJS guide.

PR #3071, opened 2017-01-04
chore: convert relative templateUrl/styleUrls to have ./

filipesilva:
/cc @Foxandxss @wardbell


wardbell:
Thank you!

I dug into Style Guide which was hiding the `moduleId=module.id` in many examples for some reason ... and thus sending the wrong message!

Fixed that. Started adding/correcting advice, and soon had a lot of Style Guide "improvements".

I'll merge these with yours ... they overlap so much ... but keep the commits separate.

PR #3082, opened 2017-01-07
docs(forms-guide): explain heroForm reference better

UnnDunn:
The current forms guide uses the heroForm template reference variable without creating or explaining it first (see the last example in the "[Add a Hero and Reset the Form](https://angular.io/docs/ts/latest/guide/forms.html#reset)" section). This could cause confusion and presents errors if someone is following along on their machine.

This change modifies that section to add the explanation before the variable is used.


UnnDunn:
I added my Git email address to my Google account. Is there a way to get Googlebot to re-check the CLA status?

PR #3091, opened 2017-01-12
docs(hierarchical-di): Correct avoidance example (#3086) and more

### LINKED ISSUE #3086: Mistake on bad-alternative example
On https://angular.io/docs/ts/latest/guide/hierarchical-dependency-injection.html, the example of a bad-alternative has an error with the declarations and bootstrap. 
I think it is changed.
I know it's a bad alternative and should not be used, but I still think it should be correct.

// Don't do this!
@NgModule({
  imports: [
    BrowserModule,
    FormsModule
  ],
  providers: [ HeroesService, RestoreService ],
  **declarations: [ HeroesListComponent ],**
  **bootstrap: [
    HeroesListComponent,
    HeroCardComponent,
    HeroEditorComponent
  ]**
})

wardbell:
closes #3086
sample reworked to conform to our sample style, to make more sense, and cover the examples discussed
copy edits to bring closer to Google docs standards.


public/docs/_examples/hierarchical-dependency-injection/e2e-spec.ts
```diff
@@ -1,56 +1,106 @@
...
+  beforeAll(() => {
     browser.get('');
   });
 
-  it('should open with a card view', function () {
-    expect(element.all(by.cssContainingText('button', 'edit')).get(0).isDisplayed()).toBe(true,
-      'edit button should be displayed');
-  });
+  describe('Heroes Scenario', () => {
+    let page = {
```
filipesilva: Interesting usage of the Page Objects pattern. There really is no reason to have it separately in smaller tests.

wardbell: That's true. This page object is only for the heroes scenario so I thought it would be more clear here.

filipesilva: Not at all, I think we should have more page objects like this in our e2e tests overall.

PR #2896, opened 2016-11-29
docs(cb-dynamic-component-loading): add dynamic component loading coo…

thelgevold:

PR #3094, opened 2017-01-13
docs(style-guide): remove rule about aliasing inputs

### LINKED ISSUE #2891: StyleGuide - exception for avoiding renaming inputs?
The style guide suggest that we avoid renaming inputs and outputs, when possible.
https://angular.io/styleguide#!#05-13

I suggest that we add an exception for attribute directives, when the attribute directive doubles as an input.

For example, you might have a selector, `[prefixHighlightColor]`, and want to pass it as an input without the prefix, `@Input('prefixHighlightColor') highlightColor: string;`.

I'm suggesting this here, because of the no-input-rename rule in Codelyzer (which implements the style guide rules for linting). https://github.com/mgechev/codelyzer/issues/138 It seems that removing the prefix should not count as "renaming" for the no-input-rename rule. The rule seems sensible rule in any other case I can think of, so it would be a shame not to use it.

@mgechev

Foxandxss:
Closes #2891


wardbell:
I disagree. I personally revised this rule in order to satisfy #2896 while keeping its original intent which was to discourage spurious aliasing. 

Instead, you just removed it all together. Please explain.

wardbell:
I think we agreed to close this PR

PR #3097, opened 2017-01-14
[WIP] chore(hierarchical-dependency-injection): Subscriber -> Observer

wardbell:
**WIP - Do not merge until the question below has been answered**

Current `heroes.service.ts` code in the form of 
```
new Observable<Hero[]>((subscriber: Subscriber<Hero[]>) => {...}); // (1)
```
That conforms to the type of the parameter for [`Observable` constructor](https://github.com/blesh/RxJS/blob/master/src/Observable.ts#L33).

But it appears that [`Subscriber` isn't supposed to be used in "public API"](https://github.com/blesh/RxJS/blob/master/src/Subscriber.ts#L12)
and that the conventional thing to do (which works also) is to write.

```
new Observable<Hero[]>((observer: Observer<Hero[]>) => {...}); // (2)
```
I have asked those who know which is the preferred style ... (1) or (2) ... and if (2 ... as this PR would do), why is the arg to the ctor (and to the `create` method) typed as `Subscriber`?

### DIFF
--- a/public/docs/_examples/hierarchical-dependency-injection/ts/app/heroes.service.ts
+++ b/public/docs/_examples/hierarchical-dependency-injection/ts/app/heroes.service.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 
 import { Observable } from 'rxjs/Observable';
-import { Subscriber } from 'rxjs/Subscriber';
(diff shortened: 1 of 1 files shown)


wardbell:
The question was answered (Paul Taylor) and the `observer` is preferred.

PR #3107, opened 2017-01-17
chore: update tslint

Foxandxss:
I also went and added a few rules to tsconfig that are new in TS2 that were previously in tslint.


Foxandxss:
I am closing this. I will migrate it by hand to the new site.

PR #3115, opened 2017-01-18
Improve instruction on including angular.png in webpack demo

kbdguy:
Instruction to create the necessary folder and download.  Current version only says "download" (not where) and only appears if you hover over the image, so it is not clear.


Foxandxss:
Included at #3283 since it changes a few things as well.

PR #3130, opened 2017-01-24
docs(toh-5): add import example

kevinsearle:
Included sample of dashboard component import for consistency.

### DIFF
--- a/public/docs/_examples/toh-5/ts/app/app.module.3.ts
+++ b/public/docs/_examples/toh-5/ts/app/app.module.3.ts
@@ -6,7 +6,9 @@ import { RouterModule }   from '@angular/router';
 
 import { AppComponent }        from './app.component';
 import { HeroDetailComponent } from './hero-detail.component';
+// #docregion import-dashboard
 import { DashboardComponent }  from './dashboard.component';
+// #enddocregion import-dashboard
 import { HeroesComponent }     from './heroes.component';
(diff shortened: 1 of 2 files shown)


chalin:
Hi Keven. Each tutorial part is filled with details specific to that part. In order to keep the text focused, we'll sometimes omit showing the details of some elementary step. For example, by the someone reaches part 5 of the tutorial, they should be well equipped to know how write an import statement without having to be shown the exact text.

I'd vote to note make this addition. @wardbell @Foxandxss ?

kevinsearle:
Makes sense, however, for your consideration I would like to point out that while actually taking this tutorial for the first time this section tripped me up for a second due to this detail being included in previous tutorial chapters.

chalin:
Thanks for the feedback Kevin.

Taking a peek at the live example sources is an option. Might that have been enough to help?

kevinsearle:
That is what I did, however, I was thinking of future tutorial readers. I'll leave it in your hands to merge or not!

wardbell:
I feel your pain. We are about to scrub the tutorial and rewrite parts of it that will affect this section. I believe in our rework we will be much more careful to show the steps we talk about (such as imports ... a problem for some here and elsewhere) while removing some of the other repetitive stuff. For that reason only, I'm closing this PR.  Please bear with us while we are in transition.

PR #3139, opened 2017-01-25
Use Google CSE for Search

alexwolfe:
@naomiblack Ok, this one should be ready to go. Please take a look. Thanks!

### DIFF
--- a/public/_includes/_scripts-minimum.jade
+++ b/public/_includes/_scripts-minimum.jade
@@ -18,6 +18,16 @@ script.
 
   _st('install','VsuU7kH5Hnnj9tfyNvfK','2.0.0');
 
+script.
+  (function() {
+    var cx = '012747927672895622801:g9kk1mbseka';
+    var gcse = document.createElement('script');
+    gcse.type = 'text/javascript';
+    gcse.async = true;
+    gcse.src = 'https://cse.google.com/cse.js?cx=' + cx;
+    var s = document.getElementsByTagName('script')[0];
+    s.parentNode.insertBefore(gcse, s);
+  })();
 
 <!-- Google Feedback -->
 script(src="//www.gstatic.com/feedback/api.js" type="text/javascript")
--- a/public/_includes/_scripts-minimum.jade
+++ b/public/_includes/_scripts-minimum.jade
@@ -8,16 +8,7 @@ script.
   ga('create', 'UA-8594346-15', 'auto');
   ga('send', 'pageview')
 
-<!-- SWIFTYPE -->
-if current.path[0] == "docs" || current.path[0] == "search"
-script.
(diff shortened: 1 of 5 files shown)


IgorMinar:
hi @alexwolfe, I see the following error in console when using the app with this PR: 

```
Refused to display 'https://cse.google.com/cse_v2/ads?q=template&r=m&cx=012747927672895622801%3…%2Fdocs%2Fts%2Flatest%2F&referer=http%3A%2F%2Flocalhost%3A3000%2F#master-1' in a frame because it set 'X-Frame-Options' to 'SAMEORIGIN'.
```

Can this be resolved?

public/_includes/_scripts-minimum.jade
```diff
@@ -18,6 +18,16 @@ script.
```
IgorMinar: can you remove the swifty script? we shouldn't load it if we are not using it any more.

public/resources/css/module/_side-nav.scss
```diff
@@ -79,37 +79,37 @@ $sidenav-width: 240px;
...
+  //   }
+  //   &:-ms-input-placeholder { /* IE 10+ */
+  //     color: $blue-grey-100;
+  //     font-size: 12px;
+  //   }
+  //   &:-moz-placeholder { /* Firefox 18- */
+  //     color: $blue-grey-100;
+  //     font-size: 12px;
+  //   }
+  // }
```
IgorMinar: if we no longer need this css we should delete them. otherwise it's unclear why they are just commented out.

if we do need them we should add a comment why this is being commented out temporarily and what action is being taken to uncomment this and by whom.

alexwolfe:
Hey @IgorMinar I addressed your comments. In terms of the X-frame origin error, that is a setting that needs to be adjust on the site (Firebase). It should be set to: 
`X-Frame-Options: ALLOW-FROM https://www.google.com` 
where that option is available.

IgorMinar:
@alexeagle do you know what's the impact on the site if these resources are blocked? the basic
functionality seems to be working in spite of the errors. but it is concerning that there are errors at all.

all the firebase configuration is here: https://github.com/angular/angular.io/blob/master/firebase.json, so if you believe that the change you propose above is the right one then please make the change to this json file. thanks

alexeagle:
@alexwolfe I think Igor's above comment is for you

IgorMinar:
Yes! Sorry!

alexwolfe:
@IgorMinar Updated the firebase.json with the new x-frame options allowing google. Looks like the search worked either way but agree it's preferable not to have these errors.

wardbell:
Closing because we aren't developing the current angular.io in this repo.

PR #3160, opened 2017-01-29
Fixes OS/X deletion of non-essential files

leebenson:
Fixes the following error:

> xargs: illegal option -- a

### DIFF
--- a/public/docs/ts/latest/guide/setup.jade
+++ b/public/docs/ts/latest/guide/setup.jade
@@ -81,9 +81,7 @@ a#non-essential
 
   ### OS/X (bash)
 code-example(language="sh" class="code-shell").
-  xargs -a non-essential-files.txt rm -rf
-  rm app/*.spec*.ts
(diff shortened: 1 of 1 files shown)


leebenson:
I signed it!

trescenzi:
This seemingly doesn't solve the problem that it is intended to solve: angular/quickstart#369. The issue doesn't seem to be the command but the file being read itself. I've commented in the issue with more detail. If you'd like I can submit a PR against angular/quickstart that solves the problem.

leebenson:
Thanks @trescenzi. Some extra thoughts [here](https://github.com/angular/quickstart/issues/369#issuecomment-276422284)

wardbell:
Closed by https://github.com/angular/angular.io/commit/b2f41484721265026c87fb5c8780a8a4a57d5b61#diff-ea8e16eabf541f0e00287f1ba4bf7a69R79

PR #3166, opened 2017-01-30
chore(quickstart): restore app.module and main

wardbell:
We seemed to create more confusion by omitting app.module (and main) in QS than we gained in simplicity of QS. Every other example requires these files and you must learn about `app.module.ts` before you can add _anything_ to an Angular app.

This PR restores those two file to QS sample and plunker _but doesn't talk about them_ in QS, leaving _that discussion_ to the Setup guide.  Therefore, from a prose perspective, the QS is unchanged.

It also removes the `appmodule.jade` guide which ONLY talked about how to overcome the missing AppModule limitations of the QS.

@StephenFluin approved this change verbally but should give his LGTM.
@IgorMinar was involved in our original, mutual decision to omit these two files and should give his LGTM.

### DIFF
--- a/public/docs/_examples/_boilerplate/src/systemjs.config.web.build.js
+++ b/public/docs/_examples/_boilerplate/src/systemjs.config.web.build.js
@@ -79,58 +79,6 @@
     }
   });
 
-  if (global.autoBootstrap) { bootstrap(); }
-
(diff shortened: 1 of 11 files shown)


StephenFluin:
LGTM

PR #3169, opened 2017-01-31
docs(upgrade): rename Angular 2 to Angular

rsreimer:


Foxandxss:
Thanks, That is on me, but in my defense, they were pretty hard to find.

For the next time, you will like to create a PR from a branch in your fork, because now, your fork will remain out of sync of this repo.

PR #3173, opened 2017-01-31
docs(events): add NG-NL to event list

fvanwijk:
Added a row for NG-NL conference to the list of events


juleskremer:
Hello, angular.io/events are listings of events where an Angular Team Member will be presenting; if we find ourselves able to support ng-NL, we will add it to the list.

fvanwijk:
Ok thanks for the update and explanation :) Now I see it in the subheading

PR #3177, opened 2017-01-31
Changed the wording to reflect new location

shairez:
But if you ask me, this paragraph doesn't make much sense without understanding what shims are we talking about here

So it might be better to move it below as well?


wardbell:
Thx!

PR #3187, opened 2017-02-03
chore: disable some server-communication tests until next angular ver…

Foxandxss:
The https change was something I tried first, noticed that it was doing an unneeded redirection. Left it there.


public/docs/_examples/server-communication/e2e-spec.ts
```diff
@@ -78,7 +78,7 @@ describe('Server Communication', function () {
     }
   });
 
-  describe('Smarter Wikipedia Demo', function () {
+  xdescribe('Smarter Wikipedia Demo', function () {
```
wardbell: Why are you disabling this test?

PR #3206, opened 2017-02-06
Update dependency-injection.jade

Josh68:
docs: Dependency Injection

"Here's our AppModule where we register a `Logger`, a `UserService`, and an `APP_CONFIG` provider." No Logger appears to be registered in the source code.


wardbell:
True. For some reason, the logger registration was moved to the `AppComponent`.  Thanks for finding.

Josh68:
Thanks, Ward, and BTW, always love your threads.

-Josh

PR #3208, opened 2017-02-07
Ordered folders/files alphabetically

hxlnt:


public/docs/ts/latest/tutorial/toh-pt1.jade
```diff
@@ -11,19 +11,20 @@ include ../_util-fns
 .filetree
   .file angular-tour-of-heroes
   .children
+    .file node_modules ... 
+    .file package.json
```
wardbell: This would not appear here in an OS file listing which shows all folders/directories first followed by files.

OTOH, what really matters from a developer perspective is the `src` folder.

Finally, in a scrub of the tutorial "coming soon" we will eliminate all of these file trees because the code itself tells the story.

Therefore, I'd rather not pursue this change at this time although I thank you for your readiness to help.

public/docs/ts/latest/tutorial/toh-pt1.jade
```diff
@@ -186,7 +187,7 @@ code-example(language="html").
 .l-main-section
 :marked
   ## The Road Ahead
-  Our Tour of Heroes only displays one hero and we really want to display a list of heroes.
+  Our Tour of Heroes only displays one hero, and we really want to display a list of heroes.
```
wardbell: This change is a matter of style, not grammar. More importantly, we have significant edits coming that will make this sentence irrelevant. Thanks for helping though.

PR #3209, opened 2017-02-07
docs - change-log - Fix i18n link in changelog

klinki:
Link in changelog was pointing to old location of i18n docs (https://angular.io/docs/ts/latest/guide/i18n.html) instead of new one (https://angular.io/docs/ts/latest/cookbook/i18n.html)


klinki:
I signed it!

PR #3226, opened 2017-02-09
Update template-syntax.jade

Josh68:
docs: Template Syntax

Remove call-out to author (these should all be removed and issued as PRs, I think) and update code-referenced property "name" (not "firstName"). Also fix mis-named "phoneCall" method reference (s/b "callPhone").


wardbell:
Thanks!

Issue #3242, opened 2017-02-12
Contradiction in "HIERARCHICAL DEPENDENCY INJECTORS" Section


justinwilaby:
In the [HIERARCHICAL DEPENDENCY INJECTORS](https://angular.io/docs/ts/latest/guide/hierarchical-dependency-injection.html) section, the text: 

> An Angular application is a tree of components. Each component instance has its own injector! The tree of components parallels the tree of injectors.

is contradicted in the immediate paragraph that follows: 

> Angular doesn't actually create a separate injector for each component. Every component doesn't need its own injector and it would be horribly inefficient to create masses of injectors for no good purpose.

Do Components have their *own* injectors or don't they?  The term **own injector** is confounded with the term **an injector** and should be clarified.

Foxandxss:
I can be mistaken in the details in here, but the idea is that each component can have their own injection, but that doesn't meant that that injector is always created but only when you actually use it.

So a normal component won't create a new one, but if you set `providers` on it, it will.

justinwilaby:
Let's read these 2 sentences as if they were side by side:

> Each component instance has its own injector! Angular doesn't actually create a separate injector for each component.

It's a direct contradiction. Any English teacher would fail a student if this were used in their final thesis. The documents should provide better clarity.

Foxandxss:
We are open for improvements my friend. Please send a PR.

wardbell:
I see your point and agree that the text can be improved. However, it is not contradictory and those two sentences are bridged by an explanation that breaks the apparent contradiction ... a bridge that you break when you jam them side-by-side.

The mental model: every component _has_ its own injector.  You will always experience it that way. That's what the first sentence means.

Under the hood is often a proxy to the actual injector (which has a reference to the real one which may ... and probably is ...shared). There is only a need to create a real injector when you _provide_ a service at that level. Otherwise, Angular proxies up to the root. That's what the second sentence means.

Maybe we should have said it that way. We probably shouldn't say _anything_ about it because it's an implementation detail that has no consequences for the developer. 

We will leave this issue open until we correct the prose, one way or the other.

I'd like to take a moment to speak to you about how you framed your issue. Perhaps you are unaware that it came across as argumentative. The bit about "any English teacher would fail a student" is disrespectful and contrary to our Code of Conduct. Please keep the feedback coming but temper your language a bit.  We humans will respond better when you do.

PR #3246, opened 2017-02-13
Fixes issue #3242 - Contradiction in "HIERARCHICAL DEPENDENCY INJECOTRS" Section

justinwilaby:
This PR provides a fix for issue #3242 based on the following changes:

1. Fixes improper sentence structure.
2. Revises language to prevent starting a paragraph with "But".
3. Revises language to prevent placing the predicate before the subject in a non-question format.
4. Removes the contradictory language regarding "injectors".
5. Provides better clarity and sentence flow.
6. Reinforces the subject to the reader.

### DIFF
--- a/public/docs/ts/latest/guide/hierarchical-dependency-injection.jade
+++ b/public/docs/ts/latest/guide/hierarchical-dependency-injection.jade
@@ -22,8 +22,10 @@ block includes
 
   In fact, there is no such thing as *the* injector.
   An application may have multiple injectors.
-  An Angular application is a tree of components. Each component instance has its own injector!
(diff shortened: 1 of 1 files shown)


justinwilaby:
I signed it!

wardbell:
Thanks for the head start. I fear that your proposed language draws us into details that don't matter. As a developer, I should think "each component has its own injector". The fact that we use injector proxies wherever we can is an implementation detail that ... I don't think should matter. At most I might have a note a long the lines of "Angular may proxy the injector for efficiency".

So, while we much appreciate your effort ... and I'll hold this PR open until we fix the text ... we're unlikely to accept it as written.

PR #3248, opened 2017-02-13
docs(hierarchical-di): clarify that component injector may be a proxy

### LINKED ISSUE #3242: Contradiction in "HIERARCHICAL DEPENDENCY INJECTORS" Section

wardbell:
closes #3242 and #3246

### DIFF
--- a/public/docs/ts/latest/guide/hierarchical-dependency-injection.jade
+++ b/public/docs/ts/latest/guide/hierarchical-dependency-injection.jade
@@ -20,10 +20,17 @@ block includes
   In the [Dependency Injection](./dependency-injection.html) guide,
   you learned how to configure a dependency injector and how to retrieve dependencies where you need them.
 
-  In fact, there is no such thing as *the* injector.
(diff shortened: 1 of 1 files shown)


Foxandxss:
I like the change, but I think that we should explain when it is a proxy and when it is not. WDYT?

wardbell:
I'd prefer not to do so. That's an implementation detail that doesn't affect developer behavior ... nor should it. If you really really want to know ... read the code.

PR #3283, opened 2017-02-20
docs(webpack): one step closer to CLI

Foxandxss:
This is one step closer to what the CLI has.

I decided **not** to move the `karma-test-shim` to `src/` because I like the `config` folder as it sit.

Included the PR #3115 since it needed some changes after this.


wardbell:
Looks great. 

One question before we merge ... do you want to include the Karma Jasmine HTML Reporter for tests? Many of us really like it and it just made it into the CLI.

The package.json is
```
    "karma-jasmine-html-reporter": "^0.2.2",
```
The changes to karma.conf.js that enable it are visible at https://github.com/angular/angular-cli/blob/master/packages/%40angular/cli/blueprints/ng2/files/karma.conf.js

Your call. Feel free to merge now or after adding that.

PR #3284, opened 2017-02-20
docs(attribute-directives): copy tweaks

kapunahelewong:
Cleaning up a few copy details.

### DIFF
--- a/public/docs/ts/latest/guide/attribute-directives.jade
+++ b/public/docs/ts/latest/guide/attribute-directives.jade
@@ -49,21 +49,11 @@ a#write-directive
   the attribute.
   The controller class implements the desired directive behavior.
 
-  This page demonstrates building a simple attribute
+  This page demonstrates building a simple _myHighlight_ attribute
   directive to set an element's background color
-  when the user hovers over that element.
-.l-sub-section
-  :marked
-    Technically, a directive isn't necessary to simply set the background color. Style binding can set styles as follows:
+  when the user hovers over that element. You can apply it like this:
 
-  +makeExample('attribute-directives/ts/src/app/app.component.1.html','p-style-background')
-
-  :marked
-    Read more about [style binding](template-syntax.html#style-binding) on the [Template Syntax](template-syntax.html) page.
-
-    For a simple example, though, this demonstrates how attribute directives work.
-
-
++makeExample('attribute-directives/ts/src/app/app.component.1.html','applied')
 
 :marked
   ### Write the directive code
@@ -210,7 +200,7 @@ a#bindings
   ## Pass values into the directive with an _@Input_ data binding
 
   Currently the highlight color is hard-coded _within_ the directive. That's inflexible.
-  Let the user of the directive set the color in the template with a binding.
+  In this section, you give the developer the power to set the default color while applying the directive.
 
   Start by adding a `highlightColor` property to the directive class like this:
 +makeExample('attribute-directives/ts/src/app/highlight.directive.2.ts','color', 'src/app/highlight.directive.ts')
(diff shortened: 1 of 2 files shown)


public/docs/ts/latest/guide/attribute-directives.jade
```diff
@@ -25,12 +25,12 @@ a#directive-overview
...
-  1. Components &mdash; directives with a template.
-  1. Structural directives &mdash; change the DOM layout by adding and removing DOM elements.
-  1. Attribute directives &mdash; change the appearance or behavior of an element, component, or another directive.
+  1. Components&mdash;directives with a template.
+  1. Structural directives&mdash;change the DOM layout by adding and removing DOM elements.
+  1. Attribute directives&mdash;change the appearance or behavior of an element, component, or another directive.
 
   *Components* are the most common of the three directives.
-  You saw a component for the first time in the [QuickStart](../quickstart.html) example.
+  You saw a component for the first time in the [QuickStart](../quickstart.html) guide.
```
chalin: I thought that we were to use the term "page" instead of "guide" (although, in this case, the original "example" seems to be ok too).

public/docs/ts/latest/guide/attribute-directives.jade
```diff
@@ -61,17 +61,17 @@ a#write-directive
   :marked
     Read more about [style binding](template-syntax.html#style-binding) on the [Template Syntax](template-syntax.html) page.
 
-    For a simple example, though, this will demonstrate how attribute directives work.
+    For a simple example, though, this demonstrates how attribute directives work.
```
wardbell: This aside about style binding is a mistake. FIrst, you couldn't use the style binding to set the background _when the user hovers_. Second, more importantly, this aside is distraction. 

In a related PR I delete this aside and replace it with a one line example of how the new directive WILL apply to an element.  I'll fix the resulting merge conflict.

public/docs/ts/latest/guide/attribute-directives.jade
wardbell: Do as you will in Dart land. We're doing differently here.

public/docs/ts/latest/guide/attribute-directives.jade
```diff
@@ -61,17 +61,17 @@ a#write-directive
...
 
 
 
 :marked
   ### Write the directive code
-  Follow the [setup](setup.html) instructions for creating a new project
+  Follow the [setup](setup.html) instructions for creating a new local project
```
wardbell: I'm not against it entirely. But I don't see that the word "local" helps. If you agree, let's revert. If you don't, leave it.

public/docs/ts/latest/guide/attribute-directives.jade
```diff
@@ -300,8 +300,9 @@ a#second-property
   ## Bind to a second property
   This highlight directive has a single customizable property. In a real app, it may need more.
 
-  At the moment, the default color &mdash; the color that prevails until the user picks a highlight color &mdash; 
-  is hard-coded as "red". Let the template developer set the default color.
+  At the moment, the default color&mdash;the color that prevails until 
+  the user picks a highlight color&mdash;is hard-coded as "red". 
+  Let the template developer set the default color.
```
wardbell: In my commit, I change this to:
>In this section, you give the developer the power to set the default color while applying the directive.

public/docs/ts/latest/guide/attribute-directives.jade
chalin: Okey dokey. I just thought it was a TS land convention as well, but apparently no longer.

wardbell: Not on my watch ;-)

PR #3286, opened 2017-02-21
Fix typo (repeated file names)

AlexanderLukin:


public/docs/ts/latest/guide/reactive-forms.jade
```diff
@@ -775,7 +775,7 @@ figure.image-display
   If you're coding along with the steps in this reactive forms tutorial, 
   create the pertinent files based on the 
   [source code displayed below](#source-code "Reactive Forms source code"). 
-  Notice that `hero-list.component.ts` and `hero-list.component.ts` 
+  Notice that `hero-list.component.ts` and `hero.service.ts` 
   import `Observable` and `finally` from `rxjs`.
```
kapunahelewong: This is a great catch, @AlexanderLukin! I was looking closely at the example files and noticed something though. How about this (or something like it):

> Notice that `hero-list.component.ts` imports `Observable` and `finally` while `hero.service.ts` imports `Observable`, `of`, and `delay` from `rxjs`.

AlexanderLukin: Yes, I think you've proposed the most accurate version. I've corrected my PR in accordance with your changes.

PR #3289, opened 2017-02-21
fix(toh-6): add missing tsconfig file

petebacondarwin:


public/docs/_examples/toh-6/ts/src/tsconfig.1.json
```diff
@@ -0,0 +1,21 @@
...
+    "target": "es5",
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "sourceMap": true,
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+    "lib": [ "es2015", "dom" ],
+    "noImplicitAny": true,
+    "suppressImplicitAnyIndexErrors": true,
+    "typeRoots": [
```
filipesilva: Can you use https://github.com/angular/angular.io/blob/master/public/docs/_examples/quickstart/ts/src/tsconfig.1.json here instead of the auto-generated one? It's incorrect insofar as it contains some invalid excludes and typeRoots for user projects.

petebacondarwin: OK

PR #3290, opened 2017-02-21
(docs) Updates for working Webpack iwith latest Angular version

Boyan-Kostadinov:
Removed the options from the TS loader rules as it was breaking the packer
Added webpack-env types as WebPack failed to build without it
Updated other packages to their latest versions

Working repository at https://github.com/Boyan-Kostadinov/angular-webpack


Foxandxss:
Thanks, there is a PR to fix the guide  already :)

PR #3293, opened 2017-02-22
Fix build for webpack example

Davepar:
The webpack example needs to use a newer Typescript compiler and point to the default location for tsconfig.json file in order to build. Right now it's failing with:

    Error in bail mode: [at-loader] node_modules/@types/jasmine/index.d.ts:39:52
    TS1005: '=' expected.

And

    Error in bail mode: [at-loader] node_modules/@angular/common/src/directives/ng_class.d.ts:48:34
    TS2304: Cannot find name 'Set'.

### DIFF
--- a/public/docs/_examples/package.json
+++ b/public/docs/_examples/package.json
@@ -79,7 +79,7 @@
     "style-loader": "^0.13.1",
     "ts-node": "^1.3.0",
     "tslint": "^3.15.1",
-    "typescript": "~2.0.10",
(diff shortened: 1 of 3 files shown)


Foxandxss:
We won't use TS 2.1 for the time being.

There is a PR with many fixes to the webpack guide already. Will be merged tomorrow.

Davepar:
Ok, cool. Thanks.

PR #3295, opened 2017-02-23
docs: update CLI quickstart to rc.0

filipesilva:
RC0 hasn't been released yet but will soon.

All shown boilerplate is copy paste of a new CLI generated project using the rc.0 blueprints.

Formatting might be a bit different from the norm in the angular.io repo but it should be left as is, since the shown files should match a generated project.

For angular.io tests to pass this PR is still using the `@angular/cli@1.0.0-beta.32.3` npm version however.

Prior to merging it should be updated to `1.0.0-rc.0` in these two places:

- `public/docs/_examples/package.json`
- `public/docs/_examples/_boilerplate/package.json`


public/docs/ts/latest/cli-quickstart.jade
```diff
@@ -287,14 +293,22 @@ table(width="100%")
...
     td <code>node_modules/...</code>
     td
       :marked
         `Node.js` creates this folder and puts all third party modules listed in 
         `package.json` inside of it.
   tr
+    td <code>angular-cli.json</code>
+    td
+      :marked
+        Configuration for Angular-CLI.
```
Foxandxss: no dash

filipesilva: Good catch!

PR #3296, opened 2017-02-23
Fix base tag subfolder example

jerolba:
The sample url for production server is "http://www.mysite.com/mysrc/app/"
but the identified subfolder is "my/app", which doesn't match.


jerolba:
I signed it!

filipesilva:
LGTM, thanks for the fix!

PR #3319, opened 2017-02-28
build: zipper revival

Foxandxss:


public/docs/ts/latest/guide/change-log.jade
```diff
@@ -5,6 +5,9 @@ block includes
   The Angular documentation is a living document with continuous improvements.
   This log calls attention to recent significant changes.
 
+  ## NEW: Downloadable examples for each guide (2017-02-28)
+  All the guides have now a download link where you can run their examples locally.
```
wardbell: >Now you can download the sample code for any guide and run it locally. Look for the new download links next to the "live example" links.

PR #3327, opened 2017-03-01
fix(zipper): add ‘use strict’ to exampleZipper.js

wardbell:
To beat error “SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode”.


Foxandxss:
I haven't seen that, but yeah, that can happen.

wardbell:
Travis is going bonkers but we know this line is harmless at worst so I'm merging it.

PR #3329, opened 2017-03-01
Cf edits copy

juleskremer:


public/docs/ts/latest/glossary.jade
```diff
@@ -237,7 +238,8 @@ a#component
   ## Dependency injection
 .l-sub-section
   :marked
-    Dependency injection is both a design pattern and a mechanism
+
```
petebacondarwin: I think this is your problem @juleskremer 
The whitespace after the `:marked` mixin is not allowed

public/docs/ts/latest/glossary.jade
```diff
@@ -604,31 +613,23 @@ a#Q
...
-  block snake-case-defn
-    :marked
-      The practice of writing compound words or phrases such that an
-      underscore (`_`) separates one word from the next. This form is also known as **underscore case**.
-
 :marked
   ## Service
 .l-sub-section
   :marked
+
```
petebacondarwin: And maybe here too

PR #3335, opened 2017-03-03
docs(hier-dep-inj): fix comp name, spelling in field name, copyedits

chalin:
- Move `@Input()` to be on setter rather than getter.
- Fix spelling of field `villaines` --> `villains`
- `TaxReturnComponent` --> `HeroTaxReturnComponent`
- Other misc copyedits

### DIFF
--- a/public/docs/_examples/hierarchical-dependency-injection/ts/src/app/hero-tax-return.component.ts
+++ b/public/docs/_examples/hierarchical-dependency-injection/ts/src/app/hero-tax-return.component.ts
@@ -15,10 +15,11 @@ import { HeroTaxReturnService } from './hero-tax-return.service';
 export class HeroTaxReturnComponent {
   message = '';
   @Output() close = new EventEmitter<void>();
-  @Input()
+
   get taxReturn(): HeroTaxReturn {
     return this.heroTaxReturnService.taxReturn;
   }
+  @Input()
   set taxReturn (htr: HeroTaxReturn) {
     this.heroTaxReturnService.taxReturn = htr;
   }
--- a/public/docs/ts/latest/guide/hierarchical-dependency-injection.jade
+++ b/public/docs/ts/latest/guide/hierarchical-dependency-injection.jade
@@ -96,7 +96,7 @@ figure.image-display
 +makeExcerpt('src/app/villains-list.component.ts (metadata)')
 
 :marked
-  By providing `VillainsService` in the `VillainsListComponent` metadata &mdash; and nowhere else &mdash;
+  By providing `VillainsService` in the `VillainsListComponent` metadata and nowhere else,
   the service becomes available only in the `VillainsListComponent` and its sub-component tree.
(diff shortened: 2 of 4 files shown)


public/docs/ts/latest/guide/hierarchical-dependency-injection.jade
```diff
@@ -93,9 +93,10 @@ figure.image-display
 
   Instead, provide the `VillainsService` in the `providers` metadata of the `VillainsListComponent` like this:
 
-+makeExample('hierarchical-dependency-injection/ts/src/app/villains-list.component.ts', 'metadata','src/app/villains-list.component.ts (metadata)')(format='.')
++makeExcerpt('src/app/villains-list.component.ts (metadata)')
+
 :marked
-  By providing `VillainsService` in the `VillainsListComponent` metadata &mdash; and nowhere else &mdash;, 
+  By providing `VillainsService` in the `VillainsListComponent` metadata &mdash; and nowhere else &mdash;
```
wardbell: The comma was correct but ugly.  Please rephrase:
>By providing `VillainsService` in the `VillainsListComponent` metadata and nowhere else,

public/docs/_examples/hierarchical-dependency-injection/ts/src/app/hero-tax-return.component.ts
```diff
@@ -15,10 +15,11 @@ import { HeroTaxReturnService } from './hero-tax-return.service';
 export class HeroTaxReturnComponent {
   message = '';
   @Output() close = new EventEmitter<void>();
-  @Input()
+
   get taxReturn(): HeroTaxReturn {
     return this.heroTaxReturnService.taxReturn;
   }
+  @Input()
```
wardbell: Why did you move this here?  I realize it is technically true that the `@Input` applies to the setter but does it looks awful and it doesn't change Angular's behavior, right?

public/docs/ts/latest/guide/hierarchical-dependency-injection.jade
chalin: done

public/docs/_examples/hierarchical-dependency-injection/ts/src/app/hero-tax-return.component.ts
chalin: As a matter of style, we should encourage readers to associate `@Input` with setters. This way, documentation and refactoring tools, for example, will attribute the decorator to the proper class member.

Right, Angular's behavior isn't affected (though IMHO it might be incidental behavior that it happens to work when attached to the getter).

If you feel strongly about it, I can undo it. Let me know.

wardbell: I don't feel strongly about it  ... yet.

If we really believe that `@Input` should adorn the setter, we should find a place in the docs to say that and say why. If you have a moment to look for that place and propose language for it, I'd be grateful.

That would be a separate PR. I'm merging this one.

PR #3345, opened 2017-03-05
docs(structural-directives): copyedits and some tweaks to code

chalin:
- Fix variable name `li` --> `i`
- a --> an `UnlessDirective`
- Jade had both #ngswitch and #ngSwitch as link targets;
  fixed to use same capitalization as template syntax page.
- Drop "scrap.txt" file
- Fixed broke fragment target (was missing #)
- `app/` --> `src/app/` in prose summary
- Other copy edits

### DIFF
--- a/public/docs/_examples/structural-directives/ts/src/app/app.component.html
+++ b/public/docs/_examples/structural-directives/ts/src/app/app.component.html
@@ -38,11 +38,11 @@
 
 <!-- #docregion display-none -->
 <p [style.display]="'block'">
-  Expression sets display to "block"" .
+  Expression sets display to "block".
   This paragraph is visible.
 </p>
 <p [style.display]="'none'">
-  Expression sets display to "none" .
+  Expression sets display to "none".
   This paragraph is hidden but still in the DOM.
 </p>
 <!-- #enddocregion display-none -->
@@ -152,11 +152,9 @@
 
 <div>Pick your favorite hero</div>
 <p>
-  <ng-container *ngFor="let h of heroes">
-    <label>
-      <input type="radio" name="heroes" [(ngModel)]="hero" [value]="h">{{h.name}}
-    </label>
-  </ng-container>
+  <label *ngFor="let h of heroes">
+    <input type="radio" name="heroes" [(ngModel)]="hero" [value]="h">{{h.name}}
+  </label>
   <label><input type="radio" name="heroes" (click)="hero = null">None of the above</label>
 </p>
 
--- a/public/docs/ts/latest/guide/structural-directives.jade
+++ b/public/docs/ts/latest/guide/structural-directives.jade
@@ -1,7 +1,7 @@
 block includes
   include ../_util-fns
 
-// The docs standard h4 style uppercases, making code terms unreadable. Override it.
+//- The docs standard h4 style uppercases, making code terms unreadable. Override it.
 style.
   h4 {font-size: 17px !important; text-transform: none !important;}
   .syntax { font-family: Consolas, 'Lucida Sans', Courier, sans-serif; color: black; font-size: 85%; }
@@ -15,13 +15,13 @@ style.
   - [What are structural directives?](#definition)
   - [*NgIf* case study](#ngIf)
   - [Group sibling elements with &lt;ng-container&gt;](#ng-container)
-  - [The asterisk (\*) prefix](#asterisk)
-  - [Inside *NgFor*](#ngfor)
+  - [The asterisk (*) prefix](#asterisk)
+  - [Inside *NgFor*](#ngFor)
     - [microsyntax](#microsyntax)
     - [template input variables](#template-input-variable)
     - [one structural directive per element](#one-per-element)
   - [Inside the *NgSwitch* directives](#ngSwitch)
-  - [Prefer the (\*) prefix](#prefer-asterisk)
+  - [Prefer the (*) prefix](#prefer-asterisk)
   - [The &lt;template> element](#template)
   - [Write a structural directive](#unless)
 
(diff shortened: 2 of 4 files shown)


public/docs/_examples/structural-directives/ts/src/app/app.component.html
```diff
@@ -152,11 +152,9 @@ <h2 id="ngSwitch">NgSwitch</h2>
 
 <div>Pick your favorite hero</div>
 <p>
-  <ng-container *ngFor="let h of heroes">
-    <label>
-      <input type="radio" name="heroes" [(ngModel)]="hero" [value]="h">{{h.name}}
-    </label>
-  </ng-container>
+  <label *ngFor="let h of heroes">
```
wardbell: Moving the ngFor to the label makes a lot of sense and we should do it **UNLESS this is an example of `<ng-container>`**.  If it is, please restore. If it isn't, then your change is an improvement.

public/docs/_examples/structural-directives/ts/src/app/scrap.txt
```diff
@@ -1,21 +0,0 @@
-// interesting but unused code
```
wardbell: Please restore. This is harmless and I wanted to keep it around as an idea that I had. Sure, I'll probably forget all about it (I already did). But humor me.

public/docs/ts/latest/guide/structural-directives.jade
```diff
@@ -15,20 +15,19 @@ style.
...
-  - [Prefer the (\*) prefix](#prefer-asterisk)
+  - [Prefer the (*) prefix](#prefer-asterisk)
   - [The &lt;template> element](#template)
   - [Write a structural directive](#unless)
 
   Try the <live-example></live-example>.
 
-a#definition
-.l-main-section
+.l-main-section#definition
```
wardbell: Please don't "fix" these. I've created separate anchors on purpose, even where the header text would match the anchor name.  I break fewer links this way and don't have to worry about what happens when I change a heading. I can also quickly find something when editing. So these kinds of changes actually make my life harder.

Most importantly, we're converting to pure markdown in v.2 and this makes it harder to figure out what's going on. 

**TAKE AWAY: Please revert them all and do not make further changes to structure/layout that have no visual effect**

public/docs/ts/latest/guide/structural-directives.jade
```diff
@@ -40,13 +39,16 @@ a#definition
   The directive then does whatever it's supposed to do with that host element and its descendents.
 
   Structural directives are easy to recognize. 
-  An asterisk (\*) precedes the directive attribute name as in this example.
-+makeExample('structural-directives/ts/src/app/app.component.html', 'ngif')(format=".")
+  An asterisk (*) precedes the directive attribute name as in this example.
+
++makeExcerpt('src/app/app.component.html', 'ngif', '')
```
wardbell: Yes, this is shorter and might be worth while if we were staying with Jade.. But we're converting all of this anyway and this kind of move is more of a distraction than a help right now.

I'll accept in THIS PR but please don't do this on future PRs.

public/docs/ts/latest/guide/structural-directives.jade
```diff
@@ -229,72 +250,68 @@ code-example(language="javascript").
...
-        Write `[ngIf]="hero"`, not `ngIf="hero"`!
-        The latter incorrectly assigns the *string* `"hero"` to `ngIf`.
-        A non-empty string is _truthy_, so `ngIf` is always `true` and Angular 
-        always tries to show the content … even when there is no `hero`.
++makeExcerpt('src/app/app.component.html', 'ngif-template', '')
 
 :marked
-  * The `*ngIf` directive moved to the `<template>` tag where it became a property binding,`[ngIf]`.
-  * The rest of the `<div>`, including its class attribute, moved inside the `<template>` tag.
+  * The `*ngIf` directive moved to the `<template>` element where it became a property binding,`[ngIf]`.
```
wardbell: Note to self:  all `<template>` will become `<ng-template>` in Angular v.4.   I have to think about what this does to what we've been saying here.

public/docs/ts/latest/guide/structural-directives.jade
```diff
@@ -436,49 +456,50 @@ a#template
   It is never displayed directly. 
   In fact, before rendering the view, Angular _replaces_ the `<template>` and its contents with a comment.
   
-  If there is no structural directive, if you merely wrap some elements in a `<template>` and do nothing with it, 
+  If there is no structural directive, and you merely wrap some elements in a `<template>` and do nothing with it,
```
wardbell: Ugh. I think the following is better ... we don't need to say "and do nothing with it"
>If there is no structural directive and you merely wrap some elements in a `<template>`, those elements disappear.

public/docs/ts/latest/guide/structural-directives.jade
```diff
@@ -574,9 +599,10 @@ a#summary
 
 :marked
   You learned
+
   * that structural directives manipulate HTML layout.
   * to use [`<ng-container>`](#ngcontainer) as a grouping element when there is no suitable host element.
-  * that the angular "de-sugars" [asterisk (\*) syntax](#asterisk) into a `<template>`.
+  * that the angular "de-sugars" [asterisk (*) syntax](#asterisk) into a `<template>`.
```
wardbell: Please fix as:
>that Angular "de-sugars" the ...

public/docs/_examples/structural-directives/ts/src/app/app.component.html
chalin: It is not an example of `<ng-container>`. This code is not shown in the page, it is only there to support the `ngSwitch` code which is shown in the page.

public/docs/_examples/structural-directives/ts/src/app/scrap.txt
chalin: Done. (I thought that this file showed up in the plnkr and zip, but I just checked and it doesn't.)

public/docs/ts/latest/guide/structural-directives.jade
chalin: All reverted.

> do not make further changes to structure/layout that have no visual effect

Ok.

public/docs/ts/latest/guide/structural-directives.jade
chalin: Ok. Noted.

public/docs/ts/latest/guide/structural-directives.jade
chalin: done

public/docs/ts/latest/guide/structural-directives.jade
chalin: Done.
Also, I dropped the hyphen in "desugars" (Kathy confirmed a few days ago that there shouldn't be any hyphen).

chalin:
All points addressed. And I've noted not to edit the `+makeEx*` mixin applications anymore (well, that means for the two files I have left to review :).

chalin:
@wardbell @kapunahelewong - the Structural Directives page currently has (essentially) this **Table of Contents**:
- What are structural directives?
- `NgIf`
- Group sibling elements with `<ng-container>`
- `NgFor`
- `NgSwitch*`
- `<template>`

I think it would **improve the flow** if we covered _all_ the structural directives first (since that is the purpose of this guide), before mentioning `<ng-container>`.

I'd like to suggest that we move the `<ng-container>` section like so:
- What are structural directives?
- `NgIf`
- `NgFor`
- `NgSwitch*`
- `<template>`
- **Group sibling elements with `<ng-container>`** <======

I can do it in this PR if you agree.

wardbell:
As noted in Slack, I'd like to get this PR in and then let Kapunahele and you work out the TOC question in a followup PR.

PR #3367, opened 2017-03-10
docs(structural-directives): move ng-container section later in page

chalin:
This is a follow-up to https://github.com/angular/angular.io/pull/3345#issuecomment-285511144.

- No changes were done to the `<ng-container>` section prose.
- At the end of the template section I linked the "write your own structural directive" text to the "unless" directive section (because now there is `<ng-container>` in between the two).

@wardbell: @kapunahelewong and I discussed the section relocation, and she agrees with the change -- other edits to this guide, if any, will come in follow-up PRs.

### DIFF
--- a/public/docs/ts/latest/guide/structural-directives.jade
+++ b/public/docs/ts/latest/guide/structural-directives.jade
@@ -14,7 +14,6 @@ style.
 
   - [What are structural directives?](#definition)
   - [*NgIf* case study](#ngIf)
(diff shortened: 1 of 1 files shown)


chalin:
Oups, it seems that I omitted to back propagate some copyedits. They've been added in the separate "post review" commit.

wardbell:
This will do for now. I'll be revisiting shortly for v.4 to cope with changes to the microsyntax and the move to `<ng-template>` whose implications I have not yet thought through.

PR #3370, opened 2017-03-12
docs: fix typos

charleshypk:
Fixing minor typos in order to be consistent.


public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -740,8 +740,7 @@ p
   | (like the title of the application or the address of a web API endpoint)
   block config-obj-maps
     | &nbsp;but these configuration objects aren't always instances of a class.
```
Foxandxss: why not this weird space?

charleshypk: Ops, I'm not sure how I missed that! I pushed a new commit that is consistent with the previous p tag.

PR #3373, opened 2017-03-14
docs(server-comm): copyedits

chalin:
- Minor prose copyedits.
- Renamed `HeroService.addHero()` to `create()`, so as to match the ToH sample.


public/docs/_examples/server-communication/ts/src/app/toh/hero.service.ts
```diff
@@ -21,32 +21,32 @@ import { Hero } from './hero';
 @Injectable()
 export class HeroService {
   // #docregion endpoint
-  private heroesUrl = 'app/heroes';  // URL to web API
+  private heroesUrl = 'api/heroes';  // URL to web API
```
wardbell: I thought we fixed that long ago but apparently not.

Issue #2436, opened 2016-09-21
perform image optimization


chalin:
Follow-up to a discussion with @robwormald, @StephenFluin.
This will benefit ng1.io and ng2.io.

If it is only a question of setting up a gulp task, I don't mind doing the legwork, but I'll need to know which commands to use. Might there be an example of such a task already in some other angular project somewhere?

cc @ericjim @naomiblack

chalin:
cc @alexwolfe

chalin:
I came across this while working on site-webdev:
> Use zopfli to compress PNGs.
>  `brew install zopfli`
> Then run it as described in https://ariya.io/2016/06/using-zopfli-to-optimize-png-images.

PR #3384, opened 2017-03-16
chore: devguide image compression

chalin:
Backport from Dart docs site of TS relevant images.

Fixes https://github.com/angular/angular.io/issues/2436

Courtesy of @kevmoo
cc @kwalrath


wardbell:
Why are we backporting images? What do they have that the current images lack? What's up?

I can't tell what's different between the images nor what this PR is about.

Blocking while we await answers.

kevmoo:
@wardbell github should show file sizes – these are a lot smaller 😄

chalin:
Images are the same, just with smaller file sizes. Something that Naomi asked be done a while back (https://github.com/angular/angular.io/issues/2436).

wardbell:
I didn't see significant size changes in the handful I checked. A few 100 bytes here and there.  And I didn't have time inspect the images that github wouldn't diff.  

I'm accepting your word (Patrice and Kevin) that this will all be fine.

PR #3388, opened 2017-03-18
[WIP] aio(cheatsheet): convert TS cheatsheet to pure HTML (in jade file)

wardbell:
Intended for `angular/aio` docs so that we don't have to port the AngularJS cheatsheet creation code (there's a lot of it) over to Angular. The cheatsheet hardly ever changes and would be easy to edit by hand.

This might even work in our angular.io world. I'm a little afraid until someone shows that it won't break our current site.

Need someone to generate the staging site with this PR to see if it works.

Alternatively, @petebacondarwin may find a way to grab this for `angular/aio` and we can leave the current state untouched.

#### Why wasn't this in markdown to begin with?

... and why didn't I convert it to MD now?

It can't be done in Markdown because of the multi-row stretches of code. There's now way to write those inside MD tables with triple-backticks. That's why we're generating with AngularJS today

### DIFF
--- a/public/docs/ts/latest/cheatsheet.jade
+++ b/public/docs/ts/latest/cheatsheet.jade
@@ -1,4 +1,0 @@
-- var base = current.path[4] ? '.' : './guide';
-
-.l-content-small.grid-fluid.docs-content.cheatsheet
-  ngio-cheatsheet(src= base + '/cheatsheet.json')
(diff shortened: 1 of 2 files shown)


petebacondarwin:
You also get to clear out the directive at https://github.com/angular/angular.io/blob/master/public/resources/js/directives/cheatsheet.js

petebacondarwin:
For another I think we should capture the HTML and then use the migrator to manually copy it into the air content folder as part of the migration processing. That will free you up to do what you want in angular.io

filipesilva:
Included in https://github.com/angular/angular.io/pull/3415

PR #3411, opened 2017-03-22
New cb third party lib

filipesilva:
Followup from https://github.com/angular/angular.io/pull/2758

Uses https://github.com/filipesilva/angular-quickstart-lib while that code is not transferred over to a Angular repo.


sirhcybe:
@filipesilva I'm trying to build this and having some trouble. The npm install is failing because I am missing the dependency on angular-quickstart-lib. However, I am not sure how to build that file and the commands in the README are not working.

filipesilva:
@sirhcybe the starter mentioned in this document can be found at https://github.com/filipesilva/angular-quickstart-lib. The needed commands are present in that repository's package.json.

sirhcybe:
@filipesilva Ah yes, that works much better. Thank you.

PR #3414, opened 2017-03-22
create() should return Hero model.

bchristie:
To keep things uniform, i suspect this should be cast like `getHero` & `getHeroes`.


bchristie:
I signed it!

PR #3418, opened 2017-03-23
chore: make IE11 work with the moduleId loader

Foxandxss:


wardbell:
Yes. The `URL` class is explicitly "experimental" and only supported in a handful of browsers. Let's try never to use it again :-)

https://developer.mozilla.org/en-US/docs/Web/API/URL

PR #3431, opened 2017-03-24
docs: followup amends for 4.0 release

filipesilva:
/cc @wardbell

### DIFF
--- a/public/docs/ts/latest/guide/animations.jade
+++ b/public/docs/ts/latest/guide/animations.jade
@@ -49,8 +49,9 @@ figure
   driven by a model attribute.
 
   Animations are defined inside `@Component` metadata. Before you can add animations, you need
-  to import a few animation-specific functions:
+  to import a few animation-specific imports and functions:
 
++makeExample('animations/ts/src/app/app.module.ts', 'animations-module', 'app.module.ts (@NgModule imports excerpt)')(format=".")
 +makeExample('animations/ts/src/app/hero-list-basic.component.ts', 'imports', 'hero-list-basic.component.ts')(format=".")
 
 :marked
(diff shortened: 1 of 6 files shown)


public/docs/ts/latest/guide/animations.jade
```diff
@@ -353,7 +354,7 @@ figure
 +makeExample('animations/ts/src/app/hero-list-multistep.component.ts', 'template', 'hero-list-multistep.component.ts (excerpt)')(format=".")
 
 :marked
-  The callbacks receive an `AnimationTransitionEvent` which contains useful properties such as `fromState`,
+  The callbacks receive an `AnimationEvent` which contains useful properties such as `fromState`,
```
wardbell: Should be `that contains` but that is a grammar nit.

public/docs/ts/latest/guide/dependency-injection.jade
```diff
@@ -890,7 +890,7 @@ block dart-map-alternative
   An `Injector` is itself an injectable service.
 
   In this example, Angular injects the component's own `Injector` into the component's constructor.
-  The component then asks the injected injector for the services it wants.
+  The component then asks the injected injector for the services it wants on `ngOnInit`.
```
wardbell: OK but better would be "in `ngOnInit()`. Use  "in" instead of "on"

Kapunahele has started highlighting method names by putting parens after them. I think it's a good idea. We'll have a lot of prose to fix if we follow her lead. We can start any time :)

filipesilva: Done

public/docs/ts/latest/guide/animations.jade
filipesilva: Done.

PR #3442, opened 2017-03-26
Added angular2 firebase

AG-Systems:


Foxandxss:
Thanks but we are migrating to a new site and we are writing this page from the ground up. So merging this here wouldn't make a difference now.

AG-Systems:
Thank you @Foxandxss

PR #3445, opened 2017-03-27
docs(structural-directives): <template> was replaced by <ng-template>

wKoza:
Angular 4 uses now `<ng-template>` for structural directives. `<template>` is deprecated.


wKoza:
@wardbell I've updated  `structural-directives.jade`

wardbell:
Thanks!  I have a few more to correct but this gets most of them.

PR #3467, opened 2017-03-30
docs(toh): change data flow to data flows

kapunahelewong:
Change "data flow" to "data flows" and edit punctuation. Also checked all the other toh pages for plural data, but found none.

@kwalrath fyi


kwalrath:
LGTM. Thanks, Kapunahele!

PR #3469, opened 2017-03-30
Remove hover pseudo-selector from selected LI.

iammcgaber:
Having the :hover pseudo-selector on the `.heroes li.selected` is redundant and only makes the selected hero distinguishable if you hover your mouse over all items in the list.  I propose removing it.


iammcgaber:
I signed it!

PR #3470, opened 2017-03-30
docs(toh-2): clarify onSelect() method placement

iammcgaber:
docs(toh-2): clarify onSelect() method placement

As a beginner to Angular, it was unclear from the instructions where the placement of the `onSelect()` method should go.  I propose clarifying that it should be placed in the export statement.


iammcgaber:
I signed it!

wardbell:
I do not understand. The `OnSelect` method is added to `AppComponent`. And there is a code snippet immediately after that sentence that illustrates.

Meanwhile, I have no idea what "placed in the export statement" means. There is no such thing as an "export statement" (there are two export _declaration_, neither of them involving this method).

We're also about to refresh the tutorial in the next few months.

For all of these reasons, with regret I must reject this PR.

PR #3476, opened 2017-03-31
docs(DI): OpaqueToken was replaced by InjectionToken

wKoza:
Since Angular 4, OpaqueToken was replaced by InjectionToken.

### DIFF
--- a/public/docs/ts/latest/glossary.jade
+++ b/public/docs/ts/latest/glossary.jade
@@ -250,7 +250,7 @@ a#decoration
     At the core, an [`injector`](#injector) returns dependency values on request.
     The expression `injector.get(token)` returns the value associated with the given token.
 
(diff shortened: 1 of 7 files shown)


public/docs/ts/latest/glossary.jade
```diff
@@ -250,7 +250,7 @@ a#decoration
     At the core, an [`injector`](#injector) returns dependency values on request.
     The expression `injector.get(token)` returns the value associated with the given token.
 
-    A token is an Angular type (`OpaqueToken`). You rarely need to work with tokens directly; most
+    A token is an Angular type (`InjectionToken`). You rarely need to work with tokens directly; most
```
wardbell: Actually, it provides type information that may be valuable but it provides almost zero type safety.

wardbell:
Thanks for taking this on. As it happens, we addressed this matter in another PR #3479

PR #3495, opened 2017-04-02
Fix minor text error in toh-pt1.jade. The 'hero' property is in the AppComponent class, not in Hero.

Double-A-92:
The 'hero' property is in the AppComponent class, not in Hero.


Double-A-92:
I signed it!

wardbell:
Good catch. Thanks.

PR #3496, opened 2017-04-02
docs(aio): mention import of OnInit

adiherzog:


wardbell:
While I agree that the current prose is a bit confusing when it says "don't copy this into your code" ... you do need to import `OnInit` from `@angular/core`, I fear your suggestion will add to the confusion.

We do NOT "extend the Angular core import statement." Rather we _declare_ that the `AppComponent` implements `OnInit` and _implement_ that interface in the body of `AppComponent` by adding the `ngOnInit()` method.

We're going to refresh the entire tutorial soon so I think it is best if we just leave this as it is. Thanks for contributing.

adiherzog:
Thanks for the update. What I meant by "extend the Angular core import statement" was just changing 
`import { Component } from '@angular/core';` to `import { Component, OnInit } from '@angular/core';` so that I have `OnInit` available to make `AppComponent` implement it. I guess we talk about the same thing :-) So please feel free to improve the passage in any way that makes it better. Thanks.

I just reported this because up to this point the tutorial is very precise and every little detail was perfectly described. Great tutorial from what I've seen so far!

PR #3501, opened 2017-04-04
Deleted Repetitive Information.

### LINKED ISSUE #318: Edits needed to Heirarchical DI page
Here's some suggested edits.

1. The first few paragraphs are a bit redundant. the first 3 and then the second 3 say the same thing.

2. the last diagram under the heading "the injector tree" and the paragraphs above it are confusing. I think it's mostly the diagram. The paragraphs talk about components labeled as A, B, and C but only the instances are labeled. 

3. in the last paragraph, the first sentence is missing a word.
"If we had only defined a binding for RestoreService on the root component, we would have exactly one instance of the across the entire applicatoin."

lazaridis-com:
This closes #318 (pos. 2 and 3 of the issue were already solved)

### DIFF
--- a/public/docs/ts/latest/guide/hierarchical-dependency-injection.jade
+++ b/public/docs/ts/latest/guide/hierarchical-dependency-injection.jade
@@ -2,11 +2,7 @@ block includes
   include ../_util-fns
 
 :marked
-  You learned the basics of Angular Dependency injection in the
-  [Dependency Injection](./dependency-injection.html) guide.
-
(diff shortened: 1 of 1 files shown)


wardbell:
Thanks. I disagree that this is repetitive. It links back to the base discussion of DI. The entire DI section is scheduled for refresh anyway. Therefore, I'm closing your PR.

lazaridis-com:
[issue.process]

@wardbell , you disagree? With a fact? - Anyway, i guess you can close issue #318 then.

PR #3517, opened 2017-04-06
docs(reactive-forms): fix "Setup guide" link

FrankiePo:


FrankiePo:
I signed it!

wardbell:
Actually, it should just be `(setup.html)` rather than `(../guide/setup.html)`.  But I'll take the correction and clean up later. Thanks.

PR #3523, opened 2017-04-11
Avoiding Hero duplication

krzysiekpiasecki:
Prevent adding duplicates when '''enter''' and then '''blur''' events occured.


krzysiekpiasecki:
I signed it!

krzysiekpiasecki:
It;s about a duplication issue @googlebot

Foxandxss:
Ohh the irony.

wardbell:
Thanks for this change. But, believe it or not, I _intended_ for the duplication in the one case and not in the other. Because it is intended (which doesn't make it right, just intentional), I'm closing your PR. Thanks again.

